PR #49, opened 2022-12-23
Fixes jobs example

mingkang111:
This PR aims to fix the bug in jobs, and update tensorflow pack used in example.


mayoor (approved):

darenr (commented):
Can you make this DRY with a single const at the top of the file to avoid duplication please.

PR #59, opened 2023-01-12
Updated requirements.txt to work when not in a project

jdesanto:
With the `-e .` the following error occurs

(ads-docs) % pip install -r requirements.txt
Obtaining file:///.../accelerated-data-science/docs (from -r requirements.txt (line 1))
ERROR: file:///.../accelerated-data-science/docs (from -r requirements.txt (line 1)) does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found.


darenr:
I think the instructions must be wrong because we need to keep `"-e ."`

PR #82, opened 2023-02-14
Odsc 36246/catboost to generic model

z7ye:


docs/source/user_guide/model_registration/quick_start.rst
```diff
@@ -263,31 +291,51 @@ Other Frameworks
...
+    catboost_estimator = CatBoostRegressor(iterations=2,
+                            learning_rate=1,
+                            depth=2)
+    # Train a CatBoostRegressor model
+    catboost_estimator.fit(X_train, y_train)
+
+    # Get predictions
+    preds = catboost_estimator.predict(X_test)
+
+    # Instantite ads.model.generic_model.GenericModel using the trained Custom Model using the trained CatBoost Classifier  model
```
mayoor, line 323: Instantiate

docs/source/user_guide/model_registration/quick_start.rst
```diff
@@ -263,31 +291,51 @@ Other Frameworks
...
+    catboost_model = GenericModel(estimator=catboost_estimator, artifact_dir=tempfile.mkdtemp())
```
mayoor, line 324: It might not be obvious to the user how are we serializing this. I think we should also provide serde argument here so that user knows what serialization is used.

I think we should also add few examples in `Other Framework` sections under Frameworks, to show how user could save the estimator themselves and copy over to the artifact dir.

docs/source/user_guide/model_registration/quick_start.rst
```diff
@@ -263,31 +291,51 @@ Other Frameworks
...
-    model_id = generic_model.save(display_name="Custom Framework Model")
+    catboost_model.prepare(
+        inference_conda_env="oci://bucket@namespace/path/to/your/conda/pack",
+        inference_python_version="your_python_version",
+        X_sample=X_train,
+        y_sample=y_train,
+    )
+
+    # Verify generated artifacts
+    catboost_model.verify(X_test, auto_serialize_data=True)
```
mayoor, line 335: Let us also add in the comment what will be payload look like when the auto_serialize is True.

docs/source/user_guide/model_registration/quick_start.rst
z7ye: good call. will add.

docs/source/user_guide/model_registration/quick_start.rst
z7ye: fixed.

docs/source/user_guide/model_registration/quick_start.rst
z7ye: resolved.

mayoor (approved):

PR #83, opened 2023-02-14
Update mvs docs

jdesanto:
Documentation updates:
* Changed "DataFlow" to "Data Flow" when referring to the OCI service
* Move Model Version set to a dedicated section following Registration
* Broaden the registration section to include managing models

Note: This work was included in an earlier PR that was withdrawn.

Screen Shot 2023-02-14 at 3 34 20 PM


jdesanto:
@darenr PDF cuts off the side bar.

docs/source/user_guide/model_registration/introduction.rst
```diff
@@ -51,9 +51,16 @@ Register
   model_schema
   model_metadata
   model_file_customization
+
+Manage Model
```
mayoor, line 55: I am wondering if it is better to push `model_version_set` under `Register Model` since the versioning happens in the context of registration.

docs/source/user_guide/model_registration/introduction.rst
```diff
@@ -51,9 +51,16 @@ Register
...
+------------
+
+.. toctree::
+  :maxdepth: 1
+
   model_version_set
```
mayoor, line 61: The title inside `model_version_set` could be task instead of our taxonomy. My suggestion is that we change the title to `Version Model`

darenr: would definitely prefer not to use the word set

mrDzurb (commented):
This PR probably needs to be merged with develop branch?

mrDzurb (approved):

liudmylaru (approved):

PR #84, opened 2023-02-16
huggingface userguide

z7ye:


docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
```diff
@@ -0,0 +1,237 @@
...
+    >>> # Installation instructions can be found here: https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/pythonsdk.htm
+
+    >>> import requests
+    >>> import oci
+    >>> import ads
+    >>> import cloudpickle
+    >>> headers = {"Content-Type": "application/octet-stream"} 
+    >>> endpoint = huggingface_pipeline_model.model_deployment.url + "/predict"
+
+    >>> preds = requests.post(endpoint, data=image_bytes, auth=ads.common.auth.default_signer()['signer'], headers=headers).json()
```
mayoor, line 172: the variable `image_bytes` will not be valid if the user copies the block. Can you please add the code to create `image_bytes`

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
```diff
@@ -0,0 +1,237 @@
...
+
+    ## download the pretrained model
+    classifier = pipeline(model="openai/clip-vit-large-patch14")
+    classifier(
+            images=image,
+            candidate_labels=["animals", "humans", "landscape"],
+        )
+
+    ## Initiate a HuggingFacePipelineModel instance
+    zero_shot_image_classification_model = HuggingFacePipelineModel(classifier, artifact_dir=empfile.mkdtemp())
```
mayoor, line 204: I dont find the import statement corresponding to `HuggingFacePipelineModel`

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
```diff
@@ -0,0 +1,237 @@
...
+
+Example
+=======
+
+.. code-block:: python3
+
+    from transformers import pipeline
+    import tempfile
+    import PIL.Image
+    import ads
```
mayoor, line 187: Instead of `import ads`, I think it will be better to import the specific model. `import ads` will take more time than `from ads.common.auth import default_signer`

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
```diff
@@ -0,0 +1,237 @@
...
+    zero_shot_image_classification_model = HuggingFacePipelineModel(classifier, artifact_dir=empfile.mkdtemp())
+
+    # Autogenerate score.py, serialized model, runtime.yaml
+    conda_pack_path = "oci://bucket@namespace/path/to/conda/pack"
+    python_version = "3.x"
+    zero_shot_image_classification_model.prepare(inference_conda_env=conda_pack_path, inference_python_version = python_version, force_overwrite=True)
+
+    ## Test data
+    data = {"images": image, "candidate_labels": ["animals", "humans", "landscape"]}
+    body = cloudpickle.dumps(data) # convert image to bytes
```
mayoor, line 213: "Convert payload to bytes'

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
```diff
@@ -0,0 +1,237 @@
...
+            images=image,
+            candidate_labels=["animals", "humans", "landscape"],
+        )
+
+    ## Initiate a HuggingFacePipelineModel instance
+    zero_shot_image_classification_model = HuggingFacePipelineModel(classifier, artifact_dir=empfile.mkdtemp())
+
+    # Autogenerate score.py, serialized model, runtime.yaml
+    conda_pack_path = "oci://bucket@namespace/path/to/conda/pack"
+    python_version = "3.x"
```
mayoor, line 208: User might overlook 3.x. Can we give the slug to the pytorch pack?

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
z7ye: yup, fixed. Thanks!

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
z7ye: good catch. fixed.

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
z7ye: fixed.

z7ye: The reason I did this is the existing pytorch conda pack would not work for this vision example since it requires to install timm(cannot remember exactly) on top of it. I thought about showing a nlp example instead. but I think it's important to show image as a payload than text which is already json serialized, so I kept this example. I will add extra comments here to address it if you have concerns about it. :)

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
z7ye: done

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
z7ye: resovled

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
mayoor: Got it! Thanks!

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
```diff
@@ -0,0 +1,246 @@
...
+
+If your model takes more than one argument, you can pass in through dictionary with the keys as the argument name and values as the value of the arguement.
+
+.. code-block:: python3
+
+    >>> your_huggingface_pipeline_model.verify({"parameter_name_1": "parameter_value_1", ..., "parameter_name_n": "parameter_value_n"})
+    >>> your_huggingface_pipeline_model.predict({"parameter_name_1": "parameter_value_1", ..., "parameter_name_n": "parameter_value_n"})
+
+
+Run Prediction with oci sdk
```
mayoor, line 152: Here the invocation is not really through OCI SDK, we are using requests. We can say - Run Prediction using `requests`. In other examples I think we have used oci raw-curl, that is another option.

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
```diff
@@ -0,0 +1,237 @@
...
+.. code-block:: python3
+
+    >>> # Deploy and create an endpoint for the huggingface_pipeline_model 
+    >>> huggingface_pipeline_model.deploy(
+    ...     display_name="HuggingFace Pipeline Model For Image Segmentation",
+    ...     deployment_log_group_id="ocid1.loggroup.oc1.xxx.xxxxx",
+    ...     deployment_access_log_id="ocid1.log.oc1.xxx.xxxxx",
+    ...     deployment_predict_log_id="ocid1.log.oc1.xxx.xxxxx",
+    ... )
+    >>> print(f"Endpoint: {huggingface_pipeline_model.model_deployment.url}")
```
mayoor, line 105: we can print the end point using `huggingface_pipeline_model.model_deployment.url`. That will be better. There is a chance that the url pattern changes over time.

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
```diff
@@ -0,0 +1,237 @@
...
+.. code-block:: python3
+
+    >>> # Download an image
+    >>> import PIL.Image
+    >>> import requests
+    >>> import cloudpickle
+    >>> image_url = "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/pipeline-cat-chonk.jpeg"
+
+    >>> image = PIL.Image.open(requests.get(image_url, stream=True).raw)
+    >>> image_bytes = cloudpickle.dumps(image)
```
mayoor, line 120: I believe this was meant for block below? In this code we are not using image_bytes

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
mayoor: I think the definition of the image_bytes in wrong block - https://github.com/oracle/accelerated-data-science/pull/84/files/805111faa5ad1900d48f89799a8cb4f686770d12#r1121144821

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
z7ye: The reason I didnt use oci raw-curl is that I checked OCI-CLI. it says on the documentation it only supports json. but I am trying to pass in bytes.

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
z7ye: removed.

docs/source/user_guide/model_registration/frameworks/huggingfacemodel.rst
z7ye: fixed.

mayoor (approved):

PR #106, opened 2023-03-13
Updated licensing

jdesanto:
These changes include third party licensing for any libraries that might be installed via setup.py.


liudmylaru (approved):

darenr:
Please do not merge, GPL and Google license addition are big red flags, these both need to go through legal approvals (Keith) - we cannot approve these. We didn't add mysql_connector for just this reason.

PR #121, opened 2023-03-27
Bug fixes in ads opctl command: container entrypoint and distributed …

mayoor:
### Following fixes are covered in this PR - 

1. `ads opctl run` for distributed training jobs works only when the image is present locally. This happens because the tool tries to automatically build and publish the image to the registry. There can be situations where user is reusing image that is already in the registry and late binding the code. User will have to use `--nobuild` option to suppress build and publish.
2. Fix for entrypoint check when user tries to run an image using ads opctl. Eg - `ads opctl run --backend local --image iad.ocir.io/ociodscdev/myimage`

pr-screenshot-disttrain

### DIFF
--- a/ads/opctl/cmds.py
+++ b/ads/opctl/cmds.py
@@ -141,7 +141,7 @@ def _save_yaml(yaml_content, **kwargs):
     yaml_content : str
         YAML content as string.
     """
-    if kwargs["job_info"]:
+    if kwargs.get("job_info"):
         yaml_path = os.path.abspath(os.path.expanduser(kwargs["job_info"]))
         if os.path.isfile(yaml_path):
             overwrite = input(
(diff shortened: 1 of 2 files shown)


ads/opctl/cmds.py
```diff
@@ -210,7 +210,7 @@ def run(config: Dict, **kwargs) -> Dict:
                 "backend operator for distributed training can either be local or job"
             )
         else:
-            if not kwargs["dry_run"]:
+            if not kwargs["dry_run"] and not kwargs["nobuild"]:
```
mayoor, line 213: The verify and publish is relavant only when the image is built or present locally. If user specifies `--nobuild`option while running distributed training, we will assume that the image is already present in the registry. This is required to address the situation where user is reusing the image that was built and published to the registry.

ads/opctl/config/resolver.py
```diff
@@ -155,7 +155,11 @@ def _resolve_source_folder_path(self) -> None:
     def _resolve_entry_script(self) -> None:
         # this should be run after _resolve_source_folder_path
         if not self._is_ads_operator():
-            if os.path.splitext(self.config["execution"]["entrypoint"])[1] == ".ipynb":
+            if (
+                self.config["execution"].get("entrypoint")
```
mayoor, line 159: Entrypoint is not mandatory

qiuosier (approved):

PR #127, opened 2023-03-29
ODSC-39737 : allow to use `GenericModel.predict()` locally

mingkang111:
# Description
Related JIRA:https://jira.oci.oraclecorp.com/browse/ODSC-39737
Currently, the GenericModel supports two methods that can be used to test the model:
* verify()
* predict()

The difference between these two methods is that verify is used for local testing, where predict sends the request to the ML endpoint. Having two different methods confuses users, they don't really understand why we need two different methods for inferring the model. 

# Changed 
* Add a new attribute - `local=True/False` to the predict method to be able to invoke the prediction locally.

# User Experience
```
prediction = GenericModel.predict(local=True)
```
Screenshot 2023-04-05 at 3 54 17 PM

# Test
Screenshot 2023-03-29 at 1 08 10 PM

## Int test in TC
TC: https://teamcity.oci.oraclecorp.com/buildConfiguration/Odsc_Ads_AdsCiCd_CiTrigger/38127161?buildTab=buildParameters
Screenshot 2023-04-05 at 3 51 59 PM

### DIFF
--- a/tests/unitary/with_extras/model/test_generic_model.py
+++ b/tests/unitary/with_extras/model/test_generic_model.py
@@ -168,7 +168,20 @@ OCI_MODEL_PROVENANCE_PAYLOAD = {
     "training_script": None,
 }
 
+INFERENCE_CONDA_ENV= "oci://service-conda-packs@ociodscdev/service_pack/cpu/General_Machine_Learning_for_CPUs/1.0/mlcpuv1"
+TRAINING_CONDA_ENV="oci://service-conda-packs@ociodscdev/service_pack/cpu/Oracle_Database_for_CPU_Python_3.7/1.0/database_p37_cpu_v1"
+DEFAULT_PYTHON_VERSION = "3.8"
+MODEL_FILE_NAME = "fake_model_name"
 
+def _prepare(model):
+    model.prepare(
+        inference_conda_env=INFERENCE_CONDA_ENV,
+        inference_python_version=DEFAULT_PYTHON_VERSION,
+        training_conda_env=TRAINING_CONDA_ENV,
+        training_python_version=DEFAULT_PYTHON_VERSION,
+        model_file_name=MODEL_FILE_NAME,
+        force_overwrite=True,
+    )
 class TestEstimator:
     def predict(self, x):
         return x**2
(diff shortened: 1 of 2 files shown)


qiuosier (approved):

tests/unitary/with_extras/model/test_generic_model.py
```diff
@@ -168,7 +168,20 @@
     "training_script": None,
 }
 
-
+INFERENCE_CONDA_ENV= "oci://service-conda-packs@ociodscdev/service_pack/cpu/General_Machine_Learning_for_CPUs/1.0/mlcpuv1"
```
mrDzurb, line 171: We don't need to use real paths, right?

mrDzurb (approved):

tests/unitary/with_extras/model/test_generic_model.py
mingkang111: Yes, we only need slug name I think. I will create a PR to clean up the paths in this file.

mayoor (dismissed):
As a followup, I have a proposal - 

If user calls `predict` and the `url` is None, when we throw an error, we can provide a message saying, `Error invoking the remote endpoint as the model is not deployed yet or the endpoint information is not available. If you intend to invoke inference using locally available model artifact, set parameter local=True`

liudmylaru (dismissed):

mingkang111:
> As a followup, I have a proposal -
> 
> If user calls `predict` and the `url` is None, when we throw an error, we can provide a message saying, `Error invoking the remote endpoint as the model is not deployed yet or the endpoint information is not available. If you intend to invoke inference using locally available model artifact, set parameter local=True`

tests/unitary/with_extras/model/test_generic_model.py
mingkang111: Replaced with fake path.

mingkang111:
> As a followup, I have a proposal -
> 
> If user calls `predict` and the `url` is None, when we throw an error, we can provide a message saying, `Error invoking the remote endpoint as the model is not deployed yet or the endpoint information is not available. If you intend to invoke inference using locally available model artifact, set parameter local=True`

That's a good idea! I just improved the error message as suggested. Thanks Mayoor.

mrDzurb (approved):

mayoor (approved):

PR #135, opened 2023-04-05
Publish to pypi action -> Main

liudmylaru:
## Description

Building action workflow to publish ADS to PiPI. This change add special yml file into .github/workflows folder. This yml has to be merged to main branch so that it can be recognized correctly by GitGub and appear and "Actions" section to be triggered. To validate that action works as expected I added command which will publish to TestPiPI - followed this documentation: https://packaging.python.org/en/latest/guides/using-testpypi/.

If actions will run correctly and ADS will be published to TestPyPI - I will update the command to publish to PyPI then and planning to have PR to develop, so that it will propagate to release branch with release and this Action will be used in next release.

Jira: https://jira.oci.oraclecorp.com/browse/ODSC-37152

## What was done

- created account in TestPiPI to validate publish working (will be able to check it after this PR merged to main)
- added file .github/workflows/publish-to-pypi.yml

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python
-      - uses: actions/setup-python@v4
+        uses: actions/setup-python@v4
         with:
           python-version: "3.x"
       - name: Build distribution 📦
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,30 @@
+name: "[DO NOT TRIGGER] Publish to PyPI"
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,30 @@
+name: "[DO NOT TRIGGER] Publish to PyPI"
+
+on:
+  # To run this workflow manually from the Actions tab
+  workflow_dispatch:
```
qiuosier, line 5: ":" is probably not needed?

qiuosier (approved):

mrDzurb (approved):

.github/workflows/publish-to-pypi.yml
liudmylaru: Most likely yes - no needed. But it should work either way. I will remove it later - when will add step to publish to real PiPI (not TestPyPI).

PR #138, opened 2023-04-06
Update Jobs YAML Schema

qiuosier:
- Adding a separated page (`docs/source/user_guide/jobs/yaml_schema.rst`) under Data Science Jobs to show YAML schema in tables.
- Adding a Python script (`docs/source/yaml_schema/schema2html.py`) to generate an HTML page containing the tables to show the YAML schema.
- The HTML and YAML schema files are saved in `docs/source/yaml_schema/jobs`.


docs/source/yaml_schema/schema2html.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python
```
z7ye, line 1: very cool that we have a script to do conversion.  Do you think it's useful to add `required` info in the table as well?

agree that we need more description for sure. and eventually we could format the table a little nicer. :)  but this is very awesome start version.

z7ye (approved):

docs/source/yaml_schema/schema2html.py
qiuosier: Yes, it is good to show whether the field is required. However, for now I don't have enough information for some of the YAML schemas. I am also not sure if we should mark a field as required if a field can be fetched from environment variable. We should first update the schema with more validation rules, then update the docs.

docs/source/yaml_schema/schema2html.py
```diff
@@ -0,0 +1,156 @@
...
+# -*- coding: utf-8 -*--
+
+# Copyright (c) 2021, 2022 Oracle and/or its affiliates.
```
liudmylaru, line 4: 2023 year - instead of 2021, 2022

liudmylaru (approved):
I builded the docs locally from current branch, opened created section, checked that internal and external links are working, checked that formatting looks good. Had not checked if context is correct - I believe yes! Thank you for great doc pages and script!

docs/source/yaml_schema/jobs/dataFlow.html
```diff
@@ -0,0 +1,177 @@
+<style>
+    table,
+    th,
+    td {
+        border: 1px solid black;
```
mrDzurb, line 5: Maybe it would be better to use a grey color (#ccc) for the borders? 
image

mrDzurb (approved):

PR #140, opened 2023-04-07
ODSC-41193: Adds API section to the telemetry.

mrDzurb:
## Description
Extends telemetry with the API information. 
API section will be populated form the **EXTRA_USER_AGENT_INFO** environment variable.

## Local Tests
image


ads/telemetry/telemetry.py
```diff
@@ -43,12 +48,33 @@ def surface(cls):
...
+
+    try:
+        config = config or {}
+        if not config.get(USER_AGENT_KEY):
+            config.update(
+                {
+                    USER_AGENT_KEY: (
+                        f"{LIBRARY}/version={__version__}/"
+                        f"surface={Surface.surface().name}/"
+                        f"api={os.environ.get(EXTRA_USER_AGENT_INFO,UNKNOWN) or UNKNOWN}"
```
mayoor, line 73: Do we need `or UNKNOWN `?

mrDzurb: This is on case when EXTRA_USER_AGENT_INFO exists but with empty value. I don't have a strong opinion here, maybe it would be better to not add api at all in this case or leave it empty.

tests/unitary/default_setup/telemetry/test_agent.py
```diff
@@ -82,17 +119,18 @@ def test_user_agent_default_signer_known_resources(
             auth_info = ads.auth.default_signer()
             assert (
                 auth_info["config"].get("additional_user_agent")
-                == f"Oracle-ads/version={ads.__version__}/surface={USER_AGENT_VALUE}"
+                == f"{LIBRARY}/version={ads.__version__}/surface={EXPECTED_RESULT['USER_AGENT_VALUE']}/api={EXPECTED_RESULT[EXTRA_USER_AGENT_INFO]}"
             )
         monkeypatch.delenv("OCI_RESOURCE_PRINCIPAL_VERSION", raising=False)
```
qiuosier, line 124: Do we need delenv? I thought monkeypatch will unset the variables once the test is finished.

qiuosier (approved):

PR #144, opened 2023-04-12
Improved model deployment progress bar

lu-ohai:
### Improved model deployment progress bar

- Updated model deployment progress bar.
- Moved progress bar API to `OCIWorkRequestMixin` class so that it can be used by other services.

### Notebook
Screen Shot 2023-04-17 at 12 05 25 AM
Screen Shot 2023-04-17 at 12 13 10 AM
Screen Shot 2023-04-17 at 12 23 08 AM
Screen Shot 2023-04-17 at 12 33 43 AM


ads/model/service/oci_datascience_model_deployment.py
```diff
@@ -512,89 +484,76 @@ def from_id(cls, model_deployment_id: str) -> "OCIDataScienceModelDeployment":
...
-        work_flow_step: int,
-        disallowed_final_states: List[str],
-        work_flow_request_id: str,
-        state: State,
-        model_deployment_id: str,
-        max_wait_time: int = DEFAULT_WAIT_TIME,
-        poll_interval: int = DEFAULT_POLL_INTERVAL,
-    ):
-        """_wait_for_progress_completion blocks until progress is completed.
+    def _wait_for_work_request(
```
mrDzurb, line 487: In the ticket it says that progress bar should be implemented as part of a Mixin class, but the implementation is different. Shouldn't we have some solution that can be reused for the different services the we do support?

lu-ohai: Thanks, the method has been moved to class `OCIWorkRequestMixin` for general use.

qiuosier (approved):

ads/common/oci_mixin.py
```diff
@@ -29,6 +30,10 @@
 logger = logging.getLogger(__name__)
 
 LIFECYCLE_STOP_STATE = ("SUCCEEDED", "FAILED", "CANCELED", "DELETED")
+WORK_REQUEST_STOP_STATE = ("SUCCEEDED", "FAILED", "CANCELED")
```
qiuosier, line 33: This might depend on the service.

ads/common/oci_mixin.py
```diff
@@ -931,6 +936,76 @@ def get_work_request_response(
             )
         return work_request_response
 
+    def wait_for_progress(
+        self, 
+        work_request_id: str, 
+        num_steps: int = DEFAULT_WORKFLOW_STEPS,
```
qiuosier, line 942: It does not seem like there is an easy way to know the number of workflow steps for different workflow. Having a default here could be confusing for the developers. I think a better behavior is, while the number of steps is not specified, it will continue waiting until work_request.status reach WORK_REQUEST_STOP_STATE. Even though WORK_REQUEST_STOP_STATE still depends on the workflow, most, if not all workflow has SUCCEEDED and FAILED states.

qiuosier (approved):
Even though I think there is still room for improvement. This update simplified the existing logic and provided a basic reusable method. We can keep improving it when we have additional use cases.

z7ye (approved):

ads/common/oci_mixin.py
lu-ohai: These are the `WorkRequest` terminal states and do the services have different values?

PR #145, opened 2023-04-12
Update `ads opctl run -b local` to print container logs and log docker run cmd for debugging.

qiuosier:
It is better to `print` instead of `log` the container logs as they are expected outputs to the terminal.

Currently the container logs are handled by `logger.info()`. The default log level is warning. So user will not see anything unless the debug flag (-d) is used. Normally users would expect to see the container logs in the terminal. For example, when a user run `echo` command, the user would expect to see the string printed. But currently nothing is showed.
image

Also, striping each line of the output will also remove tabs and spaces at both ends. It is better to print the output as is and use empty string for the `end` argument.

For example, with `print(line.decode("utf-8").strip())`, the `cat --help` command shows the following (indentations are gone):
image

If we use `print(line.decode("utf-8"), end="")`, the output looks better :)
image

For debugging purpose, it is also better to print out the equivalent docker run command to help users running the container directly.

### DIFF
--- a/ads/opctl/utils.py
+++ b/ads/opctl/utils.py
@@ -10,12 +10,12 @@ import logging
 import os
 import subprocess
 import sys
+import shlex
 import tempfile
 import urllib.parse
 from distutils import dir_util
 from subprocess import Popen, PIPE, STDOUT
 from typing import Union, List, Tuple, Dict
-import shlex
 import yaml
 
 import ads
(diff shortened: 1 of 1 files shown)


ads/opctl/utils.py
```diff
@@ -321,27 +321,47 @@ def run_container(
...
-    logger.info(f"Container ID: {container.id}")
-
-    for line in container.logs(stream=True, follow=True):
-        logger.info(line.decode("utf-8").strip())
-
-    result = container.wait()
-    container.remove()
-    return result.get("StatusCode", -1)
+    try:
+        container = client.containers.run(
```
mayoor, line 346: Since we are here, could you also add a option to pass `network` parameter? The default could be None, but user could pass it through environment variable?

This applies to both run and build API

ads/opctl/utils.py
```diff
@@ -321,27 +321,47 @@ def run_container(
...
+    if env_vars:
+        docker_run_cmd.extend([f"-e {key}={val}" for key, val in env_vars.items()])
+    if bind_volumes:
+        docker_run_cmd.extend(
+            [f'-v {source}:{bind.get("bind")}' for source, bind in bind_volumes.items()]
+        )
+    docker_run_cmd.append(image)
+    if command:
+        docker_run_cmd.append(command)
+    logger.debug(" ".join(docker_run_cmd))
```
mayoor, line 337: I like this! Thanks!

mayoor: Is it possible to generate this command using docker client? If the client has some sort of debug mode to show verbose logs while starting the container it might be helpful.

qiuosier: I don't see any api to print out the equivalent cmd. My guess is the CLI and Python API are both making REST requests under the hood.

ads/opctl/utils.py
z7ye: agree. this will be useful when I run a byoc/triton container in the future so that I can send request to it.

z7ye (approved):

z7ye (approved):

ads/opctl/utils.py
qiuosier: I see there are many arguments for the `client.containers.run()`, and a few of them are related to `network`. However we do not have a test case for adding the `network` parameter. Maybe it is better to create a use case and add it with another PR? Can you create a ticket with an example? I can look into it.

mayoor (dismissed):

mayoor (approved):

PR #164, opened 2023-05-01
Update unit tests to set compartment ID if it is not available from environment variables.

qiuosier:
A couple tests requires environment variables to be set before running them, otherwise there will be an error collecting the tests. I think it is better for the tests to have less dependency on the environment. This PR adds a default value value if the environment variable is not set.

### DIFF
--- a/.github/workflows/run-unittests-default_setup.yml
+++ b/.github/workflows/run-unittests-default_setup.yml
@@ -77,7 +77,6 @@ jobs:
         timeout-minutes: 15
         shell: bash
         env:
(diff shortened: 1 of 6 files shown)


liudmylaru:
Like this change! If we removing dependency for this env variable can you please remove setting up NB_SESSION_COMPARTMENT_OCID env variable in .github/workflows/run-unittests.yml and in .github/workflows/run-unittests-default_setup.yml files (just remove lines 99 and 77 accordingly). Thanks!

liudmylaru (approved):

mrDzurb (approved):

liudmylaru (approved):

PR #167, opened 2023-05-02
Support setting "defined tags" in data science jobs.

qiuosier:
- Adding `with_defined_tag()` to job runtime.
- Adding `with_freeform_tag()` and `with_defined_tag()` job infrastructure.
- Tags defined in runtime will overwrite the ones defined in infrastructure.
- Tags defined in runtime will also be added to job run.
- Tags defined in infrastructure will be added to job only.

### DIFF
--- a/tests/unitary/default_setup/catalog/test_catalog_project.py
+++ b/tests/unitary/default_setup/catalog/test_catalog_project.py
@@ -61,13 +61,13 @@ class ProjectCatalogTest(unittest.TestCase):
 
     with patch.object(auth, "default_signer"):
         with patch.object(oci_client, "OCIClientFactory"):
-            pc = ProjectCatalog()
+            project_id = "ocid1.projectcatalog.oc1.iad.<unique_ocid>"
+            comp_id = os.environ.get("NB_SESSION_COMPARTMENT_OCID", "ocid1.compartment.oc1.iad.<unique_ocid>")
+            date_time = datetime(2020, 7, 1, 18, 24, 42, 110000, tzinfo=timezone.utc)
+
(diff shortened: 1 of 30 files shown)


mrDzurb (approved):
Could you also update docstring examples for job and jobrun and add tags lines there?

tests/unitary/default_setup/catalog/test_catalog_project.py
```diff
@@ -61,14 +61,14 @@ class ProjectCatalogTest(unittest.TestCase):
 
     with patch.object(auth, "default_signer"):
         with patch.object(oci_client, "OCIClientFactory"):
-            pc = ProjectCatalog()
-            pc.ds_client = MagicMock()
-            pc.identity_client = MagicMock()
-
             project_id = "ocid1.projectcatalog.oc1.iad.<unique_ocid>"
```
ahosler, line 64: Should we have something here like: `project_id = os.environ.get("NB_PROJECT_OCID",  "ocid1.projectcatalog.oc1.iad.<unique_ocid>")`

ahosler (approved):
Great to see this 👍

tests/unitary/default_setup/catalog/test_catalog_project.py
qiuosier: It was like that in the unit tests. But I guess it does not matter. I am adding another PR to make unit tests have less dependency on environment variable.

mrDzurb (approved):

lu-ohai (approved):

lu-ohai (approved):

mrDzurb (approved):

PR #173, opened 2023-05-03
Support creating ADSDataset through pandas accessor

lu-ohai:
#### Example
```
df = pd.read_csv("oci://hosted-ds-datasets@bigdatadatasciencelarge/synthetic/orcl_attrition.csv")

# create ADSDataset
df.ads.dataset()

# create ADSDatasetWithTarget
df.ads.dataset_with_target(target="Attrition")
```

#### Notebook
Screen Shot 2023-05-03 at 12 47 15 PM


jdesanto:
Please add a check for if a user picks a column that does not exist.

jdesanto (approved):
I approve pending a test case for when the user selects a nonexistent column

qiuosier (dismissed):

jdesanto:
Just an additional comment about the requested improvement. The issue is whether the column exists. This should be checked and a warning should be returned if it does not exist. The corresponding unit test should test against this check.

lu-ohai:
> 

Just added the checker, please take a look

lu-ohai:
> Just an additional comment about the requested improvement. The issue is whether the column exists. This should be checked and a warning should be returned if it does not exist. The corresponding unit test should test against this check.

Just added the checker, please take a look

qiuosier (approved):

jdesanto (approved):

PR #193, opened 2023-05-09
Fixed suggested_sampling for imbalanced data in RecommendationTransformer

liudmylaru:
### Description

Investigation of reported issue that auto_transform() not applied upsampling revealed mistake in if statements in suggested sampling for imbalanced dataset. According comment for MAX_LEN_FOR_UP_SAMPLING variable (https://github.com/oracle/accelerated-data-science/blob/main/ads/common/utils.py#L61) up-sample has to be applied, but "Do Nothing" set (https://github.com/oracle/accelerated-data-science/blob/main/ads/dataset/recommendation_transformer.py#L258). Also observed that suggested_sampling has no "Up-sample" in any of if paths, which is also showing there is mistake in this if-else statement.

Jira: https://jira.oci.oraclecorp.com/browse/ODSC-33269

### What was done

- fixed mistake in suggested sampling for imbalanced data
- added unittest

### Validation

Apart of added unittests, I checked result with original data described in the ticket description.

BEFORE:
image

AFTER:
image


ads/dataset/recommendation_transformer.py
```diff
@@ -255,7 +255,7 @@ def _get_recommendations(self, df):
 
                 # Suggest down sampling if minor_majority_ratio is 1:1000
```
darenr, line 256: comment is now not entirely accurate - also the readability of this assignment would be improved if if rewritten. While not wrong, it's not as easy to follow written like this. Minor readability nitpik

darenr (dismissed):
approving since not wrong, but given that yuo changed this assignment it would be a good opportunity to improve the readability

ads/dataset/recommendation_transformer.py
liudmylaru: @darenr updated comment, please review. Thank you!

mingkang111 (approved):

ads/dataset/recommendation_transformer.py
```diff
@@ -253,14 +253,15 @@ def _get_recommendations(self, df):
...
-                    else "Down-sample"
-                    if minor_majority_ratio >= utils.MIN_RATIO_FOR_DOWN_SAMPLING
-                    else "Do nothing"
-                )
+                # up-sample if length of dataframe is less than or equal to MAX_LEN_FOR_UP_SAMPLING = 5000
+                # down-sample if minor_majority_ratio is greater than or equal to MIN_RATIO_FOR_DOWN_SAMPLING = 1/20
+                if len(df) <= utils.MAX_LEN_FOR_UP_SAMPLING:
+                    suggested_sampling = "Up-sample"
+                elif minor_majority_ratio >= utils.MIN_RATIO_FOR_DOWN_SAMPLING:
+                    suggested_sampling = "Down-sample"
```
mrDzurb, line 261: NIT: It would be nice to move sampling to constants.

ads/dataset/recommendation_transformer.py
```diff
@@ -253,14 +253,15 @@ def _get_recommendations(self, df):
...
 
-                # Suggest down sampling if minor_majority_ratio is 1:1000
-                suggested_sampling = (
-                    "Do nothing"
-                    if len(df) <= utils.MAX_LEN_FOR_UP_SAMPLING
-                    else "Down-sample"
-                    if minor_majority_ratio >= utils.MIN_RATIO_FOR_DOWN_SAMPLING
-                    else "Do nothing"
-                )
+                # up-sample if length of dataframe is less than or equal to MAX_LEN_FOR_UP_SAMPLING = 5000
```
mrDzurb, line 256: NIT: probably we don't need these comments. The logic now is pretty clear.

mrDzurb (approved):

PR #196, opened 2023-05-10
Fix API, that gives an error if target is not specified

liudmylaru:
### Description

Added user-friendly error message when methods suggest_recommendations() or get_recommendations() invoked without specifying target parameter.

Jira: https://jira.oci.oraclecorp.com/browse/ODSC-38735

### What was done

- added suggest_recommendations() with message to the user that the user needs to set the target
- added get_recommendations() with message to the user that the user needs to set the target
- added unittest

### DIFF
--- a/ads/dataset/dataset.py
+++ b/ads/dataset/dataset.py
@@ -1913,3 +1913,35 @@ class ADSDataset(PandasDataset):
 
         if visualize_features and not is_wide_dataset:
             self._visualize_feature_distribution(features)
+
+    def get_recommendations(self, *args, **kwargs):  # real signature may change
+        """
+        Returns user-friendly error message to set target variable before invoking this API.
+
+        Parameters
+        ----------
+        kwargs
+
+        Returns
+        -------
+        Exception
+            raises an :exc:`Exception`
+
+        """
+        raise Exception("Please set the 'target' variable before invoking this API.")
+
+    def suggest_recommendations(self, *args, **kwargs):  # real signature may change
+        """
+        Returns user-friendly error message to set target variable before invoking this API.
+
+        Parameters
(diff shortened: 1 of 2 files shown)


ads/dataset/dataset.py
```diff
@@ -1913,3 +1913,35 @@ def show_in_notebook(
...
+        ----------
+        kwargs
+
+        Returns
+        -------
+        Exception
+            raises an :exc:`Exception`
+
+        """
+        raise Exception("Please set the 'target' variable before invoking this API.")
```
mingkang111, line 1931: We have the `set_target()` [method](https://accelerated-data-science.readthedocs.io/en/latest/ads.dataset.html#ads.dataset.dataset.ADSDataset.set_target) in `ADSDataset` to help with that. Whenever the user hits the error of ```
Object `ds.suggest_recommendations` not found.``` , they can call `ds = ds.set_target(target="xxx")` to resolve it. 

I am wondering if we can extend this message like this:
"Please set the `target` using `set_target()` before invoking this API. See https://accelerated-data-science.readthedocs.io/en/latest/ads.dataset.html#ads.dataset.dataset.ADSDataset.set_target for the API usage."

Could anyone help to suggest on the idea of adding api doc link into the Exception message?  cc: @mrDzurb @qiuosier

mingkang111 (dismissed):

ads/dataset/dataset.py
qiuosier: I think your suggested message looks good :)
Probably we should raise an `NotImplementedError` instead of just the `Exception`.

mingkang111 (approved):

mrDzurb (approved):

PR #198, opened 2023-05-11
Added support for private key content for set_auth()

lu-ohai:
### Added support for private key content for set_auth()

- Support private key content through `config`
- Added validation for config

### Example
```
private_key_content = """
-----BEGIN RSA PRIVATE KEY-----
KEY_CONTENT
-----END RSA PRIVATE KEY-----
"""

config = {
    "user": "ocid1.user.oc1..xxx",
    "fingerprint": "34:78:73:45:0e:sd:f2:34:df:5s:f1:1b:de:b1:3d:f6",
    "tenancy": "ocid1.tenancy.oc1..xxxx",
    "region": "us-ashburn-1",
    "key_content":private_key_content,
}

ads.set_auth("api_key", config=config)
```

### DIFF
--- a/ads/common/auth.py
+++ b/ads/common/auth.py
@@ -117,7 +117,7 @@ def set_auth(
     >>> ads.set_auth("api_key", oci_config_location = "other_config_location") # use non-default oci_config_location
 
     >>> ads.set_auth("api_key", client_kwargs={"timeout": 60}) # default signer with connection and read timeouts set to 60 seconds for the client.
-    >>> ads.set_auth("api_key", )
+    >>> ads.set_auth("api_key", signer_kwargs={"key_content": "private_key_content"}) # Create config using key content
     >>> other_config = oci.config.from_file("other_config_location", "OTHER_PROFILE") # Create non-default config
     >>> ads.set_auth(config=other_config) # Set api keys type of authentication based on provided config
 
--- a/ads/common/auth.py
+++ b/ads/common/auth.py
@@ -66,6 +66,7 @@ class AuthState(metaclass=SingletonMeta):
         )
         self.oci_config = self.oci_config or {}
         self.oci_signer = self.oci_signer
+        self.oci_signer_callable = self.oci_signer_callable
         self.oci_signer_kwargs = self.oci_signer_kwargs or {}
         self.oci_client_kwargs = self.oci_client_kwargs or {}
 
--- a/ads/common/auth.py
+++ b/ads/common/auth.py
@@ -117,7 +117,7 @@ def set_auth(
     >>> ads.set_auth("api_key", oci_config_location = "other_config_location") # use non-default oci_config_location
 
     >>> ads.set_auth("api_key", client_kwargs={"timeout": 60}) # default signer with connection and read timeouts set to 60 seconds for the client.
-
+    >>> ads.set_auth("api_key", )
     >>> other_config = oci.config.from_file("other_config_location", "OTHER_PROFILE") # Create non-default config
     >>> ads.set_auth(config=other_config) # Set api keys type of authentication based on provided config
 
@@ -157,7 +157,7 @@ def set_auth(
 
     auth_state.oci_config = config
     auth_state.oci_key_profile = profile
-    if auth == AuthType.API_KEY and not signer and not signer_callable:
+    if auth == AuthType.API_KEY and not signer and not signer_callable and not signer_kwargs:
         if os.path.exists(os.path.expanduser(oci_config_location)):
             auth_state.oci_config_path = oci_config_location
         else:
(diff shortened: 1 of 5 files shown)


liudmylaru:
Maybe example can be added to the documentation section also: https://accelerated-data-science.readthedocs.io/en/latest/user_guide/cli/authentication.html#overriding-defaults. But I am not shore if it will overcrowd existing section. Maybe add "Example 4...." after  Example 3. https://github.com/oracle/accelerated-data-science/blob/main/docs/source/user_guide/cli/authentication.rst?plain=1#L138 in code section.

liudmylaru (dismissed):

lu-ohai:
> Maybe example can be added to the documentation section also: https://accelerated-data-science.readthedocs.io/en/latest/user_guide/cli/authentication.html#overriding-defaults. But I am not shore if it will overcrowd existing section. Maybe add "Example 4...." after Example 3. https://github.com/oracle/accelerated-data-science/blob/main/docs/source/user_guide/cli/authentication.rst?plain=1#L138 in code section.

Thanks Liuda, just added the example under Example 4.

liudmylaru (dismissed):

mrDzurb (dismissed):

ads/common/auth.py
```diff
@@ -66,7 +66,6 @@ def __post_init__(self):
         )
         self.oci_config = self.oci_config or {}
         self.oci_signer = self.oci_signer
-        self.oci_signer_callable = self.oci_signer_callable
```
liudmylaru, line 69: wondering why signer callable was removed form saved state of auth? will this if fail then with no attribute error?: https://github.com/oracle/accelerated-data-science/blob/main/ads/common/auth.py#L371

tests/unitary/default_setup/auth/test_auth.py
```diff
@@ -39,52 +47,80 @@ class TestEDAMixin(TestCase):
     def test_set_auth_overwrite_profile(
         self, mock_load_key_file, mock_path_exists, mock_config_from_file
     ):
-        set_auth(profile="TEST")
-        default_signer()
-        mock_config_from_file.assert_called_with("~/.oci/config", "TEST")
-        set_auth(profile="DEFAULT")
+        with pytest.raises(oci.exceptions.InvalidConfig):
```
liudmylaru, line 50: inside this with there are 5 lines of code. Which one of 5 lines expected to fail with the exception? If it fail on a 1st line others will be ignored and not run? And if now it is failing as expected, for example on a last line, we will not be able to see a change if somewhere in a future if, for example, it will start failing on a first line instead.

I see other tests were updated similar - many code lines inside with pytest.raises(oci.exceptions.InvalidConfig).

I would suggest to include into pytest.raises(oci.exceptions.InvalidConfig) only one line of code which expected to fail. If every line should fail with same exception add for each one into its own pytest.raises().

liudmylaru (dismissed):

ads/common/auth.py
```diff
@@ -506,15 +555,18 @@ def create_signer(self) -> Dict:
             configuration = ads.telemetry.update_oci_client_config(
                 oci.config.from_file(self.oci_config_location, self.oci_key_profile)
             )
+
+        oci.config.validate_config(configuration)
```
qiuosier, line 559: Why adding this here? This is why you are getting the "invalid config" error from the original unit tests.

ads/common/auth.py
```diff
@@ -205,7 +233,10 @@ def api_keys(
     >>> oc.OCIClientFactory(**auth).object_storage # Creates Object storage client with timeout set to 6000 using API Key authentication
     """
     signer_args = dict(
-        oci_config_location=oci_config,
+        oci_config=oci_config if isinstance(oci_config, Dict) else {},
+        oci_config_location=oci_config
```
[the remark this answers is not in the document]
mrDzurb: We don't modify `oci_config`, right? Here we just create a new dict with the same field name?

ads/common/auth.py
lu-ohai: The main reason to add the validation here to ensure the config read from file or dict passed in by user is legit before reaching service. If user fails to pass the legit config, then service will also throw error. We may want to handle this before hand.

mrDzurb: I think the main reason is to make sure that use didn't make typos in their config. For example 
instead of `configuration.get("key_content")` user can pass `configuration.get("key_kontent")` and our code still will work well, because we do `private_key_content=configuration.get("key_content")`. We can easily write our own validation method, but if the `oci.config.validate_config(configuration)` works, then i think we can move forward with this. However, It would be better to add this validation before the modifying the config with telemetry info.

mrDzurb (approved):

liudmylaru (approved):

PR #206, opened 2023-05-21
Removed max version pins on sklearn and pandas

jdesanto:


darenr:
This seems to break a fair number of unit tests. Whatever API change in pandas happened should be fixed first. If you unpin pandas it'll likely pick up 2.0.1+ which will have a number of breaking changes since it's a major version update.

darenr:
My suggestion is to change ADS to fix the sklearn issue.

PR #210, opened 2023-05-30
Added defined tags for data flow

lu-ohai:
### Added defined tags for data flow

- Added `with_freeform_tag()` and `with_defined_tag()` in `DataFlow` class. Note these two APIs end without `s` similar as [DataScienceJob](https://github.com/oracle/accelerated-data-science/blob/main/ads/jobs/builders/infrastructure/dsc_job.py#L1282-1300).

### Example
```
DataFlow()
.with_driver_shape("VM.Standard.E4.Flex")
.with_driver_shape_config(ocpus=2, memory_in_gbs=32)
.with_executor_shape("VM.Standard.E4.Flex")
.with_executor_shape_config(ocpus=4, memory_in_gbs=64)
.with_spark_version("3.2.1")
.with_freeform_tag(test_freeform_key="test_freeform_value")
.with_defined_tag(**{"Oracle-Tags": {"CreatedBy": "test_name@oracle.com"}})
```
Screenshot 2023-05-30 at 10 12 54 AM

Screenshot 2023-05-30 at 10 13 51 AM

### DIFF
--- a/ads/jobs/builders/infrastructure/dataflow.py
+++ b/ads/jobs/builders/infrastructure/dataflow.py
@@ -384,6 +384,8 @@ class DataFlow(Infrastructure):
     CONST_OCPUS = "ocpus"
     CONST_ID = "id"
     CONST_PRIVATE_ENDPOINT_ID = "private_endpoint_id"
+    CONST_FREEFORM_TAGS = "freeform_tags"
+    CONST_DEFINED_TAGS = "defined_tags"
 
     attribute_map = {
         CONST_COMPARTMENT_ID: "compartmentId",
@@ -402,6 +404,8 @@ class DataFlow(Infrastructure):
         CONST_OCPUS: CONST_OCPUS,
         CONST_ID: CONST_ID,
         CONST_PRIVATE_ENDPOINT_ID: "privateEndpointId",
+        CONST_FREEFORM_TAGS: "freeformTags",
+        CONST_DEFINED_TAGS: "definedTags"
     }
 
     def __init__(self, spec: dict = None, **kwargs):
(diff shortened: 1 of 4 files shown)


ads/jobs/builders/infrastructure/dataflow.py
```diff
@@ -775,6 +781,36 @@ def with_private_endpoint_id(self, private_endpoint_id: str) -> "DataFlow":
...
+
+        Returns
+        -------
+        DataFlow
+            The DataFlow instance (self)
+        """
+        return self.set_spec(self.CONST_DEFINED_TAGS, kwargs)
+
+    @property
+    def freeform_tags(self) -> dict:
```
mrDzurb, line 806: We probably don't need this property? Personally I would prefer to have them, but according to the current implementation, the ```__getattr__``` already takes care of this. However from the other side, we have different names for the  ```with_defined_tag``` and ```defined_tags```. I'm wondering how it worked for the **freeform_tags**? Did we ask users in our docs to use ```dataflow.freeform_tag``` instead of the ```dataflow.freeform_tags```?

mrDzurb:
We might also need to update user docs for this?

mrDzurb (dismissed):

ads/jobs/builders/infrastructure/dataflow.py
lu-ohai: Thanks for pointing this out. For the naming, the behavior is uniform across [DataScienceJob](https://github.com/oracle/accelerated-data-science/blob/main/ads/jobs/builders/infrastructure/dsc_job.py#L1282) and [Runtime](https://github.com/oracle/accelerated-data-science/blob/main/ads/jobs/builders/infrastructure/dsc_job.py#L1282) classes, that is use `with_defined_tag/with_freeform_tag` ending without `s` and use `defined_tags/freeform_tags` ending with `s` as properties name. If use `__getattr__`, users need to call `freeform_tag` instead of what other classes do. Also seems like we currently didn't suggest the useage of tags for dataflow in user guide.

mingkang111 (approved):

PR #211, opened 2023-05-30
Added validation for driver and executor shape details.

lu-ohai:
### Added validation for driver and executor shape details of DataFlow.

Validation for:
- `executor_shape` and `driver_shape` must be the same shape family.
- No shape config is needed for non-flex shape

Screenshot 2023-05-30 at 2 28 17 PM
Screenshot 2023-05-30 at 2 28 59 PM

### DIFF
--- a/ads/jobs/builders/infrastructure/dataflow.py
+++ b/ads/jobs/builders/infrastructure/dataflow.py
@@ -860,6 +860,18 @@ class DataFlow(Infrastructure):
             raise ValueError(
                 "Compartment id is required. Specify compartment id via 'with_compartment_id()'."
             )
+        if "executor_shape" not in payload:
+            payload["executor_shape"] = DEFAULT_SHAPE
+        if "driver_shape" not in payload:
+            payload["driver_shape"] = DEFAULT_SHAPE
+        executor_shape = payload["executor_shape"]
+        executor_shape_config = payload.get("executor_shape_config", {})
+        driver_shape = payload["driver_shape"]
+        driver_shape_config = payload.get("driver_shape_config", {})
+        if executor_shape != driver_shape:
+            raise ValueError("`executor_shape` and `driver_shape` must be from the same shape family.")
+        if (not executor_shape.endswith("Flex") and executor_shape_config) or (not driver_shape.endswith("Flex") and driver_shape_config):
+            raise ValueError("Shape config is not required for non flex shape from user end.")
         payload.pop("id", None)
         logger.debug(f"Creating a DataFlow Application with payload {payload}")
         self.df_app = DataFlowApp(**payload).create()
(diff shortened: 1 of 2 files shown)


ads/jobs/builders/infrastructure/dataflow.py
```diff
@@ -860,6 +860,18 @@ def create(self, runtime: DataFlowRuntime, **kwargs) -> "DataFlow":
...
+            payload["executor_shape"] = DEFAULT_SHAPE
+        if "driver_shape" not in payload:
+            payload["driver_shape"] = DEFAULT_SHAPE
+        executor_shape = payload["executor_shape"]
+        executor_shape_config = payload.get("executor_shape_config", {})
+        driver_shape = payload["driver_shape"]
+        driver_shape_config = payload.get("driver_shape_config", {})
+        if executor_shape != driver_shape:
+            raise ValueError("`executor_shape` and `driver_shape` must be from the same shape family.")
+        if (not executor_shape.endswith("Flex") and executor_shape_config) or (not driver_shape.endswith("Flex") and driver_shape_config):
```
mrDzurb, line 873: If so, maybe just remove shape config implicitly? Does the service throw exception if the config provided for the not flex shapes?

mrDzurb (dismissed):

ads/jobs/builders/infrastructure/dataflow.py
lu-ohai: This really depends on the service, for instance, model deployment will throw error, but for data flow, it will not. The issue here is, if adding shape config for non-flex shape, when call `create()`, the data flow will still propagate the shape config from user's input, instead of the default one from service. See below. The shape config for `VM.Standard2.1` should be `memoryInGBs: 15.0, ocpus: 1.0`, but these failed to be propagated after calling `create()`.
Screenshot 2023-05-31 at 11 02 49 AM

qiuosier: Executor shape and driver shape has to be the same even for non flex shape? If that’s the case then why we allow users to set them separately?

lu-ohai: @qiuosier Thanks for pointing this out. It should be the same shape family, not necessarily the exact same shape. I've modified the validation logic by taking reference to how the back-end handles this case.

qiuosier (approved):

mingkang111 (approved):
LGTM!

ads/jobs/builders/infrastructure/dataflow.py
```diff
@@ -860,11 +867,39 @@ def create(self, runtime: DataFlowRuntime, **kwargs) -> "DataFlow":
...
+        self._validate_shapes(payload)
         payload.pop("id", None)
         logger.debug(f"Creating a DataFlow Application with payload {payload}")
         self.df_app = DataFlowApp(**payload).create()
         self.with_id(self.df_app.id)
         return self
+    
+    @staticmethod
+    def _validate_shapes(payload: Dict):
+        if "executor_shape" not in payload:
```
mrDzurb, line 879: NIT: We can use consts here, right  - ```DataFlow.CONST_EXECUTOR_SHAPE```

Also 
```
if "executor_shape" not in payload:
    payload["executor_shape"] = DEFAULT_SHAPE
```
can be replaced with
```
payload["executor_shape"] = payload.get("executor_shape",  DEFAULT_SHAPE)
```

PR #231, opened 2023-06-13
Added support for security token authentication

lu-ohai:
### Added support for security token authentication

- Added new authenticate way `security_token`
- Updated ADS SDK and OPCTL
- Updated user guide

### Example
```
import ads
ads.set_auth("security_token") # using ~/.oci/config and DEFAULT profile
ads.set_auth("security_token", profile="test_session") # using ~/.oci/config and test_session profile
ads.set_auth("security_token", oci_config_location="test_config_location", profile="test_session") # using test_config_location config and test_session profile

# use config dict
config = {
    "region": "us-ashburn-1",
    "key_file": "~/.oci/sessions/test_session/oci_api_key.pem",
    "security_token_file": "~/.oci/sessions/test_session/token",
}
ads.set_auth("security_token", config=config)

# use AuthContext 
with AuthContext(auth='security_token', profile="test_session"):
    DataScienceJob.from_id("<job_ocid>")

# create security token signer
singer = ads.auth.security_token()
singer = ads.auth.create_signer(auth_type="security_token")

# opctl
!ads opctl run --ocid <job_ocid> --auth security_token
!ads opctl run --ocid <job_ocid> --auth security_token --oci-profile test_session
```

### DIFF
--- a/ads/common/auth.py
+++ b/ads/common/auth.py
@@ -5,8 +5,10 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 
 import copy
+from datetime import datetime
 import os
 from dataclasses import dataclass
+import time
 from typing import Any, Callable, Dict, Optional, Union
 
 import ads.telemetry
@@ -32,6 +34,11 @@ SECURITY_TOKEN_REQUIRED = [
     "key_file", 
     "region"
 ]
+SECURITY_TOKEN_LEFT_TIME = 600
+
+
+class TokenExpiredError(Exception):  # pragma: no cover
+    pass
 
 
 class AuthType(str, metaclass=ExtendedEnumMeta):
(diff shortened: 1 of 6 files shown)


ads/common/auth.py
```diff
@@ -678,6 +762,197 @@ def create_signer(self) -> Dict:
...
+                raise ValueError(
+                    f"Parameter `{parameter}` must be provided for using `security_token` authentication."
+                )
+
+        self._validate_and_refresh_token(configuration)
+
+        return {
+            "config": configuration,
+            "signer": oci.auth.signers.SecurityTokenSigner(
+                token=self._read_security_token_file(configuration.get("security_token_file")),
```
mrDzurb, line 837: Will we support only config files, or user will be able to pass tokens via dictionary as well? We had similar improvement with api keys.

ads/common/auth.py
```diff
@@ -678,6 +762,197 @@ def create_signer(self) -> Dict:
...
+        """
+        security_token = self._read_security_token_file(configuration.get("security_token_file"))
+        security_token_container = oci.auth.security_token_container.SecurityTokenContainer(
+            session_key_supplier=None,
+            security_token=security_token
+        )
+
+        if not security_token_container.valid():
+            raise SecurityTokenError(
+                "Security token has expired. Call `oci session authenticate` to generate new session."
```
mrDzurb, line 863: Could you provide more details in the message. It is not clear what needs to be done to make it workable?

ads/common/auth.py
```diff
@@ -678,6 +762,197 @@ def create_signer(self) -> Dict:
...
+            )
+        
+        time_now = int(time.time())
+        time_expired = security_token_container.get_jwt()["exp"]
+        if time_now - time_expired < SECURITY_TOKEN_LEFT_TIME:
+            try:
+                self._refresh_security_token(configuration)
+            except Exception as ex:
+                logger.info("Failed to refresh security token. Error: {}".format(ex))
```
mrDzurb, line 872: Shouldn't be an error thrown here?

ads/common/utils.py
```diff
@@ -1599,3 +1601,58 @@ def is_path_exists(uri: str, auth: Optional[Dict] = None) -> bool:
     if fsspec.filesystem(path_scheme, **storage_options).exists(uri):
         return True
     return False
+
+def apply_user_only_access_permissions(path: str):
```
mrDzurb, line 1605: What is the difference with original method? Why do we need to copy it?

ads/common/auth.py
lu-ohai: Seems like if the config in line 835 is missing the file paths of token and private key, the service will throw invalid header error. I guess maybe the service currently doesn't support token/key content for token-based authentication.

ads/common/auth.py
lu-ohai: I was think refreshing is a good thing, but not a blocking thing since if we failed to refresh the token, user can still use the one that's going to expire.

ads/common/utils.py
lu-ohai: In the original implementation, every time the token is refreshed and stored back to the token file, the token file will be [granted the user only access permission](https://github.com/oracle/oci-cli/blob/master/src/oci_cli/cli_session.py#L193) by calling this method. The method is copied from package [oci-cli](https://github.com/oracle/oci-cli/blob/master/src/oci_cli/cli_util.py#L2555), which is not included in the default dependencies of ads.

mrDzurb (dismissed):

qiuosier (approved):

qiuosier (approved):

mrDzurb (approved):

PR #235, opened 2023-06-20
Query builder fix for Joins

yogesh266:
### Description

The problem arose when utilizing the ``select()`` function from the FeatureGroup instance. This method returns a Query interface that can be employed for joining operations. However, there was an issue with the query formation when executing ``fg1.select().join(fg2.select()).join(fg3.select())``. 

Specifically, it resulted in a query of the form 
```select * from fg1 join fg2 on fg1.x=fg2.x join fg3 on fg1.x=fg3.x```
 whereas the desired query should have been 
```select * from fg1 join fg2 on fg1.x=fg2.x join fg3 on fg2.x=fg3.x```

### Unit tests
TODO

### Integration tests
TODO

### CODE REVIEW

File: ads/feature_store/query/query.py [hunk 1]
```diff
@@ -109,6 +109,20 @@ def with_left_feature_group(self, feature_group) -> "Query":
         """
         return self.set_spec(self.CONST_LEFT_FEATURE_GROUP, feature_group)
 
+    def get_last_joined_feature_group(self):
```
KshitizLohia, line 112: Is it possible to include test cases for the left feature group? Additionally, I'm curious about how all the unit test cases for the left feature group are currently passing.

ads/feature_store/common/spark_session_singleton.py
```diff
@@ -87,6 +87,7 @@ def __init__(self, metastore_id: str = None):
             self.spark_session = spark_builder.getOrCreate()
 
         self.spark_session.conf.set("spark.sql.execution.arrow.pyspark.enabled", "true")
+        self.spark_session.sparkContext.setLogLevel("OFF")
```
KshitizLohia, line 90: 👍

KshitizLohia (approved):
LGTM

PR #250, opened 2023-07-07
Feature/automlx

prasankh:
* ODSC-44009
* Added feature for time series forecasting using the AutoMLx model.

### DIFF
--- a/ads/operators/forecast/automlx.py
+++ b/ads/operators/forecast/automlx.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*--
+
+# Copyright (c) 2023 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
+import logging
+import sys
+import pandas as pd
+import time
+import traceback
+import datapane as dp
+from ads.operators.forecast.utils import load_data_dict, _write_data
+
+logger = logging.getLogger(__name__)
+handler = logging.StreamHandler(sys.stdout)
+handler.setLevel(logging.INFO)
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+
+
+def operate(operator):
+    try:
+        import automl
+        from automl import init
+    except Exception as ex:
+        print(
+            "Please run `pip3 install \
+            --extra-index-url=https://artifacthub-phx.oci.oraclecorp.com/artifactory/api/pypi/automlx-pypi/simple \
+            automlx==23.2.1` to install the required dependencies for automlx."
+        )
+        logger.debug(ex)
+        logger.debug(traceback.format_exc())
+        exit()
+    operator = load_data_dict(operator)
+    full_data_dict = operator.full_data_dict
+    models = dict()
+    outputs = dict()
+    outputs_legacy = []
+    selected_models = dict()
+    date_column = operator.datetime_column['name']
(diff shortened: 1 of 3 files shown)


ahosler (approved):
This looks awesome!

ads/operators/forecast/automlx.py
```diff
@@ -0,0 +1,124 @@
...
+import sys
+import traceback
+
+import datapane as dp
+import numpy as np
+import pandas as pd
+from sktime.forecasting.model_selection import temporal_train_test_split
+from ads.operators.forecast.utils import load_data_dict, _write_data
+
+logger = logging.getLogger(__name__)
```
mrDzurb, line 17: I think it would be better to setup logger once somewhere, maybe in __init__.py file?

ads/operators/forecast/automlx.py
```diff
@@ -0,0 +1,124 @@
...
+handler = logging.StreamHandler(sys.stdout)
+handler.setLevel(logging.INFO)
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+
+
+def operate(operator):
+    try:
+        import automl
```
mrDzurb, line 27: The ```@runtime_dependency``` decorator can be used instead.

```
from ads.common.decorator.runtime_dependency import (
    runtime_dependency,
    OptionalDependency,
)

@runtime_dependency(module="automl", err_msg="Error details")
def operate(operator):
```

ads/operators/forecast/automlx.py
```diff
@@ -0,0 +1,124 @@
...
+        from automl import init
+    except Exception as ex:
+        print(
+            "Please run `pip3 install \
```
mrDzurb, line 31: For now this message is fine, but we need to add TODO action item, to fix a message in the future, before GA. TODO should contain a ticket number. Please add a corresponding ticket to the story.

ads/operators/forecast/automlx.py
```diff
@@ -0,0 +1,124 @@
...
+    full_data_dict = operator.full_data_dict
+    models = dict()
+    outputs = dict()
+    outputs_legacy = []
+    selected_models = dict()
+    n_algos_tuned = operator.model_kwargs.get('n_algos_tuned', 4)
+    date_column = operator.datetime_column['name']
+    horizon = operator.horizon["periods"]
+    for i, (target, df) in enumerate(full_data_dict.items()):
+        print("Running automl for {} at position {}".format(target, i))
```
mrDzurb, line 48: I think it would be better to use logger here.

ads/operators/forecast/automlx.py
prasankh: Done, used logger from util. init.py was causing circular dependency

ads/operators/forecast/automlx.py
prasankh: Updated the same

ads/operators/forecast/automlx.py
prasankh: Done

ads/operators/forecast/automlx.py
prasankh: Changed

ads/operators/forecast/automlx.py
mrDzurb: You mean this import would make a circular dependency?
```
from ads.opctl import logger
```

mrDzurb (approved):

ads/operators/forecast/automlx.py
prasankh: No, If I setup logger inside init file.

PR #254, opened 2023-07-11
Deprecated DatasetFactory class and refactored code.

lu-ohai:
###  Deprecated DatasetFactory class and refactored code

- Deprecated `DatasetFactory` in ads `2.8.8`
- Moved and refactored internally-used APIs from `DatasetFactory` to `helper` class

### DIFF
--- a/ads/dataset/helper.py
+++ b/ads/dataset/helper.py
@@ -15,7 +15,7 @@ import re
 from collections import defaultdict
 import inspect
 import importlib
-from typing import List, Union
+from typing import Callable, List, Tuple, Union
 import fsspec
 
 # from pandas.io.common import _compression_to_extension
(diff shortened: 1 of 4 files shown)


ads/dataset/helper.py
```diff
@@ -849,3 +851,602 @@ def get_target_type(target, sampled_df, **init_kwargs):
...
+            target_type=target_type,
+            positive_class=positive_class,
+            **init_kwargs,
+        )
+    raise ValueError(
+        f"Unable to identify problem type. Specify the data type of {target} using 'types'. "
+        f"For example, types = {{{target}: 'category'}}"
+    )
+
+def open(
```
qiuosier, line 977: Since this method is also being deprecated. Can we update the code calling this method to use pandas instead of keeping the this method here? If we remove any internal usage of this `open` method, I think we can also remove the `CustomFormatReaders` and related functions.

qiuosier (approved):

mingkang111 (approved):

PR #260, opened 2023-07-14
Interoperability with model catalog

najiyacl:
## Description

Taking the endpoint url from env variable than setting it via ads.set_auth 
generic_model.save() 
  add dataset details to model custom data if featurestore dataset present
  update dataset with model details.

 
Screenshot 2023-07-21 at 1 24 40 PM

## Test cases
Integration Tests
(https://teamcity.oci.oraclecorp.com/buildConfiguration/Odsc_FeatureStore_FeatureStoreAdsIntegrationTests_FeatureStoreAdsIntegrationTestsPython38/41058084?buildTab=overview&hideProblemsFromDependencies=false&hideTestsFromDependencies=false)


ads/model/generic_model.py
```diff
@@ -1825,6 +1826,7 @@ def save(
         remove_existing_artifact: Optional[bool] = True,
         model_version_set: Optional[Union[str, ModelVersionSet]] = None,
         version_label: Optional[str] = None,
+        featurestore_dataset=None,
```
KshitizLohia, line 1829: we can type cast it to `Optional[Datset]`

ads/feature_store/common/utils/utility.py
```diff
@@ -47,7 +49,7 @@
 
 
 def get_execution_engine_type(
-    data_frame: Union[DataFrame, pd.DataFrame]
+        data_frame: Union[DataFrame, pd.DataFrame]
```
KshitizLohia, line 52: are we using `black` for pre commit formatting?

ads/model/generic_model.py
najiyacl: circular import problem

ads/feature_store/dataset.py
```diff
@@ -518,7 +519,18 @@ def add_models(self, model_details: ModelDetails) -> "Dataset":
             for item in items:
                 model_details.items.append(item)
         self.with_model_details(model_details)
-        return self.update()
+        self.update()
+
+        self._update_from_oci_dataset_model(self.oci_dataset)
+        updated_model_details = self.model_details
+        if updated_model_details and updated_model_details.items and model_details:
```
KshitizLohia, line 526: what happens if `self.model_details` is None and there is a call to `add_models`

ads/feature_store/dataset.py
```diff
@@ -518,7 +519,18 @@ def add_models(self, model_details: ModelDetails) -> "Dataset":
...
+            for model_id in model_details.items:
+                if model_id not in updated_model_details["items"]:
```
KshitizLohia, line 528: why this check if the model will get updated once we invoke `self.update()`. also, this will silently fail

ads/feature_store/mixin/oci_feature_store.py
```diff
@@ -6,13 +6,19 @@
...
+import os
 
 
 class OCIFeatureStoreMixin(OCIModelMixin):
     @classmethod
     def init_client(
         cls, **kwargs
     ) -> oci.feature_store.feature_store_client.FeatureStoreClient:
+        # TODO: Getting the endpoint from authorizer
+        fs_service_endpoint = os.environ.get("OCI_FS_SERVICE_ENDPOINT")
```
KshitizLohia, line 18: we might need to also update all the notebook examples to state initialisation of environment variable `OCI_FS_SERVICE_ENDPOINT `

ads/feature_store/dataset.py
najiyacl: Need more clarification
as per the current implementation if add_models(xxx) is called it will get added to the existing list.if self.model_details is null also, the added one will go to API side.
The xtra code added is because, the in memory dataset object gets updated even if the model ids are not valid, to avoid ,I am taking the details from oci_dataset which has updated details.
Will discuss for more clarification

ads/feature_store/dataset.py
najiyacl: the property 'get' gets us the dictionary

ads/feature_store/mixin/oci_feature_store.py
najiyacl: The one more change which should be going is to update the url from the config file which Harsh is trying to populate if the env variable is missing.

ads/feature_store/dataset.py
```diff
@@ -516,9 +517,32 @@ def add_models(self, model_details: ModelDetails) -> "Dataset":
...
+                    model_details.items.append(item)
         self.with_model_details(model_details)
-        return self.update()
+        try:
+            return self.update()
+        except Exception as ex:
+            logger.error(
+                f"Dataset update Failed with : {type(ex)} with error message: {ex}"
+            )
+            if existing_model_details:
```
KshitizLohia, line 529: why are we updating self and adding `with_model_details` in case of exception?

najiyacl: to get dataset back in old state

najiyacl: fir eq:
step 1 : dataset.add_models(["invalid model","valid_model"]) will fail but the in memory dataset will be associated with these two models.
step 2 : dataset.add_model([""valid_model"]) again will fail because the in memory dataset will be associated with invalid one too in the Step1

KshitizLohia (dismissed):
please run integration tests and paste in the pr before merging the pr.

ads/feature_store/docs/source/dataset.rst
```diff
@@ -317,3 +317,16 @@ Below is an example of the output.
...
 .. figure:: figures/dataset_lineage.png
   :width: 400
+
+
+Add Model Details
+=================
+
+You can call the ``add_models()`` method of the Dataset instance to add model ids to dataset.
+The ``.add_models()`` method takes the following parameter:
+``ModelDetails()``  with which we can associate model ids.
```
KshitizLohia, line 327: - ``ModelDetails()``  with which we can associate model ids.

KshitizLohia (approved):

PR #274, opened 2023-07-31
to_viz interface implementation

guptadivyank:
Adds the to_viz interface implementation which plots various statistics as matplotlib table chart

### DIFF
--- a/ads/feature_store/statistics.py
+++ b/ads/feature_store/statistics.py
@@ -1,10 +1,8 @@
 import pandas as pd
-from typing import Dict
-from copy import deepcopy
 
 from ads.feature_store.response.response_builder import ResponseBuilder
-from ads.jobs.builders.base import Builder
-from ads.common import utils
+from ads.feature_store.common.utils import utility as fs_utils
+import json
 
 
 class Statistics(ResponseBuilder):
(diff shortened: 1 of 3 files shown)


guptadivyank:
Screenshot 2023-07-31 at 10 18 27 PM
Screenshot 2023-07-31 at 10 19 07 PM

ads/feature_store/statistics.py
```diff
@@ -22,25 +22,22 @@ def kind(self) -> str:
...
-            if 'value' in data and isinstance(data.get('value', 0), (float, int, bool)):
-                scaler_metrics.append((metric_name, data.get('value')))
-        return scaler_metrics
+            statistics_payload = json.loads(self.content)
+            categorical_features, numerical_features = fs_utils.extract_scaler_metrics(
+                statistics_payload, features
+            )
+
+            if categorical_features:
+                fs_utils.plot_table(categorical_features)
```
KshitizLohia, line 40: is it feasible to check if we could do plotting of visulisation using the grids instead of just doing the sys out in the tabular format? we can check if we could use `plotly subplots: make_subplots` or `matplotlib subplots` for the same.

mingkang111: Hi @qiuosier , do you remember any context of removing `plotly`? https://bitbucket.oci.oraclecorp.com/projects/ODSC/repos/advanced-ds/pull-requests/1323/overview

PR #277, opened 2023-08-04
ODSC-44564: Improving fs cli endpoint interface

harsh97:


ads/feature_store/mixin/oci_feature_store.py
```diff
@@ -4,20 +4,60 @@
...
+
+        default_kwargs:dict = cls._get_auth().get("client_kwargs",{})
+
+        fs_service_endpoint = kwargs.get(cls.FS_SERVICE_ENDPOINT, None) or kwargs.get(cls.SERVICE_ENDPOINT, None) or \
+                              default_kwargs.get(cls.FS_SERVICE_ENDPOINT, None)
+
+        if not fs_service_endpoint:
+            try:
+                mod_time = os.stat(user_fs_config_loc()).st_mtime
+                if mod_time > cls.__mod_time:
```
KshitizLohia, line 47: can you please elaborate what are we intending to do via this?

harsh97: we are trying to keep a time record of the 'modified time' last time we opened the file. If the modified time is still the same we just skip reading it for optimisation

tests/unitary/with_extras/feature_store/test_mixin_endpoints.py
```diff
@@ -0,0 +1,26 @@
...
+    ads.set_auth(auth=AuthType.API_KEY, client_kwargs={"fs_service_endpoint": TEST_URL})
+    client = OCIFeatureStoreMixin.init_client(fs_service_endpoint=TEST_URL)
+    assert(client.base_client.endpoint==f"{TEST_URL}/20230101")
+
+def test_manual_with_service_endpoint():
+    ads.set_auth(auth=AuthType.API_KEY, client_kwargs={"fs_service_endpoint": TEST_URL, "service_endpoint": "service.com"})
+    client = OCIFeatureStoreMixin.init_client(fs_service_endpoint=TEST_URL)
+    assert(client.base_client.endpoint==f"{TEST_URL}/20230101")
+
+def test_manual_with_service_endpoint():
```
najiyacl, line 17: looks same as above

ads/feature_store/mixin/oci_feature_store.py
```diff
@@ -4,20 +4,60 @@
...
+    SERVICE_ENDPOINT = "service_endpoint"
+
     @classmethod
     def init_client(
         cls, **kwargs
     ) -> oci.feature_store.feature_store_client.FeatureStoreClient:
-        # TODO: Getting the endpoint from authorizer
-        fs_service_endpoint = os.environ.get("OCI_FS_SERVICE_ENDPOINT")
+
+        default_kwargs:dict = cls._get_auth().get("client_kwargs",{})
```
KshitizLohia, line 39: i think all these are not correctly formatted. can you double check if this is formatted via black? for example `default_kwargs:dict` to `default_kwargs: dict`

ads/feature_store/mixin/oci_feature_store.py
```diff
@@ -4,20 +4,60 @@
...
+                mod_time = os.stat(user_fs_config_loc()).st_mtime
+                if mod_time > cls.__mod_time:
+                    with open(user_fs_config_loc()) as ccf:
+                        cls.__template = FsTemplate(yaml.load(ccf, Loader=loader))
+                    cls.__mod_time = mod_time
+            except NameError:
+                logger.info("%s", "Feature store configuration helpers are missing. "
+                            "Support for reading service endpoint from config file is disabled")
+            except FileNotFoundError:
+                pass
```
KshitizLohia, line 55: can we log a warning in this case instead of just eating the exception?

harsh97: sure...I will log it at info level since it is ok for the file to be not their

tests/unitary/with_extras/feature_store/test_mixin_endpoints.py
guptadivyank: The tests seems to be misleading. The intention is to use the fs endpoint overridden by OCIFeatureStoreMixin client treated with highest priority followed by one set using ads.set_auth and followed by config file. In above test case we are basically using the same end point during override. we can also add a test showcasing absence of fs_endpoint and we consuming service endpoint instead. Pls correct me if my understanding is wrong.

guptadivyank (approved):

PR #279, opened 2023-08-08
added as_of interface to the feature store

yogesh266:
added as_of interface to the feature store

### DIFF
--- a/ads/feature_store/dataset.py
+++ b/ads/feature_store/dataset.py
@@ -838,13 +838,13 @@ class Dataset(Builder):
     def as_of(
         self,
         version_number: int = None,
-        timestamp: datetime = None,
+        commit_timestamp: datetime = None,
     ):
         """preview the feature definition and return the response in dataframe.
 
         Parameters
         ----------
-        timestamp: datetime
+        commit_timestamp: datetime
             commit date time to preview in format yyyy-MM-dd or yyyy-MM-dd HH:mm:ss
             commit date time is maintained for every ingestion commit using delta lake
         version_number: int
@@ -858,11 +858,11 @@ class Dataset(Builder):
         """
         self.check_resource_materialization()
 
-        validate_delta_format_parameters(timestamp, version_number)
+        validate_delta_format_parameters(commit_timestamp, version_number)
         target_table = self.target_delta_table()
 
         return self.spark_engine.get_time_version_data(
-            target_table, version_number, timestamp
+            target_table, version_number, commit_timestamp
         )
 
     def profile(self):
--- a/ads/feature_store/dataset.py
+++ b/ads/feature_store/dataset.py
@@ -8,6 +8,8 @@ from typing import Dict, List, Union
 
 import pandas
 from great_expectations.core import ExpectationSuite
+
+from ads import deprecated
 from ads.common import utils
 from ads.common.oci_mixin import OCIModelMixin
 from ads.feature_store.common.enums import (
@@ -15,6 +17,7 @@ from ads.feature_store.common.enums import (
     ExpectationType,
     EntityType,
 )
+from ads.feature_store.common.exceptions import NotMaterializedError
 from ads.feature_store.common.utils.utility import (
     get_metastore_id,
     validate_delta_format_parameters,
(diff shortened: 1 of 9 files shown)


ads/feature_store/dataset.py
```diff
@@ -806,6 +828,43 @@ def preview(
...
+    def as_of(
+        self,
+        version_number: int = None,
+        timestamp: datetime = None,
+    ):
+        """preview the feature definition and return the response in dataframe.
+
+        Parameters
+        ----------
+        timestamp: datetime
```
KshitizLohia, line 847: please change to `commit_timestamp`. btw, this is awesome

ads/feature_store/feature_group.py
```diff
@@ -1020,10 +1022,41 @@ def preview(
...
         sql_query = f"select * from {target_table} LIMIT {row_count}"
 
         return self.spark_engine.sql(sql_query)
 
+    def as_of(
+        self,
+        version_number: int = None,
+        timestamp: datetime = None,
+    ):
+        """preview the feature definition and return the response in dataframe.
```
KshitizLohia, line 1035: please add `as_of` for dataset also.

yogesh266: Done

ads/feature_store/dataset.py
yogesh266: Done

ads/feature_store/feature_group.py
```diff
@@ -1020,10 +1022,41 @@ def preview(
 
         if version_number is not None:
             logger.warning("Time travel queries are not supported in current version")
+
         sql_query = f"select * from {target_table} LIMIT {row_count}"
 
         return self.spark_engine.sql(sql_query)
 
+    def as_of(
```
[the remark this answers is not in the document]
yogesh266: Done

yogesh266:
Screenshot 2023-08-09 at 5 53 25 PM
Screenshot 2023-08-09 at 5 54 32 PM

ads/feature_store/execution_strategy/engine/spark_engine.py
```diff
@@ -36,6 +37,71 @@ def __init__(self, metastore_id: str = None, spark_session: SparkSession = None)
         else:
             self.spark = SparkSessionSingleton(metastore_id).get_spark_session()
 
+        self.managed_table_location = (
+            SparkSessionSingleton().get_managed_table_location()
```
KshitizLohia, line 41: if we use `SparkSessionSingleton()` without metastore id, this could result in spark session not taking metastore id into account which could result in singleton spark session without any connection to hive metastore. can you double check if this situation will not arise and can you add integration tests for as_of functionality with
1. error in case of timestamp and version both
2. pass in case of timestamp only
3. pass in case of version only

ads/feature_store/docs/source/dataset.rst
```diff
@@ -245,21 +245,20 @@ You can call the ``get_input_schema_dataframe()`` method of the Dataset instance
   df = dataset.get_input_schema_dataframe()
   df.show()
 
-Preview
-========
+as_of
```
KshitizLohia, line 248: we should not remove the functions completely and should use https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated to marking any function deprecated

ads/feature_store/docs/source/feature_group.rst
```diff
@@ -283,21 +283,20 @@ Feature store provides an API similar to Pandas to join feature groups together
   # Filter feature group
   feature_group.filter(feature_group.col1 > 10).show()
 
-Preview
+as_of
```
KshitizLohia, line 286: same comment. please don't remove deprecated functions from the document

KshitizLohia:
we should refrain from merging all the prs till the new version of ads as this will update the artifact with the same version in artifactory which is used in conda pack.

ads/feature_store/dataset.py
```diff
@@ -806,6 +828,43 @@ def preview(
...
+    def check_resource_materialization(self):
+        """Checks whether the target Delta table for this resource has been materialized in Spark.
+        If the target Delta table doesn't exist, raises a NotMaterializedError with the type and name of this resource.
+        """
+        if not self.spark_engine.is_delta_table_exists(self.target_delta_table()):
+            raise NotMaterializedError(self.type, self.name)
+
+    def as_of(
+        self,
+        version_number: int = None,
```
najiyacl, line 840: what if both are not provided?

tests/integration/feature_store/test_base.py
```diff
@@ -22,7 +22,7 @@
 
 client_kwargs = dict(
     retry_strategy=oci.retry.NoneRetryStrategy,
-    service_endpoint=os.getenv("service_endpoint"),
+    service_endpoint="http://127.0.0.1:21000",
```
najiyacl, line 25: remove

ads/feature_store/dataset.py
yogesh266: it will take latest one.

tests/integration/feature_store/test_base.py
yogesh266: Yeah that i was using just for testing.

ads/feature_store/execution_strategy/engine/spark_engine.py
yogesh266: above case wont arise as it sets the spark_session already so it wont be an issue.

PR #285, opened 2023-08-10
Added auto option in schema.yaml

govarsha:
Added auto option in schema.yaml and added a function in utils.py to select arima if number of cols > 15, otherwise automlx

### DIFF
--- a/ads/opctl/operator/lowcode/forecast/utils.py
+++ b/ads/opctl/operator/lowcode/forecast/utils.py
@@ -294,6 +294,6 @@ def human_time_friendly(seconds):
     return ", ".join(accumulator)
 
 def select_auto_model(columns):
-    if len(columns) > 15:
+    if columns!=None and len(columns) > 15:
         return SupportedModels.Arima
     return SupportedModels.AutoMLX
--- a/ads/opctl/operator/lowcode/forecast/utils.py
+++ b/ads/opctl/operator/lowcode/forecast/utils.py
@@ -21,6 +21,7 @@ from sklearn.metrics import (
 )
 
 from ads.dataset.label_encoder import DataFrameLabelEncoder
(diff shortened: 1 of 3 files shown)


ahosler (approved):
Looks awesome! We will continue to update the auto-select method as we go.

prasankh (approved):

codeloop (approved):

ads/opctl/operator/lowcode/forecast/utils.py
```diff
@@ -21,6 +21,7 @@
 )
 
 from ads.dataset.label_encoder import DataFrameLabelEncoder
+from ads.opctl.operator.lowcode.forecast.const import SupportedModels
```
mrDzurb, line 24: I think here would be better to use relative imports. Something like this.
```
from .const import SupportedModels
```

ads/opctl/operator/lowcode/forecast/utils.py
```diff
@@ -291,3 +292,8 @@ def human_time_friendly(seconds):
             )
     accumulator.append("{} secs".format(round(seconds, 2)))
     return ", ".join(accumulator)
+
+def select_auto_model(columns):
```
mrDzurb, line 296: Could you add some description for this method? Also I would suggest to move ```15``` into constants.

ads/opctl/operator/lowcode/forecast/utils.py
govarsha: @mrDzurb Yeah I will change that.

ads/opctl/operator/lowcode/forecast/utils.py
govarsha: @mrDzurb I will move 15 into constants. Where should I add description? Should I just add a comment there?

PR #289, opened 2023-08-16
optional params fixes

prasankh:
Made `confidence_interval_width`, `test_data` and `output_directory` optional parameters


ahosler (approved):
Looks great! We can clean up the duplicate temp folder in a later PR.

PR #302, opened 2023-08-25
Added validations for forecast operator

prasankh:
[ODSC-46047](https://jira.oci.oraclecorp.com/browse/ODSC-46047)
**Added following validations** 
- Additional data should have future values for horizon.
- Dates in primary data should be subset of dates in additional data.

**Other changes** 
- Dropping extra data points from additional if additional data has more data points than requested in the horizon.
- Those series which are failing the validation criteria will be skipped and forecast will not be generated for such series. The same message will be conveyed in logs as warning messages. 
- Throws error If all the series are failing to meet the validation criteria.


ads/opctl/operator/lowcode/forecast/utils.py
```diff
@@ -108,18 +107,53 @@ def _clean_data(data, target_column, datetime_column, target_category_columns=No
     )
 
 
+def _validate_and_clean_data(cat, horizon, primary, additional):
```
mrDzurb, line 110: Let's start adding typing to the methods. We will need to refactor the other code as well. But at least for the new methods let's do it right now.

ads/opctl/operator/lowcode/forecast/utils.py
```diff
@@ -108,18 +107,53 @@ def _clean_data(data, target_column, datetime_column, target_category_columns=No
...
+    """
+    Functions checks compatibility between primary and additional dataframe for a category.
```
mrDzurb, line 112: NIT: ```Checks compatibility between primary and additional dataframe for a category.```

ads/opctl/operator/lowcode/forecast/utils.py
```diff
@@ -108,18 +107,53 @@ def _clean_data(data, target_column, datetime_column, target_category_columns=No
...
+
+    Parameters
+    ----------
+        cat : Category for which data is being validated.
```
mrDzurb, line 116: NIT: the extra space is not required. 
Here a good example: https://github.com/oracle/accelerated-data-science/blob/feature/forecasting/ads/opctl/operator/cmd.py#L224

ads/opctl/operator/lowcode/forecast/utils.py
```diff
@@ -108,18 +107,53 @@ def _clean_data(data, target_column, datetime_column, target_category_columns=No
...
+        horizon : horizon value for the forecast.
+        primary : primary dataframe.
+        additional : additional dataframe.
+
+    Returns
```
mrDzurb, line 121: Let's add typing for the return value as well.

ads/opctl/operator/lowcode/forecast/utils.py
```diff
@@ -108,18 +107,53 @@ def _clean_data(data, target_column, datetime_column, target_category_columns=No
...
+    -------
+        Updated primary and additional dataframe or None values if the validation criteria does not satisfy.
+    """
+    # Additional data should have future values for horizon
+    data_row_count = primary.shape[0]
+    data_add_row_count = additional.shape[0]
+    additional_surplus = data_add_row_count - horizon - data_row_count
+    if additional_surplus < 0:
```
mrDzurb, line 129: Can we do?
```
if additional_surplus <= 0:
    ....
    return None, None
```
in this case the else condition would not be required.

mrDzurb: It would be better to use the pre-commit hook to format the code.

ads/opctl/operator/lowcode/forecast/utils.py
```diff
@@ -158,14 +192,22 @@ def _build_indexed_datasets(
                 .set_index(datetime_column)
                 .fillna(0)
             )
+
+            valid_primary, valid_add = _validate_and_clean_data(cat, horizon, data_by_cat_clean, data_add_by_cat_clean)
+            if valid_primary is None:
+                invalid_categories.append(cat)
+                continue
```
mrDzurb, line 199: I would suggest to avoid using ```continue``` operator, it makes code less readable. Let's use if else statement instead?

ads/opctl/operator/lowcode/forecast/utils.py
prasankh: When additional_surplus is equal to zero the return values are not None. Additionally there are two conditions to check two different validations.

ads/opctl/operator/lowcode/forecast/utils.py
prasankh: Updated

ads/opctl/operator/lowcode/forecast/utils.py
prasankh: Done

ads/opctl/operator/lowcode/forecast/utils.py
prasankh: Done

codeloop (approved):

mrDzurb (approved):

ahosler (approved):

PR #309, opened 2023-08-29
Fixed generic model multiple deploy.

lu-ohai:
### Fixed generic model multiple deploy issue

- The `model_deployment` property of generic model is a place holder for properties. The default runtime of it is container. After user deploys a conda runtime model and call `deploy` again, the `model_deployment` property will be updated to conda runtime, and property error happens. Though this use case is rare, but we need to cover it.
- Need to cover the following unique properties that are only available for container runtime:
-- `image`
-- `cmd`
-- `entrypoint`
-- `image_digest`
-- `server_port`
-- `health_check_port`

### DIFF
--- a/ads/model/generic_model.py
+++ b/ads/model/generic_model.py
@@ -2151,7 +2151,7 @@ class GenericModel(MetadataMixin, Introspectable, EvaluatorMixin):
                 "log_id", None
             )
             or self.properties.deployment_predict_log_id,
-            deployment_image = existing_runtime.image
+            deployment_image = getattr(existing_runtime, "image", None)
             or self.properties.deployment_image,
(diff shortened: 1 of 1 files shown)


ads/model/generic_model.py
```diff
@@ -2229,16 +2229,24 @@ def deploy(
         runtime = None
         if self.properties.deployment_image:
```
mrDzurb, line 2230: I'm not clear what's going on here. In which case the condition ```if self.properties.deployment_image:``` will pass and ```existing_runtime``` will not be a docker? If I run deployment twice, why it start thinking that on the second time the container needs to be used?

mrDzurb (approved):

qiuosier (approved):

PR #316, opened 2023-09-04
small fix

govarsha:
List has been used as type for columns in select_auto_model. It was not imported, so added the import line.


ads/opctl/operator/lowcode/forecast/utils.py
prasankh, line 1: def select_auto_model(columns:List)->str:
This can be changed to 
def select_auto_model(columns: List[str]) -> str:

govarsha: Changed it @prasankh

prasankh (approved):

PR #318, opened 2023-09-05
ODSC-46458/add metrics.csv

govarsha:
https://jira.oci.oraclecorp.com/browse/ODSC-46458
Added code to store Holdout Data Evaluation Metrics as metrics.csv.
Added optional parameter index in _write_data so that index can also be written if needed.
In operator_config, 
1. metrics_filename is the file name in which metrics should be stored. Set as metrics.csv by default.
2. forecast_filename is the file name in which forecasts should be stored. Set as forecast.csv by default.
image


ahosler (approved):
Looks great! Thanks for cleaning up some of the hardcoded variables and my poor naming 😅

PR #333, opened 2023-09-15
Adds opctl environment validator decorator.

mrDzurb:
## Description
https://jira.oci.oraclecorp.com/browse/ODSC-47201
Adds opctl environment validator decorator.

A Python decorator to validate whether a OPCTL CLI command can be executed in the current environment. This decorator will help ensure that CLI commands are only executed when the necessary conditions are met within the environment. For instance the operator's image cannot be created in the NB session or OCI Jobs.

## Pre-commit
image

## Tests
image

### DIFF
--- a/ads/opctl/decorator/common.py
+++ b/ads/opctl/decorator/common.py
@@ -4,22 +4,73 @@
 # Copyright (c) 2023 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 
-from typing import Dict, Callable
 from functools import wraps
+from typing import Callable, Dict, List
+
+from ads.config import (
+    JOB_RUN_OCID,
+    NB_SESSION_OCID,
+    PIPELINE_RUN_OCID,
+    DATAFLOW_RUN_OCID,
+    MD_OCID,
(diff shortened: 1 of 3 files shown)


ads/opctl/decorator/common.py
```diff
@@ -26,6 +54,28 @@ def wrapper(*args, **kwargs) -> Dict:
...
 
 
+def validate_environment(func: callable) -> Callable:
+    """Validates whether an opctl command can be executed in the current environment."""
+
+    @wraps(func)
+    def wrapper(*args: List, **kwargs: Dict) -> Dict:
+        if any(
+            value
+            for value in (
```
qiuosier, line 64: Checking only the environment variable might not cause problem if user set the value manually in a supported environment. I think, since the root cause of the operation not being supported is that the environment does not support certain dependency, for example, docker. Maybe it is enough to tell users that since docker is not installed, the operation is not supported.

ahosler: Good point! User may not have set up dataflow, for example

mrDzurb:
In the next PR will add extra validation:
```
try:
    import docker
    docker.from_env().version()
except:
    raise OpctlEnvironmentError()
```

PR #334, opened 2023-09-15
Cleans up the operator's redundant functionality.

mrDzurb:
## Description
https://jira.oci.oraclecorp.com/browse/ODSC-47490

Cleans up the operator's redundant functionality.

## Pre-commit
image


z7ye (approved):

qiuosier:
The integration tests also need to be updated since some related to files deleted by this PR. Do we still want to support simulating operator run in OCI notebook session with local container? I think that's the reason we have pre-defined dockerfile for building the images.

PR #337, opened 2023-09-19
Moved geopandas file load out of constructor

jdesanto:
ADSDataset fails without geopandas installation because its parent class PandasDataset attempts to load a file using geopandas in the constructor. The file is only used in a plot method. This PR moves the geopandas load to the plot routine where it used and adds protection so that it will only be loaded once.

There are no functionality changes.

Jira: https://jira.oci.oraclecorp.com/browse/ODSC-47518

### DIFF
--- a/ads/dataset/sampled_dataset.py
+++ b/ads/dataset/sampled_dataset.py
@@ -47,13 +47,13 @@ from ads.common.decorator.runtime_dependency import (
     OptionalDependency,
 )
 
+NATURAL_EARTH_DATASET = "naturalearth_lowres"
 
 class PandasDataset(object):
     """
     This class provides APIs that can work on a sampled dataset.
     """
 
-    @runtime_dependency(module="geopandas", install_from=OptionalDependency.GEO)
     def __init__(
         self,
         sampled_df,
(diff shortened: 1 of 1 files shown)


darenr (approved):

ads/dataset/sampled_dataset.py
```diff
@@ -706,6 +703,12 @@ def _visualize_feature_distribution(self, html_widget):
                 gdf = geopandas.GeoDataFrame(
                     df, geometry=geopandas.points_from_xy(df["lon"], df["lat"])
                 )
+
+                if not hasattr(self, "world"):
```
mrDzurb, line 707: I think it would be better to initiate the ```self.world = None``` in the constructor. And instead of checking the  ```if not hasattr(self, "world"):```, check if the ```self.world``` is not None.  

**NIT:** I would also suggest to move the "naturalearth_lowres" to the constants.

Probably we also need to remove the ```@runtime_dependency``` from the init.
```
 @runtime_dependency(module="geopandas", install_from=OptionalDependency.GEO)
    def __init__(
```
It should be just
```
def __init__(
```

jdesanto: Done

darenr (approved):

liudmylaru (approved):

PR #338, opened 2023-09-20
Added reference for nb_session to avoid unreferenced variable error

jdesanto:
When DSCNotebookSession fails to return a notebook session object, the try/except block does not define nb_session. Subsequent reference to nb_session generates an UnboundedLocalError. Because this occurs in a function that just populates a default settings dictionary, nb_session should be None if DSCNotebookSession fails.

https://jira.oci.oraclecorp.com/browse/ODSC-47429

### DIFF
--- a/ads/model/deployment/model_deployment_infrastructure.py
+++ b/ads/model/deployment/model_deployment_infrastructure.py
@@ -221,6 +221,7 @@ class ModelDeploymentInfrastructure(Builder):
         defaults[self.CONST_BANDWIDTH_MBPS] = DEFAULT_BANDWIDTH_MBPS
         defaults[self.CONST_WEB_CONCURRENCY] = DEFAULT_WEB_CONCURRENCY
         defaults[self.CONST_REPLICA] = DEFAULT_REPLICA
+        nb_session = None
 
         if NB_SESSION_OCID:
             try:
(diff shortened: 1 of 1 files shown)


ads/model/deployment/model_deployment_infrastructure.py
```diff
@@ -231,6 +231,7 @@ def _load_default_properties(self) -> Dict:
                     f"session: {NB_SESSION_OCID}. {e}"
                 )
                 logger.debug(traceback.format_exc())
+                nb_session = None
```
mrDzurb, line 234: I think this fix might not work. In the next line, when we do
```
nb_config = (
                getattr(nb_session, "notebook_session_config_details", None)
                or getattr(nb_session, "notebook_session_configuration_details", None)
            )
```
it will fail in attempt to extract ```notebook_session_config_details``` from the nb_session.

I would suggest to do something like this:
```
 if NB_SESSION_OCID:
            nb_config = None
            try:
                nb_session = DSCNotebookSession.from_ocid(NB_SESSION_OCID)
                nb_config = (
                     getattr(nb_session, "notebook_session_config_details", None)
                     or getattr(nb_session, "notebook_session_configuration_details", None)
                )
            except Exception as e:
                logger.warning(
                    f"Error fetching details about Notebook "
                    f"session: {NB_SESSION_OCID}. {e}"
                )
                logger.debug(traceback.format_exc())
```

jdesanto: `>>> a = None
>>> x=getattr(a, "whatever", None)
>>> print(x)
None`

thats's what the `None` in the third place is handling.

jdesanto: The alternative that @darenr and I discussed is to throw a RuntimeError. That's a viable alternative. I chose this because the function is just building a dictionary.

mrDzurb (approved):

lu-ohai:
I met this issue and changed this already in my pr to fix the unit test, so this PR is probably is not needed: https://github.com/oracle/accelerated-data-science/pull/340/files#diff-d47d586daf78d03597a529a929f5a95eebb3c571e260e9aa717667b0811ec616R226  @mrDzurb @jdesanto

jdesanto:
Closing. Duplicate of @lu-ohai's PR

jdesanto:
Not needed

PR #343, opened 2023-09-22
updated release notes

KshitizLohia:


yogesh266 (approved):

harsh97 (changes requested):
We need to add new notes and not modify existing ones

PR #349, opened 2023-09-25
ODSC-46457/refactor_forecast.csv

govarsha:
https://jira.oci.oraclecorp.com/browse/ODSC-46457
Changed the forecast.csv such that it has the following format 
Prophet
image
image
Neural Prophet
image
image
Automlx
image
Arima
image

Arima and Automlx does not return fitted values, so kept them as empty for now.


codeloop (approved):

ahosler (approved):
Awesome, thank you!

PR #351, opened 2023-09-25
ODSC-47630/fix_datapane_failure_for_only_one_series

govarsha:
When the dataset only contains 1 series, datapane throws a select error because there is only 1 series to select among. Refactored the code to resolve it.
https://jira.oci.oraclecorp.com/browse/ODSC-47630


codeloop (approved):

ahosler (approved):
Good catch

PR #374, opened 2023-10-10
Re-arrange unitary tests workflows

liudmylaru:
### Description

Solving issue "no space left on device", which is seen in all recent PRs, when running unit tests.

It does not matter if I split tests into different workflows or not, each setup is running in its own runner (agent), and even though I played a little with caches, it is still too much we install. So I am going back to solution to clean up space with running sudo rm, which is suggested in here - https://github.com/actions/runner-images/issues/2840#issuecomment-790492173

And I still would like to re-arrange workflow files, making use of composite actions, which make it possible to remove duplicated steps. And another re-arrangement will make 2 workflows: first will run on py3.8 and generating conv report, second will run on 3.9 and 3.10 and no repot. This making workflow more readable and easy maintain, with less if's.

Jira: https://jira.oci.oraclecorp.com/browse/ODSC-48419


qiuosier (approved):

lu-ohai (approved):
LGTM Thanks!

PR #375, opened 2023-10-12
ODSC-47259 Local Explainer Class : Feature/forecast explain

codeloop:
Adds support for local explanation of the forecast horizon

Local explanations for a sample dataset:
image
- All the values are the same as the additional features have the same value in the forecast horizon, also the positive & negative values in the feature attribution represent the movement of the forecast value up/down from the mean forecast values.


ahosler (approved):
looks good. Let's get this merged and we can iterate and continue to build it up.

PR #387, opened 2023-10-26
[ODSC-48860/48861] Global & Local explainability for prophet model

codeloop:
- Adds Global & local explanations for the Prophet model
- Update the forecast reports to include the normalized model explainability

Global exp:
image

Local explanation per time step in forecast horizon:
image


mrDzurb (approved):

ahosler (approved):
Awesome work, thank you!

PR #388, opened 2023-10-26
create archive when env yaml is provided in publish command

mayoor:
# Description

To create and publish in one step, one could use `ads opctl conda publish` with `-f env.yaml` option. In the current implementation, we execute a `docker run` to create conda pack and a different `docker run` to export the conda pack as conda pack archive. It is noticed that in some environments, the manifest yaml is somehow not shown up upon mounting the folders in the second docker run which causes the conda pack creation to fail. 

In order to prevent this situation, we can combine the conda create and archive in the same `docker run` step. Since the create API is used idependently of publish as well, we need to provide a switch to turn on or off the conda pack archive creation. This PR is for combining the create and publish step into single docker run when user provides `-f env.yaml` with `ads opctl conda publish`

# Testing

To run test - 

* Create a conda env.yaml with following contents - 
```
channels:
  - defaults
  - conda-forge
dependencies:
  - main::scikit-learn
  - main::pip
  - pip:
    - oracle-ads[geo,notebook,text]
```
* Run `CONTAINER_NETWORK=host ads opctl conda publish -n "Testing create and publish" -f test-conda/env.yaml -o -p oic://<bucket>@<namespace>/test-pack-publish/v1--verbose -d --auth security_token --oci-profile xxxxxx`

image

### DIFF
--- a/ads/opctl/conda/pack.py
+++ b/ads/opctl/conda/pack.py
@@ -18,9 +18,9 @@ import conda_pack
 import yaml
 
 
-def main(pack_folder_path):
+def main(pack_folder_path, manifest_file=None):
     slug = os.path.basename(pack_folder_path)
-    manifest_path = glob.glob(os.path.join(pack_folder_path, "*_manifest.yaml"))[0]
+    manifest_path = manifest_file or glob.glob(os.path.join(pack_folder_path, "*_manifest.yaml"))[0]
     with open(manifest_path) as f:
         env = yaml.safe_load(f.read())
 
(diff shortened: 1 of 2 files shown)


ads/opctl/conda/pack.py
```diff
@@ -59,8 +62,10 @@ def main(pack_folder_path):
             raise RuntimeError(
                 "Error creating the pack file using `conda_pack.pack()`."
             )
+        print(f"Copy {pack_file} to {pack_folder_path}")
```
mrDzurb, line 65: Shouldn't we use logger instead?

mrDzurb (approved):

liudmylaru (dismissed):

ads/opctl/conda/pack.py
mayoor: Inside pack.py there is no logger so far. This runs inside the container.

mrDzurb (approved):

liudmylaru (approved):

PR #396, opened 2023-10-30
Feature/forecasting autots bug

ahosler:


mrDzurb:
The target branch needs to be changed to the ``feature/forecasting``?

PR #407, opened 2023-11-03
Feature/forecasting auto algorithm improvements

prasankh:
- Auto algorithm improvements [ODSC-49027](https://jira.oci.oraclecorp.com/browse/ODSC-49027)
- Made model parameter optional in schema.yaml and the default value is set to 'auto'.
- Introduced ForecastDatasets class and moved data reading part out of base_model class which enables data aware model selection mechanism


codeloop (approved):
LGTM

ahosler (approved):
Awesome work!

PR #410, opened 2023-11-03
ODSC-49028/sort by datetime col & ODSC-48265/training metrics mismatch

govarsha:
https://jira.oci.oraclecorp.com/browse/ODSC-49028
Added code to sort primary and additional data by datetime column
Added format parmater to the to_datetime function calls, so that it doesn't infer wrong format.
Added unit test for automlx on unsorted data

https://jira.oci.oraclecorp.com/browse/ODSC-48265
Added fix in evaluate_metrics function

### DIFF
--- a/ads/opctl/operator/lowcode/forecast/model/transformations.py
+++ b/ads/opctl/operator/lowcode/forecast/model/transformations.py
@@ -9,6 +9,7 @@ from ads.opctl import logger
 
 class Transformations:
     """A class which implements transformation for forecast operator"""
+
     def __init__(self, data, dataset_info):
         """
         Initializes the transformation.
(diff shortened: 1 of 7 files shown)


ads/opctl/operator/lowcode/forecast/model/transformations.py
```diff
@@ -69,10 +73,39 @@ def _outlier_treatment(self, df):
...
+            df : The Pandas DataFrame.
+
+        Returns
+        -------
+            A new Pandas DataFrame with sorted dates for each category
+        """
+        import pandas as pd
+
+        # Temporary column for sorting
+        df["tmp_col_for_sorting"] = pd.to_datetime(
```
codeloop, line 101: We can avoid the groupby, by just sorting by series_id, date col, this will be in place and should be faster.
`
df.sort_values([self.series_id_column, self.date_column], ascending=True, inplace=True)
`

codeloop (approved):

ahosler (approved):
Awesome work! Ran some benchmarks and integration tests locally 👍

PR #412, opened 2023-11-07
Bug Fix : Dropping target column if present in additional data

prasankh:
[ODSC-49029](https://jira.oci.oraclecorp.com/browse/ODSC-49029)


ahosler (approved):
Looks great!

PR #415, opened 2023-11-07
ODSC-49565/corrections in metrics calculation per horizon

govarsha:
https://jira.oci.oraclecorp.com/browse/ODSC-49565
1. When test data has more values than horizon, operator fails.
2. When there is a horizon value missing for all series in test data, then test dataframe wont have that horizon. In the present code, that missing horizon is not removed from outputs so there is mismatch in the metrics calculation per horizon i.e; mismatch in the horizon picked from outputs and test data.
Changes:
outputs and test data frames should have same dates. So remove the extra dates that are not present in both test data & outputs and ensure that the dates are sorted in both dataframes so that there is no mismatch in metrics calculation.


ahosler (commented):
Great work!

PR #421, opened 2023-11-13
Fixed model deployment failed return value

lu-ohai:
### Fixed model deployment failed return value

- Every time when a service error raises, log this error and return the model deployment object. The model deployment object can continue calling logging
- Raise error for model deployment that's not in `ACTIVE` state

### DIFF
--- a/ads/model/deployment/model_deployment.py
+++ b/ads/model/deployment/model_deployment.py
@@ -85,7 +85,7 @@ class LogNotConfiguredError(Exception):  # pragma: no cover
     pass
 
 
-class ModelDeploymentFailedError(Exception):  # pragma: no cover
+class ModelDeploymentPredictError(Exception):  # pragma: no cover
     pass
 
 
@@ -607,11 +607,6 @@ class ModelDeployment(Builder):
         -------
         ModelDeployment
            The instance of ModelDeployment.
-
-        Raises
-        ------
-        ModelDeploymentFailedError
-            If model deployment fails to deploy
         """
         create_model_deployment_details = (
             self._build_model_deployment_details()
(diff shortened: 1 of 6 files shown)


ads/model/deployment/model_deployment.py
```diff
@@ -607,31 +607,24 @@ def deploy(
...
+        try:
+            response = self.dsc_model_deployment.create(
+                create_model_deployment_details=create_model_deployment_details,
+                wait_for_completion=wait_for_completion,
+                max_wait_time=max_wait_time,
+                poll_interval=poll_interval,
             )
-
-        return self._update_from_oci_model(response)
+        except Exception as ex:
```
VipulMascarenhas, line 624: nit: can we raise specific exceptions instead of generic here?

VipulMascarenhas (commented):
reviewing PR to understand the codebase/changes. I have a question regarding the checks failing - is that expected or those are run/tested locally?

ads/model/deployment/model_deployment.py
lu-ohai: Yes, what I'm doing here is to include all type of possible exceptions, but if only limited types of exception will be raised, then we're good to change it to specific ones.

lu-ohai:
> reviewing PR to understand the codebase/changes. I have a question regarding the checks failing - is that expected or those are run/tested locally?

I think it's some issue related to the github: https://github.com/oracle/accelerated-data-science/actions/runs/6852181858/job/18630113404?pr=421

ads/model/service/oci_datascience_model_deployment.py
```diff
@@ -212,6 +212,8 @@ def activate(
                         f"Error while trying to activate model deployment: {self.id}"
                     )
                     raise e
```
mrDzurb, line 214: This doesn't make sense to me. Let's add just logging statements and make sure that this will not breake the logic, maybe some parts still rely on the rising error solution

ads/model/deployment/model_deployment.py
```diff
@@ -890,6 +890,12 @@ def predict(
             Prediction results.
 
         """
+        current_state = self.sync().lifecycle_state
+        if current_state != "ACTIVE":
```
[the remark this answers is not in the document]
lu-ohai: Updated

ads/model/generic_model.py
```diff
@@ -1574,7 +1574,10 @@ def from_model_deployment(
 
         current_state = model_deployment.state.name.upper()
         if current_state != ModelDeploymentState.ACTIVE.name:
-            raise NotActiveDeploymentError(current_state)
+            logger.warning(
```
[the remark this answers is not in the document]
lu-ohai: Updated

ads/model/service/oci_datascience_model_deployment.py
lu-ohai: Updated

liudmylaru (approved):

mrDzurb (approved):

PR #448, opened 2023-11-22
Odsc 50180/negative r2

ahosler:


lyudmil-pelov (approved):

darenr (approved):

.github/workflows/run-operators-unit-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+permissions:
+  contents: read
+
+# hack for https://github.com/actions/cache/issues/810#issuecomment-1222550359
+env:
+  SEGMENT_DOWNLOAD_TIMEOUT_MINS: 5
+
+jobs:
+  test:
+    name: python ${{ matrix.python-version }}, default_setup
```
liudmylaru, line 26: you can remove ", default_setup".

.github/workflows/run-operators-unit-tests.yml
```diff
@@ -0,0 +1,60 @@
+name: "[Py3.8][Py3.10] Operators Tests"
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "ads/opctl/operator/**"
+      - "**requirements.txt"
```
liudmylaru, line 8: add here also "test-requirements-operators.txt"

test-requirements.txt
```diff
@@ -7,5 +7,6 @@ pip
 pytest
 pytest-codecov
 pytest-xdist
-# darts
+darts
+plotly
```
liudmylaru, line 11: test-requirements.txt file should not include any optional dependencies for ads tests. This is pure  to install ads (only default dependencies) and some libraries for pytest itself, coverage and running tests in parallel. We should avoid adding extra dependencies here. This txt installed only when we want to run tests from "default_setup" folder. The idea behind - we want to check ads when no optional dependencies installed.

For tests which require extra dependencies we maintain dev-requirements.txt file and install libraries from it when running. And we run "with_extras" folder.

In your case, when you creating special test suite for operators, add all extra libraries into test-requirements-operators.txt

test-requirements-operators.txt
```diff
@@ -0,0 +1,31 @@
+-e .
+click
```
liudmylaru, line 2: to not duplicate all optional dependencies here, you can add -e ".[forecast]", which will install libraries from pyproject.toml "forecast" section

to not duplicate pip, pytest, pytest-codecov and other libraries for pytest iteslf, you can add "-r test-requirements.txt" here. So it will install all from test-requirements.txt.

This file can be smth like:

-r test-requirements.txt
-e ".[forecast]"
- other_library_which_not_listed_in_test-requirements.txt
- other_library_which_not_listed_in_forecast_section_of_pyproject.toml

test-requirements.txt
ahosler: Thank you!

.github/workflows/run-operators-unit-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+            "test-requirements-operators.txt"
+
+      - uses: ./.github/workflows/set-dummy-conf
+        name: "Test config setup"
+
+      - name: "Run Operators Tests"
+        timeout-minutes: 60
+        shell: bash
+        env:
+          NoDependency: True
```
liudmylaru, line 55: this env variable can be removed, it is used in some tests in unitary folder to mark them to be skipped. So line 53-54 can be removed

test-requirements-operators.txt
ahosler: I like that, very clean. Thank you 🙏

codeloop (approved):

PR #461, opened 2023-11-28
adding serializer for vector store

z7ye:


ads/llm/serialize.py
```diff
@@ -111,7 +236,7 @@ def _load(obj: Any) -> Any:
         # Legacy chain
         return load_chain_from_config(obj, **kwargs)
 
-    return _load(obj)
+    return lc_load(obj, secrets_map=secrets_map, valid_namespaces=valid_namespaces)
```
qiuosier, line 239: I think keeping `_load(obj)` should work for your case. Can you test it again?

z7ye: done

PR #462, opened 2023-11-29
BugFix/add reload option in GenericModel.save()

mingkang111:
Jira: https://jira.oci.oraclecorp.com/browse/ODSC-49939

# Description
When we save the model using `GenericModel.save` API, currently we reload the score.py. If the user tries to save the model from the location where the score.py cannot run error free, the user cannot save the model. This can be a problem if the score.py is independently tested in a compatible environment and we need to save the artifact to the model deploy using ADS APIs.

# Changed
* Provide an option to not reload score.py and assume that the user has provided correct score.py. Log a warning that score.py is not tested and it can lead to deployment error if the score.py has any errors.
* Add test for save(reload=False)

# User Experience
Default value set to True to be compatible with current user experience.
```
model.save(reload=False)
```


z7ye (approved):
LGTM

qiuosier (approved):

PR #457, opened 2023-11-27
Fix version of lightgbm to be <4.0.0

liudmylaru:
### Description

This PR fixes several issues:
1. `ValueError: initial_types can not be detected. Please directly pass initial_types.` This error observed in model/test_model_framework_sklearn_model.py and model/test_model_framework_lightgbm_model.py. Issue is coming from https://github.com/sdpython/mlprodict/issues/488. Suggesting to fix lightgbm version <4.0.0 and relax it later when when the official releases of skl2onnx (v1.16.0) and onnxmltools (1.11.3).
2. tests hang on 99% of completion and never ends. This issue is coming from hpo tests, when they run together with whole unitary folder, tests hangs. To fix - reorganizing for tests made to run hpo tests first, and then all other
3. Some operators tests leaving behind config values, which are messing up with default values for telemetry, which running after operators tests. Fix for the removing env variable EXTRA_USER_AGENT_INFO before running telemetry tests. Here is error:
```
FAILED tests/unitary/default_setup/telemetry/test_agent.py::TestUserAgent::test_user_agent_api_keys_using_test_profile - AssertionError: assert 'Oracle-ads/version=2.9.0#surface=WORKSTATION#api=Operator' == 'Oracle-ads/version=2.9.0#surface=WORKSTATION#api=UNKNOWN'
  - Oracle-ads/version=2.9.0#surface=WORKSTATION#api=UNKNOWN
  ?                                                  ---- ^^
  + Oracle-ads/version=2.9.0#surface=WORKSTATION#api=Operator
  ?                                                   ^^^^^^^
```

### What was done

- updated tests workflow ymls so that hpo tests runs first and after all other
- added limit for version of lightgbm < 4.0.0
- fixed telemetry tests removing EXTRA_USER_AGENT_INFO env variable before running them

PR #463, opened 2023-11-29
Fixed hanging unit tests

liudmylaru:
### Description

With failing unit tests there are 2 issues. First is coming from https://github.com/sdpython/mlprodict/issues/488. Suggesting to fix lightgbm version <4.0.0 and relax it later when when the official releases of skl2onnx (v1.16.0) and onnxmltools (1.11.3). 

Second issue is when running unitary tests folder thy hang for 1h and never finish. Most likely they are coming from hpo. I could not reproduce locally, because hpo tests running separately finishing ok. My suggestion is to make them run first, and after all other tests. For that I just renaming folders to not add specific loging in running several times pytests by portions, which will also require to update coverage calculation login. So renaming folders seems and easy solution.

Renaming folder triggered all tests run with pre-commit, some changes in formatting and fixed to copyright message will be in this PR.

### What was done

- fixed lightgbm version <4.0.0
- renamed folder with hpo tests, so that they run first
- renamed with_extras to all_optional_deps, so that hpo tests in it will run first. With this change simplified steps in GH action yamls
- updated action yaml files to utilize new name and reorg matrix values for readability
- added copyright to 3 files after pre-commit
- pre-commit modified ~20 files, 110 left unchanged


liudmylaru:
This PR https://github.com/oracle/accelerated-data-science/pull/457 will do insted

PR #488, opened 2023-12-08
ODSC-34939: Implement optional loading of artifacts in the GenericModel

VipulMascarenhas:
### Description
This code change provides option to explicitly avoid downloading large model artifact in case the user does not wish to do so. 
https://jira.oci.oraclecorp.com/browse/ODSC-34939

This PR adds a new class method  `download_artifact` that allows user to load the GenericModel object with or without downloading model artifact.  Previously, `GenericModel.from_model_catalog(ocid,...)` would call the `download_artifact` method from `DataScienceModel` to download the artifact and return a GenericModel object.  

This PR calls the `DataScienceModel.download_artifact` call within the `GenericModel.download_artifact` method so that it gives the user some flexibility to call `GenericModel.download_artifact(artifact_dir, bucket_uri, ...)` later if user would want to to modify artifacts and redeploy the model, or test model locally.

Note: We're naming this method as `GenericModel.download_artifact` instead of `GenericModel.load_artifact` to maintain naming consistency with the `DataSciencModel` class.

### Example Of Usage
The various scenarios are documented [here](https://confluence.oci.oraclecorp.com/display/ODSC/Expected+Model+Artifact+Loading+Scenarios).

First, all metadata will be loaded including model deployment details if model deployment ocid is passed
```
model = GenericModel.from_id(ocid="ocid", load_artifacts=False)
```
Next, the user can 
1. Update any model related metadata
2. Invoke the model deployment end-point

If user would want to to modify artifacts and redeploy the model,  or test model locally then they'll need to download the artifacts.  They can then call:
```
model.download_artifact()
```

Now the the user can call:
```
model.verify(X)
model.predict(X, local=True)
```

If the user want to download the artifact , then they can do:
```
model = GenericModel.form_id(ocid="ocid")
OR 
model = GenericModel.form_id(ocid="ocid", load_artifacts=True)
```

### Tests
Added a few tests to check for behavior of download_artifact.

image


ads/model/generic_model.py
```diff
@@ -2969,6 +2973,226 @@ def upload_artifact(
             if tmp_artifact_zip_path:
                 os.remove(tmp_artifact_zip_path)
 
+    @classmethod
```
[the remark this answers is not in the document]
VipulMascarenhas: this is where I wanted some feedback.  This can be an instance method if we use download_artifact() and we don't pass the id like Ziqun mentioned [here](https://confluence.oci.oraclecorp.com/display/ODSC/Expected+Model+Artifact+Loading+Scenarios?focusedCommentId=3627101255#comment-3627101255).  If artifacts were not downloaded the first time, and then the calling instance will invoke download_artifact with flag set to True, then should we return a new object or should the existing object be updated with the artifacts?

ads/model/generic_model.py
```diff
@@ -1385,6 +1397,117 @@ def from_model_artifact(
...
+            )
+
+        if not artifact_dir:
+            artifact_dir = self.artifact_dir
+        artifact_dir = _prepare_artifact_dir(artifact_dir)
+
+        target_dir = (
+            artifact_dir
+            if not ObjectStorageDetails.is_oci_path(artifact_dir)
+            else tempfile.mkdtemp()
```
mayoor, line 1452: @lu-ohai had a PR to ensure that TMPDIR is respected even in the notebook session scenario. Can we put the artifact_dir creation code in a common place and call it here?

ads/model/generic_model.py
```diff
@@ -1385,6 +1397,117 @@ def from_model_artifact(
...
+                "`model_id` is not available, load the GenericModel object first."
+            )
+
+        if not artifact_dir:
+            artifact_dir = self.artifact_dir
+        artifact_dir = _prepare_artifact_dir(artifact_dir)
+
+        target_dir = (
+            artifact_dir
+            if not ObjectStorageDetails.is_oci_path(artifact_dir)
```
mayoor, line 1451: If it is object storage uri, do we download into object storage? @mingkang111 , this is to support the spark use case right?

mingkang111: That's right!

ads/model/generic_model.py
```diff
@@ -1385,6 +1397,117 @@ def from_model_artifact(
...
+        model_artifact = ModelArtifact.from_uri(
+            uri=target_dir,
+            artifact_dir=artifact_dir,
+            model_file_name=self.model_file_name,
+            force_overwrite=force_overwrite,
+            auth=auth,
+            ignore_conda_error=self.ignore_conda_error,
+        )
+        self.dsc_model = dsc_model
+        self.local_copy_dir = model_artifact.local_copy_dir
```
mayoor, line 1474: Instead of making a copy of this attribute, can we directly use `self.model_artifact.local_copy_dir` wherever we are referring to?

ads/model/generic_model.py
```diff
@@ -1385,6 +1397,117 @@ def from_model_artifact(
...
+        self.model_artifact = model_artifact
+        self.reload_runtime_info()
+
+        self._summary_status.update_status(
+            detail="Generated score.py",
```
mayoor, line 1479: Let us define these strings as constant.

ads/model/generic_model.py
```diff
@@ -1385,6 +1397,117 @@ def from_model_artifact(
...
+            force_overwrite=force_overwrite,
+            auth=auth,
+            ignore_conda_error=self.ignore_conda_error,
+        )
+        self.dsc_model = dsc_model
+        self.local_copy_dir = model_artifact.local_copy_dir
+        self.model_artifact = model_artifact
+        self.reload_runtime_info()
+
+        self._summary_status.update_status(
```
mayoor, line 1478: We should probably define these as methods to define the status

ads/model/generic_model.py
VipulMascarenhas: not sure if this is intended design, but `self.local_copy_dir` is set first which is passed on the ModelArtifact() constructor while setting up the `self.model_artifact` object in prepare().  One model_artifact is available then we can refer to `self.model_artifact.local_copy_dir` directly.

ads/model/generic_model.py
VipulMascarenhas: yea, I found around 40+ occurrences similar to this, might be good to have something like `GenericModel.update_summary_status` and `GenericModel.update_summary_action`. Will take this one and the below comment to define these as string constants in a separate PR.

ads/model/generic_model.py
VipulMascarenhas: Called Lu's code to create temp dir inside _prepare_artifact_dir().

mingkang111:
Functionality test looks good to me. 
Screenshot 2023-12-19 at 10 53 16 PM

ads/model/generic_model.py
mingkang111: A quick fix is to define dedicated update methods for 4 `from_xx` methods to update status. For example, `_update_status_from_model_catalog()`, moving all status related into the function. This is example to hard-code the summary status for `from_model_catalog`. Having dynamic update method would be much better though.

ads/model/generic_model.py
mingkang111: When artifact_dir is object storage, we will download to local dir as local backup. This is for spark case.

ads/model/generic_model.py
```diff
@@ -1386,6 +1398,117 @@ def from_model_artifact(
...
+        dsc_model.download_artifact(
+            target_dir=target_dir,
+            force_overwrite=force_overwrite,
+            bucket_uri=bucket_uri,
+            remove_existing_artifact=remove_existing_artifact,
+            auth=auth,
+            region=kwargs.pop("region", None),
+            timeout=kwargs.pop("timeout", None),
+        )
+        model_artifact = ModelArtifact.from_uri(
```
mingkang111, line 1466: Can we call `from_model_artifact(...)` here? We can adjust `from_model_artifact` to be @class_or_instance_method . cc : @qiuosier  Please correct me if I am wrong with using this decorator. Then in `from_model_artifact` we check whether it is instance or class before calling `model = cls(...)`.

ads/model/generic_model.py
```diff
@@ -1459,14 +1585,60 @@ def from_model_catalog(
...
+                detail="Conducted Introspect Test",
+                status=ModelState.NOTAPPLICABLE.value,
+            )
+            result_model._summary_status.update_status(
+                detail="Uploaded artifact to model catalog",
+                status=ModelState.NOTAPPLICABLE.value,
+            )
+            result_model.dsc_model = dsc_model
+            return result_model
+
```
mingkang111, line 1641: Looks like we can replace the logic below with new added `download_artifact` method. Then it can be :
```
if load_artifact:
    return self.download_artifact(...)
else:
    result_model = cls(
                artifact_dir=artifact_dir,
                bucket_uri=bucket_uri,
                auth=auth,
                properties=properties,
                ignore_conda_error=ignore_conda_error,
                **kwargs,
            )
    # update status
    result_model.dsc_model = dsc_model
    return result_model
```

ads/model/generic_model.py
```diff
@@ -1714,6 +1890,7 @@ def from_id(
         bucket_uri: Optional[str] = None,
         remove_existing_artifact: Optional[bool] = True,
         ignore_conda_error: Optional[bool] = False,
+        load_artifact: Optional[bool] = True,
```
mingkang111, line 1893: Kindly suggest to change the name of the argument `load_artifact` to `download_artifact`. It might cause confusing with `reload_artifact` used in `verify()` or argument `reload` in ModelArtifact.  `load` sounds more like loading something into environment. 
May I know how do you think? cc: @mayoor @qiuosier

ads/model/generic_model.py
qiuosier: @mingkang111 I think you are right in terms of using the `@class_or_instance_method`. However, even though I added the `@class_or_instance_method` to ADS, I am not a big fan of it :) Because for people outside ADS, this may not be easy to read. I think in this case, we can do something like the following to reduce the code repetition:
```
def _set_model_artifact(...):
    # the repeated code here

@classmethod
def from_model_artifact(...):
    # exiting code
    ...
    model = cls(...)
    model._set_model_artifact(...)

def download_artifact(...):
...
    dsc_model = DataScienceModel.from_id(model_id)
    dsc_model.download_artifact(...
    )
    self._set_model_artifact(...)
...
```

ads/model/generic_model.py
qiuosier: yeah, I personally had the confusion. I was thinking `load_artifact` means "load the artifact from disk to memory"

ads/model/generic_model.py
VipulMascarenhas: talked to QQ about this - we decided that instead of making a `class_or_instance_method` and then use if-else inside the `from_model_artifact`, it is probably better for readability for future developers if we just make the repetitive code as a private method. Most of it is around updating summary status, I have a separate ticket to clean this up, will make the changes there.

VipulMascarenhas: sounds good! Most of the code repetitive code here is for updating summary status, I have a different [ticket](https://jira.oci.oraclecorp.com/browse/ODSC-51216) for it. Will clean it up in a different PR to avoid partial update/clean up now.

ads/model/generic_model.py
VipulMascarenhas: changed the flag to `download_artifact`.

ads/model/generic_model.py
VipulMascarenhas: Kinda agree with your thinking here, this is something I did initially when I had download_artifact as a class method, but calling download_artifact() as an instance method here can lead to some messy code overall. 

Since `from_model_catalog` is a class method, we'll have to create an instance of GenericModel first to call `download_artifact`. Even then, the code within `download_artifact` can't use `from_model_artifact` (unless we call like GenricModel.from_model_artifact()) as the latter is a class method and the former isn't. 
This requires a little more thought into refactoring the class methods which might be an overkill for this PR, I'd like to discuss with this with the team and make the changes in a separate PR.

ads/model/generic_model.py
VipulMascarenhas: tracking in a separate [ticket](https://jira.oci.oraclecorp.com/browse/ODSC-51216).

ads/model/generic_model.py
VipulMascarenhas: tracking in a separate [ticket](https://jira.oci.oraclecorp.com/browse/ODSC-51216).

mayoor (approved):

ads/model/generic_model.py
```diff
@@ -1386,6 +1398,117 @@ def from_model_artifact(
...
+        artifact_dir: Optional[str] = None,
+        auth: Optional[Dict] = None,
+        force_overwrite: Optional[bool] = False,
+        bucket_uri: Optional[str] = None,
+        remove_existing_artifact: Optional[bool] = True,
+        **kwargs,
+    ) -> "GenericModel":
+        """Downloads model artifacts from the model catalog.
+
+         Parameters
```
qiuosier, line 1412: Looks like the indentation is not right here. There is an extra leading space?

VipulMascarenhas: yea, nice catch. I noticed it just now, maybe black formatter also didn't catch it. Will update this.

qiuosier (approved):

PR #489, opened 2023-12-08
Modified Watch in Model Deployment

lu-ohai:
### Modified Watch in Model Deployment

- Set default logs to be both `access` and `predict` logs
- Kept streaming logs while model deployment is in `ACTIVE` state.

## Notebook
Screenshot 2023-12-07 at 10 48 55 PM
Screenshot 2023-12-07 at 10 49 05 PM

### DIFF
--- a/ads/model/deployment/model_deployment.py
+++ b/ads/model/deployment/model_deployment.py
@@ -64,8 +64,6 @@ DEFAULT_WORKFLOW_STEPS = 6
 DELETE_WORKFLOW_STEPS = 2
 DEACTIVATE_WORKFLOW_STEPS = 2
 DEFAULT_RETRYING_REQUEST_ATTEMPTS = 3
-TERMINAL_STATES = [State.ACTIVE, State.FAILED, State.DELETED, State.INACTIVE]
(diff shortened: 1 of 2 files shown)


ads/model/deployment/model_deployment.py
```diff
@@ -65,6 +65,7 @@
 DEACTIVATE_WORKFLOW_STEPS = 2
 DEFAULT_RETRYING_REQUEST_ATTEMPTS = 3
 TERMINAL_STATES = [State.ACTIVE, State.FAILED, State.DELETED, State.INACTIVE]
+WATCH_TERMINAL_STATES = [State.FAILED, State.DELETED, State.INACTIVE]
```
mrDzurb, line 68: I think it might be a bit confusing having many terminal states. Maybe it would be better to have two separate methods to check different stop conditions, one for the deployment operation by itself and another one for watch?

lu-ohai: Yes I can update this, but if we have two stop condition methods, we still need two terminal states parameters, or should we write those directly in the methods?

qiuosier (approved):

mrDzurb (approved):

PR #495, opened 2023-12-12
Update langchain/test_serializers.py to avoid modifying get_type_to_cls_dict

qiuosier:
The current way of saving the original LangChain's `get_type_to_cls_dict` method is not correct.
```
cls.original_type_to_cls_dict = deepcopy(llms.get_type_to_cls_dict())
```
Because `get_type_to_cls_dict` is a function but the code is saving as a dictionary.

After running this test, the subsequent calls to `get_type_to_cls_dict` will fail with
```
TypeError: 'dict' object is not callable
```

Depending on the order when running the test, some tests may appear to fail randomly due to this issue:
https://github.com/oracle/accelerated-data-science/actions/runs/7137179363/job/19453780865#step:8:3440
https://github.com/oracle/accelerated-data-science/actions/runs/7115504771/job/19371812840

Since the embedding and LLM models in this test are for serialization only without inference. There is no need to use the "fake" one. We can use the real embedding and LLM models to avoid modifying `get_type_to_cls_dict` in LangChain.

### DIFF
--- a/tests/unitary/with_extras/langchain/test_serialization.py
+++ b/tests/unitary/with_extras/langchain/test_serialization.py
@@ -6,6 +6,7 @@
 
 
 import os
+from copy import deepcopy
 from unittest import TestCase, mock, SkipTest
 
 from langchain.llms import Cohere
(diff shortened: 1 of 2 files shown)


lu-ohai (dismissed):

z7ye (approved):

z7ye (approved):

lu-ohai (approved):

mingkang111 (approved):
LGTM 💯

qiuosier:
Tests status are not updated :(

PR #496, opened 2023-12-12
ODSC-50649/raise ValueError when `inference_python_version` cannot be resolved

mingkang111:
# Description
JIRA: https://jira.oci.oraclecorp.com/browse/ODSC-50649
We currently showing the requirement of `inference_python_version` as warning message during `.prepare()`. However, user might not take any action as `.prepare()` can still work. Then `.deploy()` will fail because of this missing information.

This change aims to raise ValueError in `.prepare()` when this field cannot be resolved automatically, mandating `inference_python_version`. 

In addition, changed some warning message which could lead to misunderstanding.  For example, when user passes `inference_python_version` and the manifest of given `inference_conda_env` doesn't provide python version, the warning message will show as below:
Screenshot 2023-12-12 at 12 40 31 PM
Use debug level instead.

# User Experience
* `.prepare()` without `inference_python_version` and it cannot be resolved automatically:
Screenshot 2023-12-12 at 12 44 28 PM

* `.prepare()` without `inference_python_version` and it can be resolved automatically:
Screenshot 2023-12-12 at 12 38 53 PM

* `.prepare()` with `inference_python_version` 
Screenshot 2023-12-12 at 12 45 53 PM

### DIFF
--- a/ads/model/artifact.py
+++ b/ads/model/artifact.py
@@ -273,7 +273,8 @@ class ModelArtifact:
             == ""
         ):
             raise ValueError(
-                "Cannot automatically detect the inference python version. `inference_python_version` must be provided."
+                "Cannot automatically detect the inference python version. "
+                "`inference_python_version` must be provided."
             )
         runtime_file_path = os.path.join(self.artifact_dir, "runtime.yaml")
         if os.path.exists(runtime_file_path) and not force_overwrite:
--- a/ads/model/artifact.py
+++ b/ads/model/artifact.py
@@ -30,13 +30,13 @@ SCORE_VERSION = "1.0"
 ADS_VERSION = __version__
 
 
-class ArtifactNestedFolderError(Exception):   # pragma: no cover
+class ArtifactNestedFolderError(Exception):  # pragma: no cover
     def __init__(self, folder: str):
         self.folder = folder
         super().__init__("The required artifact files placed in a nested folder.")
 
 
-class ArtifactRequiredFilesError(Exception):   # pragma: no cover
+class ArtifactRequiredFilesError(Exception):  # pragma: no cover
     def __init__(self, required_files: Tuple[str]):
         super().__init__(
             "Not all required files presented in artifact folder. "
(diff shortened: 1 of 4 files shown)


ads/model/runtime/env_info.py
```diff
@@ -157,9 +157,13 @@ def from_path(cls, env_path: str, auth: dict = None) -> "EnvInfo":
...
                     python_version = metadata_json.get("python", None)
                     env_slug = metadata_json.get("slug", None)
+                    if not python_version:
+                        raise ValueError(
+                            f"The manifest metadata of {env_path} doesn't contains inforamtion for python version."
+                        )
                 except Exception as e:
-                    logging.warning(e)
-                    logging.warning(
+                    logging.debug(e)
```
mayoor, line 165: If we are raising the value error, we dont have to catch it immediately. We can let it surface to the user.

mingkang111: Hi @mayoor, we don't want to fail at the time of `populate_env_info`, so it originally just make as warning when `fetch_metadata_of_object()` fail and not raise it. The check here I added is for debugging purpose. `fetch_metadata_of_object` may not fail, but python_version is not always in the manifest. So if user want to debug, it will be nice to have this.

qiuosier (approved):

ads/model/artifact.py
```diff
@@ -272,8 +272,9 @@ def prepare_runtime_yaml(
             or runtime_info.model_deployment.inference_conda_env.inference_python_version.strip()
             == ""
         ):
-            warnings.warn(
-                "Cannot automatically detect the inference python version. `inference_python_version` must be provided."
+            raise ValueError(
```
VipulMascarenhas, line 275: nit:  Update "Raises" section in docstrings of prepare_runtime_yaml() function to account for `inference_python_version` as well.

VipulMascarenhas (approved):
lgtm 👍

z7ye (approved):

ads/model/artifact.py
mingkang111: Thank you @VipulMascarenhas , I updated the docstring in [here](https://github.com/oracle/accelerated-data-science/pull/502/commits/f91746c967672f336afbe50ba4784b8744b48b0e) to avoid requesting approval one more time.

PR #497, opened 2023-12-13
Fix UnboundLocalError when saving model to model catalog

qiuosier:
When saving model to model catalog using artifacts from object storage bucket, the `UnboundLocalError` may happen if the service did not create the work request in 3 second. This could be due to a networking delay. The error is caused by local variable `new_work_request_logs` being used before assignment. This PR fixes the issue by using it only it work request logs exist.


lu-ohai (approved):

mingkang111 (approved):
LGTM 💯

z7ye (approved):

PR #499, opened 2023-12-14
ODSC-50786/GenericModel: add reload options to `ModelArtifact.from_uri`

mingkang111:
# Description
JIRA: https://jira.oci.oraclecorp.com/browse/ODSC-50786
The current working environment might not fit for the model it trying to loading from MD or MC. reload's failure will crash the call.  In the case of loading MD using container runtime, the message to remind user to set `ignore_conda_error=True` appears after long waiting time for loading model artifacts. 

In fact, user don't need to load the model into environment if they're not going to invoke model locally.

# Change
Not load model into environment when calling `GenericModel.from_id` by default. This will not break current user experience as when user trying to invoke model locally by `verify()`, the `reload_artifact` parameter in `verify()` is True by default. So it will reload at the time of need.

# User Experience
Check out [here](https://confluence.oci.oraclecorp.com/x/TKqP2Q) for more details.
* By default, not reload model into environment
```
model = GenericModel.from_id(md_ocid)

# load model into environment when invoking verify()
model.verify()
```

* User wants to reload model into environment at the time of `from_id`
```
# load model into environment when invoking from_id with reload=True
model = GenericModel.from_id(md_ocid, reload=True)
```

### DIFF
--- a/ads/model/artifact.py
+++ b/ads/model/artifact.py
@@ -171,22 +171,17 @@ class ModelArtifact:
         self.ignore_conda_error = ignore_conda_error
         self.model = None
         self.auth = auth or authutil.default_signer()
-        try:
-            if reload and not ignore_conda_error:
-                self.reload()
-                # Extracts the model_file_name from the score.py.
-                if (
-                    not self.model_file_name
-                    and self.score
-                    and hasattr(self.score, "model_name")
-                    and self.score.model_name
-                ):
-                    self.model_file_name = self.score.model_name
-        except:
-            logger.error(
-                "Failing to load model into environemnt. Please set `reload=False` or `ignore_conda_error=True`."
-            )
-            raise
+
+        if reload and not ignore_conda_error:
+            self.reload()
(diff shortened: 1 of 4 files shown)


ads/model/artifact.py
```diff
@@ -412,10 +413,11 @@ def from_uri(
         cls,
         uri: str,
         artifact_dir: str,
-        model_file_name: str = None,
-        force_overwrite: Optional[bool] = False,
         auth: Optional[Dict] = None,
+        force_overwrite: Optional[bool] = False,
```
mrDzurb, line 417: We should avoid changing the position of the attributes, because someone might use it without named arguments, in this case their code will stop working after this update.

mingkang111: Thanks @mrDzurb . It has been fixed.

qiuosier (approved):

z7ye (approved):

VipulMascarenhas (approved):
lgtm 👍

PR #500, opened 2023-12-15
Improved Progress Bar in Model Deployment and Model

lu-ohai:
### Notebooks

- Create model deployment

Screenshot 2023-12-18 at 10 51 31 AM
Screenshot 2023-12-18 at 10 51 58 AM
Screenshot 2023-12-18 at 10 52 42 AM
Screenshot 2023-12-18 at 10 55 04 AM
Screenshot 2023-12-18 at 11 19 47 AM

- Deactivate model deployment

Screenshot 2023-12-18 at 11 11 15 AM

- Activate model deployment

Screenshot 2023-12-18 at 11 18 25 AM
Screenshot 2023-12-18 at 11 18 33 AM
Screenshot 2023-12-18 at 11 19 15 AM
Screenshot 2023-12-18 at 11 35 25 AM

- Delete model deployment

Screenshot 2023-12-18 at 11 35 48 AM


ads/common/work_request.py
```diff
@@ -0,0 +1,128 @@
...
+    ) -> None:
+        self.id = id
+        self._description = description
+        self._percentage = 0
+        self._status = None
+        super().__init__(config, signer, client_kwargs, **kwargs)
+        
+
+    def _sync(self):
+        work_request = self.client.get_work_request(self.id).data
```
mrDzurb, line 48: Don't we need to check if the operation was successful or not?

ads/common/work_request.py
```diff
@@ -0,0 +1,128 @@
...
+        ).data
+
+        self._percentage= work_request.percent_complete
+        self._status = work_request.status
+        self._description = work_request_logs[-1].message if work_request_logs else "Processing"
+
+    def watch(
+        self, 
+        progress_callback: Callable,
+        max_wait_time: int,
```
mrDzurb, line 60: We can specify some default values for the interval and wait time?

ads/common/work_request.py
```diff
@@ -0,0 +1,128 @@
...
+        poll_interval: int,
+    ):
+        previous_percent_complete = 0
+
+        start_time = time.time()
+        while self._percentage < 100:
+
+            seconds_since = time.time() - start_time
+            if max_wait_time > 0 and seconds_since >= max_wait_time:
+                logger.error(f"Max wait time ({max_wait_time} seconds) exceeded.")
```
mrDzurb, line 70: Maybe ```Exceeded max wait time of {max_wait_time} seconds.```

ads/common/work_request.py
```diff
@@ -0,0 +1,128 @@
...
+            previous_percent_complete = self._percentage
+            progress_callback(percent_change, self._description)
+
+            if self._status in WORK_REQUEST_STOP_STATE:
+                if self._status != oci.work_requests.models.WorkRequest.STATUS_SUCCEEDED:
+                    if self._description:
+                        raise Exception(self._description)
+                    else:
+                        raise Exception(
+                            "Error occurred in attempt to perform the operation. "
```
mrDzurb, line 91: Do we know how to check service logs? :) If we don't maybe we can point out user on where they can find more details?

ads/common/work_request.py
```diff
@@ -0,0 +1,128 @@
...
+                            "Check the service logs to get more details. "
+                        )
+                else:
+                    break
+
+        progress_callback(0, "Done")
```
mrDzurb, line 97: I think it would be better to use named parameters for the callback. 

```
progress_callback(progress=0, description="Done")
```

Shouldn't the progress be a 100?

ads/common/work_request.py
```diff
@@ -0,0 +1,128 @@
...
+            self.id
+        ).data
+
+        self._percentage= work_request.percent_complete
+        self._status = work_request.status
+        self._description = work_request_logs[-1].message if work_request_logs else "Processing"
+
+    def watch(
+        self, 
+        progress_callback: Callable,
```
mrDzurb, line 59: We should add more details here, which attributes allowed for the ```progress_callback```. Otherwise it would be hard to support this.

ads/common/work_request.py
```diff
@@ -0,0 +1,128 @@
...
+                        )
+                else:
+                    break
+
+        progress_callback(0, "Done")
+
+
+def wait_work_request(
+    id: str, 
+    progress_bar_description: str,
```
mrDzurb, line 102: Maybe we can have some default value for the description as well?

ads/common/work_request.py
```diff
@@ -0,0 +1,128 @@
...
+                            "Error occurred in attempt to perform the operation. "
+                            "Check the service logs to get more details. "
+                        )
+                else:
+                    break
+
+        progress_callback(0, "Done")
+
+
+def wait_work_request(
```
mrDzurb, line 100: Could you add more pydocs please, it would be easier to support this code in the future.

mrDzurb (commented):
This is great!

ads/common/work_request.py
lu-ohai: Yes, it has the logic in the `watch` method to catch the exception if the operation is unsuccessful. And if so, the logger will print out the error as warning and continue the loop.

ads/common/work_request.py
lu-ohai: This pr is WIP and I'll add those in the next commit

ads/common/work_request.py
lu-ohai: Will add

ads/common/work_request.py
lu-ohai: Will update

ads/common/work_request.py
lu-ohai: Will add the work request id here

ads/common/work_request.py
lu-ohai: The `progress` passed here is the increment, not the final accumulated number. Will update

ads/common/work_request.py
lu-ohai: Will add in the next commit

ads/common/work_request.py
lu-ohai: Will add

ads/common/work_request.py
```diff
@@ -0,0 +1,193 @@
...
+                            "Check the service logs to get more details. "
+                            f"Work request id: {self.id}."
+                        )
+                else:
+                    break
+
+        progress_callback(progress=0, description="Done")
+
+
+def wait_work_request(
```
qiuosier, line 147: Looks like `ADSWorkRequest` will always be used with this `wait_work_request` function? Can we make this function as a method of the `ADSWorkRequest`? So we can do something like:
```
ADSWorkRequest(id, description).wait()
```
Also, shall we call this `DataScienceWorkRequest`?

qiuosier (dismissed):

ads/common/work_request.py
mrDzurb: Oh in my original idea, I expected that we would improve the ```OCIWorkRequestMixin``` class, where the ```wait_for_work_request```  method should've used under the hood the ADSWorkRequest class. But agree, adding the ```wait``` into the ```ADSWorkRequest``` totally make sense for me. We can still have the

```class OCIWorkRequestMixin:
    """Mixin class containing methods related to OCI work request"""

    def wait_for_work_request(
        self,
        work_request_id: str,
        wait_for_state: Union[str, tuple],
        max_wait_seconds: int = None,
        wait_interval_seconds: int = None,
    ):
         ADSWorkRequest(id, description, ...).wait()
    
```

mrDzurb (dismissed):

qiuosier (approved):

mingkang111 (approved):
LGTM!

PR #505, opened 2023-12-20
[WIP] converting datapane to report_creator

ahosler:


pyproject.toml
```diff
@@ -176,6 +176,7 @@ forecast = [
   "sktime",
   "statsmodels",
   "plotly",
+  "report_creator @ git+https://github.com/darenr/report_creator.git",
```
mrDzurb, line 179: Why don't we make it part of the ADS codebase?

PR #515, opened 2024-01-04
Create dummy wrappers for AquaModel and AquaDeployment

VipulMascarenhas:
### Description

This code change creates a dummy wrapper class with a list of operations that need to be a part of the AquaModel and AquaDeployment class. 

The relevant tickets for this are [ODSC-51353](https://jira.oci.oraclecorp.com/browse/ODSC-51353) and [ODSC-51358](https://jira.oci.oraclecorp.com/browse/ODSC-51358).


ads/aqua/aqua_deployment.py
qiuosier, line 1: I have a minor suggestion for the naming. Since we have the `aqua` folder to contain the files, I would suggest we use `deployment.py` instead of `aqua_deployment.py`. This way we have less repetition when doing the import
```
from ads.aqua.deployment import AquaDeployment
```

VipulMascarenhas: yes, makes sense. Same thing for model as well, i.e. `model.py` instead of `aqua_model.py`?

ads/aqua/aqua_deployment.py
```diff
@@ -0,0 +1,50 @@
...
+class AquaDeployment(ModelDeployment):
+    """
+
+    Attributes
+    ----------
+
+    Methods
+    -------
+    create(self, **kwargs) -> "AquaDeployment"
+        Creates an instance of Aqua deployment.
```
lu-ohai, line 22: Think it would be better to describe it `Creates an instance of model deployment via Aqua`. Similar for other methods.

qiuosier (approved):

lu-ohai (approved):

mingkang111 (approved):
LGTM!

PR #517, opened 2024-01-08
Subsampling data for faster plot rendering

prasankh:
- Subsampling the data down to 5k rows when rendering the plot in the report. 
- Added annotation when using subsampling .
- This limit can be adjusted using RENDER_LIMIT constant.

Sample plot with rendering limit set to 500 

Screenshot 2024-01-08 at 10 49 43 AM


ahosler (approved):
This will immediately make our lives better 👍

PR #525, opened 2024-01-11
[Discarded] ODSC-50335: Pass by reference for model catalog

VipulMascarenhas:
### Description

The Model Store team is working on a feature that enables users to pass [Model Artifact by reference](https://confluence.oci.oraclecorp.com/pages/viewpage.action?pageId=3575450491#expand-Model) when saving their models. This eliminates the need to copy model artifacts from a user's Object Storage bucket to the System Model Store bucket. 
Relevant jira ticket: [ODSC-50335](https://jira.oci.oraclecorp.com/browse/ODSC-50335)

On the ADS side, we implement the following:
1. When user creates a model using GenericModel.save(), we expect them to set the model_by_reference flag. This will ensure that the custom metadata will be updated with the OSS details and will be passed on to the Model Store.
2. For loading the model, GenericModel.from_model_catalog() will be called, which will read the custom metadata and if it finds that the original model entry was created by reference, it will skip the download from the catalog, but will copy directly from the artifact uri created from parsing the custom metadata.

Note: 

- The model store team is working on an updated [design](https://confluence.oci.oraclecorp.com/pages/viewpage.action?spaceKey=ODSC&title=Model+by+Reference%3A+Design+summary#ModelbyReference:Designsummary-AppendixH-Objectversionbasedintegritychecks), which might change our implementation slightly. 
- parts of `remove_existing_artifact` can be deprecated, but we haven't included that in this PR as it needs more end to end testing.

### Testing
Testing has been done only through mocks/unit tests so far since the changes from the Model Store team are not in dev/int yet.


VipulMascarenhas:
Soliciting feedback on the approach in this PR, let me know if the changes make sense. End to end testing isn't complete yet since we're awaiting API to be made available for testing from the model store team.

PR #528, opened 2024-01-17
AQUA JupyterLab Extension Example

qiuosier:
This PR adds the example code for AQUA JupyterLab server extension.

- Separating the "App" and "Model". The models are data classes holding the properties. The App will be handling the CRUDL operations on models. 
- Each "App" has a corresponding "APIHandler" to handle API requests with Jupyterlab.

How to start the extension with Juptyer Lab? (You may need to install the ads code into your base
- Setup your local environment to run ADS and JupyterLab.
- Add the extension to Jupyter Lab: `jupyter serverextension enable --py ads.aqua.extension`
- Start Jupyter Lab `jupyter lab` or `jupyter lab --no-browser`.

If you are getting import error when starting jupyterlab, you may need to install ADS into your base environment. You can do this by:
- Clone ADS repo
- Checkout the branch `qq/aqua`
- Install ADS with dev mode so that you can easily update the code `pip install -e .`

Test the APIs. You can test the APIs with your browser once the `jupyter lab` is running. By default, JupyterLab uses `http://localhost:8888` as the base URL.

Here are some example API responses:

Getting a list of fine tuning jobs: `http://localhost:8888/aqua/finetuning?compartment_id=my_comparment`
```
{
    "data": [
        {
            "id": "ocid0.xxx",
            "compartment_id": "my_comparment",
            "project_id": null,
            "model_id": "ocid.model.xxx",
            "task": "fine-tuning"
        },
        {
            "id": "ocid1.xxx",
            "compartment_id": "my_comparment",
            "project_id": null,
            "model_id": "ocid.model.xxx",
            "task": "fine-tuning"
        },
        {
            "id": "ocid2.xxx",
            "compartment_id": "my_comparment",
            "project_id": null,
            "model_id": "ocid.model.xxx",
            "task": "fine-tuning"
        },
        {
            "id": "ocid3.xxx",
            "compartment_id": "my_comparment",
            "project_id": null,
            "model_id": "ocid.model.xxx",
            "task": "fine-tuning"
        },
        {
            "id": "ocid4.xxx",
            "compartment_id": "my_comparment",
            "project_id": null,
            "model_id": "ocid.model.xxx",
            "task": "fine-tuning"
        }
    ]
}
```

Missing compartment ID when getting a list of fine tuning jobs: `http://localhost:8888/aqua/finetuning`
Note that in this case, the 400 status code will be returned with the HTTP request.
```
{
    "message": "Missing argument compartment_id",
    "reason": null
}
```

Getting the details of a fine-tuning job: `http://localhost:8888/aqua/finetuning/123`
```
{
    "id": "123",
    "compartment_id": "ocid.compartment.xxx",
    "project_id": "ocid.project.xxx",
    "model_id": "ocid.model.xxx",
    "task": "fine-tuning",
    "dataset": "dummy",
    "epoch": 2
}
```


mrDzurb (approved):

lu-ohai (approved):

ads/aqua/extension/base_handler.py
```diff
@@ -0,0 +1,35 @@
...
+    def serialize(obj: Any):
+        """Serialize the object.
+        If the object is a dataclass, convert it to dictionary.
+        Otherwise, convert it to string.
+        """
+        if is_dataclass(obj):
+            return asdict(obj)
+        return str(obj)
+
+    def finish(self, payload, status_code: int = 200):
```
mrDzurb, line 26: Would it make sense for the payload to have some structure like this?
```
@dataclass(repr=True)
class Response:
    """Class representing server response.

    Attributes
    ----------
    status: str
        Response status. Supported values: ["success", "error"]
    code: (str, optional). Defaults to empty string.
        Response code.
    msg: (str, optional). Defaults to empty string.
        The response message.
    payload: (Dict, optional). Defaults  to None.
        The response payload.
    """

    status: str = None
    code: str = None
    msg: str = None
    payload: Dict = None
```

We have such structure in the ODSC CLI as well, and it was pretty convenient structure to use.

ads/aqua/extension/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+
+from .model_handler import AquaModelHandler
+from .job_handler import AquaFineTuneHandler
+
+
+def _jupyter_server_extension_paths():
+    return [{"module": "ads.aqua.extension"}]
+
+
+def load_jupyter_server_extension(serverapp):
```
mrDzurb, line 15: We can keep handler end-points in the handlers by itself, and here just plugin these handlers. 
Something like this
```
from jupyter_server.utils import url_path_join
from ads.aqua.extension.handler.playground import __handlers__ as __playground_handlers__

__handlers__ = __playground_handlers__


def load_jupyter_server_extension(nb_server_app):
    web_app = nb_server_app.web_app
    host_pattern = ".*$"
    route_pattern = url_path_join(web_app.settings["base_url"], "aqua")

    web_app.add_handlers(
        host_pattern,
        [(url_path_join(route_pattern, url), handler) for url, handler in __handlers__],
    )

```

where in playground handler can have
```
__handlers__ = [
    ("playground/session/?([^/]*)", SessionHandler),
    ("playground/thread/?([^/]*)", ThreadHandler),
    ("playground/message/?([^/]*)", MessageHandler),
]
```

mrDzurb (approved):

ads/aqua/extension/base_handler.py
mrDzurb: The reason is, that we might want to return our internal statuses or description for the failed operations. Let's say if the tunning failed, we want to help user to debug this problem.

ads/aqua/model.py
```diff
@@ -35,8 +47,13 @@ def __init__(self, **kwargs):
...
 
-    def deploy(self, **kwargs) -> "AquaModel":
-        pass
+    def get(self, model_id) -> "AquaModel":
+        """Gets the information of an Aqua model."""
+        return AquaModel(id=model_id, compartment_id="ocid1.compartment", project_id="ocid1.project")
 
-    def list(self, **kwargs) -> List["AquaModel"]:
-        pass
+    def list(self, compartment_id, project_id=None, **kwargs) -> List["AquaModelSummary"]:
```
mingkang111, line 54: Should we return List[dict] here?

mingkang111 (approved):

ads/aqua/base.py
```diff
@@ -0,0 +1,15 @@
...
+
+import oci
+from ads.common.auth import default_signer
+
+
+class AquaApp:
+    """Base Aqua App to contain common components."""
+
+    def __init__(self) -> None:
+        self.client = oci.data_science.DataScienceClient(**default_signer())
```
VipulMascarenhas, line 15: this makes sense, I wasn't sure what client to use hence used the Flask example to understand the flow.

ads/aqua/model.py
```diff
@@ -5,14 +5,26 @@
...
 
-class AquaModel(DataScienceModel):
-    """Represents an Aqua Model.
+
+@dataclass
+class AquaModel(AquaModelSummary):
+    """Represents an Aqua model."""
+    icon: str = None
+
+class AquaModelApp:
```
lu-ohai, line 26: The constructor calls `super()` but seems like `AquaModelApp` doesn't inherit any class

mingkang111: This will be handled in my pr later:
https://github.com/oracle/accelerated-data-science/compare/ODSC-51434/list_model?expand=1

mingkang111 (approved):

ads/aqua/model.py
qiuosier: The finish() method in the handler will serialize the data class. So we don't have to do it in each of the methods.

ads/aqua/extension/base_handler.py
qiuosier: Thanks @mrDzurb . I will let the actual implementation take care you this :)

PR #544, opened 2024-01-25
[WIP] Implemented get aqua model and list aqua model deployments.

lu-ohai:


ads/aqua/deployment.py
```diff
@@ -67,4 +85,38 @@ def suggest(self, **kwargs) -> Dict:
         pass
 
     def stats(self, **kwargs) -> Dict:
```
mayoor, line 87: Maybe we should make this also a dataclass? The UI anyway cannot process arbitrary json payload without knowing the structure of the payload.

ads/aqua/model.py
```diff
@@ -75,49 +74,52 @@ def create(self, **kwargs) -> "AquaModel":
...
+        """
+        try:
+            oci_model = self.client.get_model(model_id).data
+        except Exception as e:
+            # show opc-request-id and status code
+            logger.error(f"Failing to retreive model information. {e}")
+            return None
+        
+        if not self._if_show(oci_model):
+            logger.info(f"Target model {oci_model.id} is not AQUA model. Skipped showing.")
```
mayoor, line 97: This can be debug statement

ads/aqua/model.py
```diff
@@ -75,49 +74,52 @@ def create(self, **kwargs) -> "AquaModel":
...
-                "organization": "Meta AI",
-                "is_fine_tuned_model": False,
-                "model_card": model_card,
-            }
+            compartment_id=oci_model.compartment_id,
+            project_id=oci_model.project_id,
+            name=oci_model.display_name,
+            id=oci_model.id,
+            time_created=oci_model.time_created,
+            icon=self._read_file(f"{artifact_path}/{ICON_FILE_NAME}"),
```
mayoor, line 109: if `_read_file` throws path error, we should assign `None` to icon. We cannot fail the API if the icon is missing for any reason.

ads/aqua/deployment.py
lu-ohai: Just updated the API. The name was changed to `get` and the return value was changed to dataclass `AquaDeployment` as per @VipulMascarenhas 's clarification.

ads/aqua/model.py
lu-ohai: Updated

ads/aqua/model.py
lu-ohai: Thanks, just updated

ads/aqua/base.py
```diff
@@ -9,6 +9,8 @@
 import logging
 import oci
 from ads.common.auth import default_signer
+from ads.aqua.exception import AquaServiceError, AquaClientError
+from oci.excpetion import ServiceError, ClientError
```
lu-ohai, line 13: NIT: `oci.excpetions`

VipulMascarenhas (approved):
changes look good. We should discuss how kwargs should be used across APIs when Ming is back.

ads/aqua/base.py
```diff
@@ -47,7 +49,7 @@ def list_resource(
             ).data
 
             return items
-        except Exception as e:
-            # show opc-request-id and status code
-            logger.error(f"Failing to retreive resources in the given compartment. {e}")
-            return []
+        except ServiceError as se:
```
mrDzurb, line 52: I see these except statements in many places, shouldn't we move them into the decorator?

ads/aqua/deployment.py
```diff
@@ -65,25 +67,31 @@ def list(self, **kwargs) -> List["AquaDeployment"]:
...
+
+        results = []
+        for model_deployment in model_deployments:
+            aqua_service_model=(
+                model_deployment.freeform_tags.get(AQUA_SERVICE_MODEL, None)
+                if model_deployment.freeform_tags else None
+            )
+            if aqua_service_model:
+                results.append(
+                    AquaDeployment(
```
mrDzurb, line 83: NIT: I left this comment in some other PR, we can have something like:
```
AquaDeployment.from_oci_model_deploument(oci_model_deployment)
```
Similar for the other DataClasses?

ads/aqua/base.py
lu-ohai: Yes, this PR was a little behind Ming's PR and will be updated with the latest error handler/decorator is merged. Merging this is mainly for the UI team to initial test.

ads/aqua/deployment.py
lu-ohai: Got it, will update it together with the error handler/decorator.

PR #548, opened 2024-01-26
Aqua Exception Design

mingkang111:
# Description
Currently we pass the str of service payload to the client. We would like to pass more structure response to the client so that FE can build better error page. 

# Key points
* When encountering an error during the processing of a request, RequestHandler.write_error can be called to send a custom error response to the client. 
* `AquaError` will be raised in Aqua's APIs. Developer need to register certain type of Error in `ads.aqua.exception.py` before raising it in app method (define the status code). When we want to raise an AquaError, we need to provide user friendly error message and HTTPError's status code. If we have more details want to give to user, we can record it into service_payload.
* Use exception_handler decorator in handler to pick the different error up and generate json response.

# Sample Response
* ADS Error
```
{
    "status": 500,
    "message": "Internal Server Error",
    "service_payload": {},
    "reason": "SecurityTokenError: Security token has expired. Call `oci session authenticate` to generate new session.",
    "request_id": "e6b2270c-02a7-46cf-8686-e022dbdf5a50"
}
```
* AquaError
```
{
  "status": 400,
  "message": "Bad Request",
  "service_payload": {},
  "reason": "Target model ocid1.datasciencemodel.oc1.iad.<OCID> is not Aqua model.",
  "request_id": "e6b2270c-02a7-46cf-8686-xxxxxx"
}
```
* OCI ServiceError
Plz checkout confluence page for more details.


ads/aqua/extension/model_handler.py
```diff
@@ -3,32 +3,54 @@
...
     def get(self, model_id=""):
         """Handle GET request."""
-        if not model_id:
-            return self.list()
-        return self.read(model_id)
+        try:
+            if not model_id:
+                return self.list()
+            return self.read(model_id)
+        except AquaServiceError as service_error:
```
mrDzurb, line 21: We probably can have a decorator for this? To not duplicate the code?

mingkang111: Fixed. Thanks for the suggestion!

ads/aqua/exception.py
```diff
@@ -0,0 +1,138 @@
...
+    """Handles AquaError."""
+
+    def inner_function(*args, **kwargs):
+        try:
+            func(*args, **kwargs)
+        except AquaServiceError as service_error:
+            raise HTTPError(
+                service_error.status or 500,
+                service_error.message,
+                **service_error.to_payload(),
```
qiuosier, line 103: The payload has to be one positional args, not keyword args.
```suggestion
                service_error.to_payload(),
```

ads/aqua/exception.py
```diff
@@ -0,0 +1,138 @@
...
+            )
+        except AquaClientError as client_error:
+            raise HTTPError(
+                client_error.status or 400,
+                client_error.message,
+                **client_error.to_payload(),
```
qiuosier, line 109:
```suggestion
                client_error.to_payload(),
```

ads/aqua/base.py
```diff
@@ -6,11 +6,10 @@
 
 from typing import List
 
-import logging
 import oci
 from ads.common.auth import default_signer
-
-logger = logging.getLogger(__name__)
+from ads.aqua.exception import oci_exception_handler
+from ads.aqua.utils import get_logger
```
mrDzurb, line 12: ```from ads.aqua import logger```?

ads/aqua/base.py
```diff
@@ -19,7 +18,9 @@ class AquaApp:
     def __init__(self) -> None:
         self._auth = default_signer()
         self.client = oci.data_science.DataScienceClient(**self._auth)
+        self.logger = get_logger()
```
mrDzurb, line 21: Why do we need this?

ads/aqua/base.py
```diff
@@ -40,14 +41,9 @@ def list_resource(
...
-                list_func_ref,
-                **kwargs,
-            ).data
-
-            return items
-        except Exception as e:
-            # show opc-request-id and status code
-            logger.error(f"Failing to retreive resources in the given compartment. {e}")
-            return []
+        items = oci.pagination.list_call_get_all_results(
```
mrDzurb, line 44: NIT: ```return oci.pagination.list_call_get_all_results(```?

ads/aqua/exception.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2024 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
+"""Exception module."""
```
mrDzurb, line 6: Nice! :) Adding more details would be even bettter.

ads/aqua/exception.py
```diff
@@ -0,0 +1,138 @@
...
+
+    The base exception from which all exceptions raised by Aqua
+    will inherit.
+    """
+
+    def to_payload(self) -> dict:
+        """Builds error payload."""
+        return asdict(
+            ErrorPayload(
+                message=self.message,
```
mrDzurb, line 32: We expect that AquaError has: ```message```, ```reason``` and ```service_payload``` attributes, but it is not really clear from where we get them? Wouldn't it be better to add them into the constructor?Just copy from ```AquaClientError```.  We might expect that child classes will bring different attributes, such as client_payload or similar, in this case in base class we can put the common attributes.

Probably in this case we don't need to maintain the ErrorPayload class, instead we can directly convert the AquaError to dict, because that what we return eventually. However I'm not sure why do we need manually convert it to the dict, can we return ```ErrorPayload``` class as is and then who ever wants to convert it to dict will do it?

ads/aqua/exception.py
```diff
@@ -0,0 +1,138 @@
...
+        """Initializes an AquaServiceError.
+
+        Parameters
+        ----------
+        message: str
+            User friendly error message.
+        service_payload: dict
+            `oci.exceptions.ServiceError`.
+        status: int
+            status code
```
mrDzurb, line 54: Could you add more details, which status we expect here? Status from service?

ads/aqua/exception.py
```diff
@@ -0,0 +1,138 @@
...
+                client_error.message,
+                **client_error.to_payload(),
+            )
+        except Exception as internal_error:
+            raise HTTPError(500, str(internal_error))
+
+    return inner_function
+
+
+def oci_exception_handler(func):
```
mrDzurb, line 117: Not sure that I get the reason having two decorators: oci_exception_handler and exception_handler. 
Will we use the ```oci_exception_handler``` for the APP methods, and the  ```exception_handler``` in the handlers? When the app methods will be used in the handler, and app generates exception, then the ```exception_handler``` handler will pick the error up and generate HTTP error? If so, can we change the names of the decorators? Maybe: 

* ```handle_handler_exception```
* ```handle_application_exception```

I think it would also make sense to leave exception custom classes in the ```exception.py```, but move decorators into a separate place, maybe we should create ```decorator.py``` file next to the ```exception.py```.

ads/aqua/extension/base_handler.py
```diff
@@ -40,3 +43,31 @@ def finish(self, payload=None):  # pylint: disable=W0221
         # Convert the payload to a JSON serializable object
         payload = json.loads(json.dumps(payload, default=self.serialize))
         return super().finish(payload)
+
+    def write_error(self, status_code, **kwargs):
```
mrDzurb, line 47: Correct me if i'm wrong, but i'm expecting that this method will be called whenever we rise a error in the handler? I might be wrong, but why can't we accept as input parameter the ```AquaError``` object and convert it into the ```ErrorPayload``` here?

qiuosier:
I made some changes on the exception_handler decorator. Here are my suggestions:
* I think we can use a single decorator to handle all exceptions including OCI exceptions. 
* Whenever there is an error. The API will return either 4XX (client error) or 5XX (server error) code.
* The error payload will contain:
** message, this will be a text describing the error code. This is filled automatically by Jupyter
** reason, this is a user friendly message describing the actual error
** service_payload, this will be the payload returned by the service, if any. Or some other payload for debugging purpose.
** traceback, optionally we can return traceback.
* The error code will be consistent with whatever error code returned by OCI.
I put my example code here: https://github.com/oracle/accelerated-data-science/tree/qq/aqua_exception
See the `aqua.model.AquaModelApp.get()` for example on different errors.

ads/aqua/base.py
```diff
@@ -19,7 +18,9 @@ class AquaApp:
     def __init__(self) -> None:
         self._auth = default_signer()
         self.client = oci.data_science.DataScienceClient(**self._auth)
```
mayoor, line 20: We should use the client factory - ads.common.oci_client

mayoor:
Once all the review comments are incorporated, @mingkang111 , could you please let me know? I would like to look at the final code before it is merged. Thanks for bringing this all together!

ads/aqua/extension/base_handler.py
mingkang111: You're right! The design has been modified.

ads/aqua/base.py
mingkang111: Removed. Using logger in __init__.py.

ads/aqua/base.py
mingkang111: Fixed.

ads/aqua/exception.py
mingkang111: `ErrorPayload` is no longer needed.

ads/aqua/exception.py
mingkang111: Originally, I want to group all errors into two type of errors: AquaServiceError and AquaClientError. So we need `oci_exception_handler` to convert oci exception into AquaError. 

In the latest design, the oci error will not be converted to AquaError. We handle it directly when we catch it.

mingkang111:
> I played with the exception_handler decorator. Here are my suggestions:
> 
> * I think we can use a single decorator to handle all exceptions including OCI exceptions.
> * Whenever there is an error. The API will return either 4XX (client error) or 5XX (server error) code.
> * The error payload will contain:
>   ** message, this will be a text describing the error code. This is filled automatically by Jupyter
>   ** reason, this is a user friendly message describing the actual error
>   ** service_payload, this will be the payload returned by the service, if any. Or some other payload for debugging purpose.
>   ** traceback, optionally we can return traceback.
> * The error code will be consistent with whatever error code returned by OCI.
> 
> I put my example code here: https://github.com/oracle/accelerated-data-science/tree/qq/aqua_exception See the `aqua.model.AquaModelApp.get()` for example on different errors.

Thanks QQ! I updated the design based on your suggestions.

mingkang111:
> Once all the review comments are incorporated, @mingkang111 , could you please let me know? I would like to look at the final code before it is merged. Thanks for bringing this all together!

Hi @mayoor , the comments above are all addressed. Could you please take a look when you have time? Please checkout the confluence page shared in team channel for more sample response. Thank you!

ads/aqua/decorator.py
```diff
@@ -0,0 +1,75 @@
...
+    MultipartUploadError,
+    RequestException,
+    ServiceError,
+)
+
+from ads.aqua.exception import AquaError
+from ads.aqua.extension.base_handler import AquaAPIhandler
+
+
+def exception_handler(func):
```
mrDzurb, line 25: NIT: Wouldn't be something ```handle_exceptions``` better name for this decorator? Could you also add in description more details when this decorator can be used?

ads/aqua/deployment.py
```diff
@@ -339,74 +341,67 @@ def get(self, model_deployment_id: str, **kwargs) -> "AquaDeployment":
...
         Dict:
             A dict of allowed deployment configs.
         """
         try:
-            oci_model = self.ds_client.get_model(
-                model_id
-            ).data
-        except Exception as se:
+            oci_model = self.ds_client.get_model(model_id).data
+        except Exception as ex:
```
mrDzurb, line 379: This will be cleaned up?

mingkang111: Yes. When this design is being populated into current implementation, it will be cleaned up/revisit.

mayoor:
Thanks @mingkang111. Let us merge the PR.

mayoor:
> Thanks @mingkang111. Let us merge the PR.

We need to drop the traceback from the payload and replace it with requestId. For now let the requestid be random number. We will finalize the details in a couple of days

ads/aqua/decorator.py
mingkang111: Renamed and added example of usage.

mingkang111:
> > Thanks @mingkang111. Let us merge the PR.
> 
> We need to drop the traceback from the payload and replace it with requestId. For now let the requestid be random number. We will finalize the details in a couple of days

Replaced the `traceback` field with `request_id`.

VipulMascarenhas (approved):

mayoor (approved):

PR #551, opened 2024-01-27
[ODSC 51354] Create new Aqua deployment

VipulMascarenhas:
### Description

Jira ticket: [ODSC-51354](https://jira.oci.oraclecorp.com/browse/ODSC-51354)

The code changes in this PR an Aqua model deployment. Following changes are done:
1. Creating a new handler called `deployment_handler.py`. The handler accepts post request and there are a few required parameters that caller will have to provide (need comments on what can be optional).
2. In `deployment.py`, implemented create() functionality. It does not create a catalog entry right now (needs AquaModel.create implemented - either mock or otherwise), so assumes a few attributes will be available (image, model id, entrypoint) but this will be removed. (need comments on what deployment config needs to be hardcoded vs. expected from caller).
3.  Added Errors class in base_handler, but this will be removed once Error handler PR goes though.
4. Added support for serializing dataclass using `ads.common.serializer.DataClassSerializable` for CLI commands.

User can invoke create by calling:

```
curl -X POST -H "Content-Type: application/json" -d '{"compartment_id": "ocid1.compartment.oc1..xxx", "project_id": "ocid1.datascienceproject.oc1.iad.xxx", "display_name": "Test deploy via extensions", "instance_count": 1, "instance_shape": "VM.Standard.E4.Flex", "log_group_id": "ocid1.loggroup.oc1.iad.xxx", "access_log_id": "ocid1.log.oc1.iad.xxx", "predict_log_id": "ocid1.log.oc1.iad.xxx", "model_id": "ocid1.datasciencemodel.oc1.iad.xxx"}' http://localhost:8888/aqua/deployments
```

It should return a json response:
```
{
    "display_name": "Test deploy via extensions",
    "aqua_service_model": "xxx.xxx.xxx.xxx.xxx#aqua_service_model_value",
    "state": "CREATING",
    "description": null,
    "created_on": "2024-01-27 01:46:06.593000+00:00",
    "created_by": "ocid1.user.oc1..xxx",
    "endpoint": "https://modeldeployment.us-ashburn-1.oci.customer-oci.com/ocid1.datasciencemodeldeployment.oc1.iad.xxx"
}
```

or, user can use cli:
```
ads aqua deployment create --compartment_id "ocid1.compartment.oc1..xxx" --project_id "ocid1.datascienceproject.oc1.iad.xxx" --display_name "Test aqua deploy via CLI" --instance_count 1 --instance_shape "VM.Standard.E4.Flex" --model_id "ocid1.datasciencemodel.oc1.iad.xxx" --deployment_image "iad.ocir.io/ociodscdev/aqua_deploy:1.0.0" --entrypoint '["python", "/opt/api/api.py"]' --aqua_service_model "xxx.xxx.xxx.xxx.xxx#aqua_service_model_value" --log_group_id "ocid1.loggroup.oc1.iad.xxx" --access_log_id "ocid1.log.oc1.iad.xxx" --predict_log_id "ocid1.log.oc1.iad.xxx"
```

Next steps:
- replacing hardcoded values with aqua model entries 
- tests


ads/aqua/deployment.py
```diff
@@ -42,16 +48,141 @@ class AquaDeploymentApp(AquaApp):
...
+        # configure model deployment runtime
+        # todo : any other runtime params needed?
+        container_runtime = (
+            ModelDeploymentContainerRuntime()
+            .with_image(deployment_image)
+            .with_entrypoint(entrypoint)
+            .with_server_port(server_port)
+            .with_health_check_port(health_check_port)
+            .with_env(env_var)
+            .with_deployment_mode("HTTPS_ONLY")
```
lu-ohai, line 160: Think we can use the constant `ModelDeploymentMode.HTTPS` here.

ads/aqua/deployment.py
```diff
@@ -116,23 +241,28 @@ def get(self, **kwargs) -> "AquaDeployment":
...
-        if not kwargs.get("model_deployment_id", None):
-            raise AquaClientError("Aqua model deployment ocid must be provided to fetch the deployment.")
-        
+        if not model_deployment_id:
+            raise AquaClientError(
+                "Aqua model deployment ocid must be provided to fetch the deployment."
+            )
+
         try:
             model_deployment = self.client.get_model_deployment(**kwargs).data
```
lu-ohai, line 250: If `model_deployment_id` is not passed with `kwargs`, then it should be passed to `get_model_deployment` specifically:
```
self.client.get_model_deployment(model_deployment_id=model_deployment_id, **kwargs).data
```

VipulMascarenhas: ah thanks for noticing this. Might have mixed this up while fixing merge issue.

ads/aqua/deployment.py
```diff
@@ -69,14 +200,17 @@ def list(self, **kwargs) -> List["AquaDeployment"]:
         """
         compartment_id = kwargs.get("compartment_id", None)
```
mrDzurb, line 201: NIT: We can pop compartment from kwargs and explicitly pass into the list_resource?

ads/aqua/deployment.py
```diff
@@ -69,14 +200,17 @@ def list(self, **kwargs) -> List["AquaDeployment"]:
...
         for model_deployment in model_deployments:
-            aqua_service_model=(
+            aqua_service_model = (
                 model_deployment.freeform_tags.get(AQUA_SERVICE_MODEL, None)
-                if model_deployment.freeform_tags else None
+                if model_deployment.freeform_tags
+                else None
             )
             if aqua_service_model:
                 results.append(
```
mrDzurb, line 216: We can have something like
```
AquaDeployment.from_oci_model(...)
```

ads/aqua/deployment.py
```diff
@@ -87,22 +221,13 @@ def list(self, **kwargs) -> List["AquaDeployment"]:
...
-        pass
-
-    def suggest(self, **kwargs) -> Dict:
-        pass
-
-    def stats(self, **kwargs) -> Dict:
-        pass
-
-    def get(self, **kwargs) -> "AquaDeployment":
+    def get(self, model_deployment_id, **kwargs) -> "AquaDeployment":
```
mrDzurb, line 230: NIT: typing is missed  for the ```model_deployment_id```

ads/aqua/deployment.py
```diff
@@ -116,23 +241,28 @@ def get(self, **kwargs) -> "AquaDeployment":
...
-            raise AquaClientError("Aqua model deployment ocid must be provided to fetch the deployment.")
-        
+        if not model_deployment_id:
+            raise AquaClientError(
+                "Aqua model deployment ocid must be provided to fetch the deployment."
+            )
+
         try:
             model_deployment = self.client.get_model_deployment(**kwargs).data
         except ServiceError as se:
```
mrDzurb, line 251: we add everywhere these exception statements, probably would be better to move them into the decorator?

ads/aqua/deployment.py
```diff
@@ -116,23 +241,28 @@ def get(self, **kwargs) -> "AquaDeployment":
...
+        aqua_service_model = (
+            model_deployment.freeform_tags.get(AQUA_SERVICE_MODEL, None)
+            if model_deployment.freeform_tags
+            else None
         )
 
         if not aqua_service_model:
-            raise AquaClientError(f"Target deployment {model_deployment.id} is not Aqua deployment.")
+            raise AquaClientError(
+                f"Target deployment {model_deployment.id} is not Aqua deployment."
```
mrDzurb, line 264: Probably we should say something like target deployment is not compatible with the AI Quick Actions, and point out users somewhere to check the docs

ads/aqua/deployment.py
```diff
@@ -116,23 +241,28 @@ def get(self, **kwargs) -> "AquaDeployment":
...
+            )
 
         return AquaDeployment(
```
mrDzurb, line 267: NIT: ```AquaDeployment.from_oci_model()```?

ads/aqua/extension/deployment_handler.py
```diff
@@ -0,0 +1,158 @@
...
+        try:
+            input_data = self.get_json_body()
+        except Exception:
+            raise HTTPError(400, Errors.INVALID_INPUT_DATA_FORMAT)
+
+        if not input_data:
+            raise HTTPError(400, Errors.NO_INPUT_DATA)
+
+        # required input parameters
+        compartment_id = input_data.get("compartment_id")
```
mrDzurb, line 57: For parameters it will be automatically handled, right?

ads/aqua/extension/deployment_handler.py
```diff
@@ -0,0 +1,158 @@
...
+        if not compartment_id:
```
mrDzurb, line 58: I would suggest to add some MDRequest dataclass and put all of the logic with constructing the input attributes and validation there.

ads/aqua/extension/deployment_handler.py
```diff
@@ -0,0 +1,158 @@
...
+            # todo: call create method to create a catalog entry
+            # aqua_model = AquaModelApp.create()
+            # todo: hardcoded image, get deployment_image and aqua_service_model tag from aqua model
+            #  Also, entrypoint would be added in the image
+            deployment_image = "iad.ocir.io/ociodscdev/aqua_deploy:1.0.0"
+            entrypoint = ["python", "/opt/api/api.py"]
+            aqua_service_model = f"xxx.xxx.xxx.xxx.xxx#aqua_service_model_value"
+
+            self.finish(
+                AquaDeploymentApp().create(
```
mrDzurb, line 112: if we have MDRequest dataclass, then the code could be 
```
AquaDeploymentApp().create(**md_request.to_dict())
```

ads/aqua/deployment.py
```diff
@@ -8,23 +8,28 @@
...
 AQUA_SERVICE_MODEL = "aqua_service_model"
 
-
 logger = logging.getLogger(__name__)
 
+
 @dataclass
 class AquaDeployment:
     """Represents an Aqua Model Deployment"""
+
```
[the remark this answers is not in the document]
VipulMascarenhas: during the design discussion we decided to have that in the aqua_service_model tag (ocid#model_name). We can have a separate field, but could get redundant due to the tag.

ads/aqua/deployment.py
VipulMascarenhas: yes, we'll revisit this and update the code once the error handler is finalized.

ads/aqua/deployment.py
VipulMascarenhas: agree. updating the message and will add a todo for linking to docs.

ads/aqua/deployment.py
VipulMascarenhas: yes, makes sense. will add it in a [separate PR](https://confluence.oci.oraclecorp.com/display/ODSC/Discussion+topics+regarding+Aqua+API+Implementation#DiscussiontopicsregardingAquaAPIImplementation-Deployment).

ads/aqua/extension/deployment_handler.py
VipulMascarenhas: agree. will add it in a [separate PR](https://confluence.oci.oraclecorp.com/display/ODSC/Discussion+topics+regarding+Aqua+API+Implementation#DiscussiontopicsregardingAquaAPIImplementation-Deployment).

PR #556, opened 2024-01-30
Operator in Jobs should auto-detect RP [WIP]

govarsha:
[ODSC-51737](https://jira.oci.oraclecorp.com/browse/ODSC-51737)
Adds resource_principal as auth in command args when operator is run as local backend inside job

### DIFF
--- a/ads/opctl/operator/common/backend_factory.py
+++ b/ads/opctl/operator/common/backend_factory.py
@@ -421,7 +421,7 @@ class BackendFactory:
         for resource_type in supported_backends:
             try:
                 for runtime_type_item in RUNTIME_TYPE_MAP.get(
-                        resource_type.lower(), []
+                    resource_type.lower(), []
                 ):
                     runtime_type, runtime_kwargs = next(iter(runtime_type_item.items()))
 
--- a/ads/opctl/operator/common/backend_factory.py
+++ b/ads/opctl/operator/common/backend_factory.py
@@ -14,6 +14,7 @@ from typing import Dict, List, Tuple, Union
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


ads/opctl/operator/common/backend_factory.py
```diff
@@ -199,6 +200,13 @@ def backend(
                     "An error occurred while attempting to load the "
                     f"configuration for the `{backend_kind}.{runtime_type}` backend."
                 )
+            
+        # make auth as resource principal if ads operator -f <yaml> command is run inside job
```
mrDzurb, line 204: Not sure why do we even need this? There is a proper way to run operator as a job, right? I think we better should inspire users to use this approach - 
https://accelerated-data-science.readthedocs.io/en/latest/user_guide/operators/common/run.html#run-with-conda-environment

However if we still want to achieve the desirable result when users run ```ads operator run -f forecast.yaml``` in Jobs and RP should be used as default auth type, then the actual change needs to be done here - 

https://github.com/oracle/accelerated-data-science/blob/main/ads/opctl/config/merger.py#L117

Similar to the ```is_in_notebook_session``` or maybe we can have some generic method for this, that can check if RP available in the given resource, then use RP as default.

govarsha: Actually epm seems to be running the ads operator inside the job and they faced issue because they did not pass -a resource_principal. So a ticket was cut to make resource principal as default for jobs. Yeah I can change in merger, I was not sure to make changes outside of operator folder, hence I tried to add it in cmd args in backend/-factory. I will change in merger.py

PR #558, opened 2024-01-31
Operator in Jobs should auto-detect RP

govarsha:
[ODSC-51737](https://jira.oci.oraclecorp.com/browse/ODSC-51737)
Added static function supported in ResourcePrincipal, to default to RP

### DIFF
--- a/ads/common/auth.py
+++ b/ads/common/auth.py
@@ -687,6 +687,10 @@ class ResourcePrincipal(AuthSignerGenerator):
         logger.info(f"Using 'resource_principal' authentication.")
         return signer_dict
 
+    @staticmethod
+    def supported():
+        return any(os.environ.get(var) for var in ['JOB_RUN_OCID', 'NB_SESSION_OCID', 'DATAFLOW_RUN_ID'])
+
 
 class InstancePrincipal(AuthSignerGenerator):
     """
--- a/ads/common/auth.py
+++ b/ads/common/auth.py
@@ -689,7 +689,10 @@ class ResourcePrincipal(AuthSignerGenerator):
 
     @staticmethod
     def supported():
-        return any(os.environ.get(var) for var in ['JOB_RUN_OCID', 'NB_SESSION_OCID', 'DATAFLOW_RUN_ID'])
+        return any(
+            os.environ.get(var)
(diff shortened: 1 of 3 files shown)


ads/opctl/config/merger.py
```diff
@@ -115,7 +115,7 @@ def _fill_config_with_defaults(self, ads_config_path: str) -> None:
         )
         # set default auth
         if not self.config["execution"].get("auth", None):
-            if is_in_notebook_session():
+            if is_in_notebook_session() or is_in_jobrun():
```
mrDzurb, line 118: LGTM. However as I mentioned in this review - https://github.com/oracle/accelerated-data-science/pull/556
We can easily check if the given resource supports RP, and use RP by default. In this case it will work everywhere. Let's say tomorrow we want to run operator in the same way in DataFlow service, we will have to add one more method. We can have something like this:

We have ```ResourcePrincipal``` class in the ```ads/common/auth.py```. We can add a static/class method or property there
```
class ResourcePrincipal:
     @staticmethod
            def supported():
                   # add validation here
```

Then whenever we to make decision which auth to use:

```
if ResourcePrincipal.supported:
    ...
else :
    ...
```

ads/common/auth.py
```diff
@@ -687,6 +687,10 @@ def create_signer(self) -> Dict:
         logger.info(f"Using 'resource_principal' authentication.")
         return signer_dict
 
+    @staticmethod
+    def supported():
+        return any(os.environ.get(var) for var in ['JOB_RUN_OCID', 'NB_SESSION_OCID', 'DATAFLOW_RUN_ID'])
```
mrDzurb, line 692: Could you also add - PIPELINE_RUN_OCID?

mrDzurb (approved):

ads/opctl/config/merger.py
govarsha: Yeah this looks better. I have made the changes. Please review it.

ads/common/auth.py
govarsha: Yeah, added it.

ahosler (approved):
Awesome Job! This will be critical for all of our operators!

PR #560, opened 2024-02-01
Deployment params for UI

VipulMascarenhas:
Adding additional fields for UI.


ads/aqua/deployment.py
```diff
@@ -160,7 +165,7 @@ def create(
             .with_env(env_var)
             .with_deployment_mode(ModelDeploymentMode.HTTPS)
             .with_model_uri(model_id)
-            .with_region(region)
+            .with_region(self.region)
```
lu-ohai, line 168: Where does this `self.region` come from?

PR #561, opened 2024-02-01
[ODSC_52446_52447] Get loggroups logids and compartment list

VipulMascarenhas:
### Description

This PR covers the following:

1. Use oc.OCIClientFactory to instantiate clients
2. Add LoggingManagementClient to oci_client.py
3. Adds 3 additional APIs to list_log_groups, list_logs and list_compartments
4. Update deployment handler to accommodate 3 additional APIs

###  Local Usage
These APIs can be tested from a local jupyter lab instance: 

Get Log Groups
```
GET http://localhost:8888/aqua/logging/?compartment_id=<ocid>
```

Get Log ids
```
GET http://localhost:8888/aqua/logging/<loggroup_ocid>
```

Get compartments
```
GET http://localhost:8888/aqua/compartments
```

### To Do

Unit Tests


ads/aqua/deployment.py
```diff
@@ -16,8 +16,12 @@
...
 from ads.aqua.exception import AquaClientError, AquaServiceError
-from ads.config import COMPARTMENT_OCID
+from ads.config import COMPARTMENT_OCID, ROOT_COMPARTMENT_OCID
+from oci.exceptions import ServiceError, ClientError
+from oci.logging.models import LogGroupSummary, LogSummary
+from oci.identity.models import Compartment
 
+# todo: move this to constants or have separate functions
 AQUA_SERVICE_MODEL = "aqua_service_model"
 CONSOLE_LINK_URL = (
```
mayoor, line 26: push this to a util method - get_console_link or something like that.

ads/aqua/deployment.py
```diff
@@ -277,18 +287,17 @@ def from_oci_model_deployment(cls, oci_model_deployment, region) -> "AquaDeploym
...
+            oci_model_deployment.model_deployment_configuration_details.model_configuration_details.instance_configuration
         )
         instance_shape_config_details = (
             instance_configuration.model_deployment_instance_shape_config_details
         )
         return AquaDeployment(
             id=oci_model_deployment.id,
             display_name=oci_model_deployment.display_name,
-            aqua_service_model=oci_model_deployment.freeform_tags.get(AQUA_SERVICE_MODEL),
+            aqua_service_model=oci_model_deployment.freeform_tags.get(
```
mayoor, line 298: This should be boolean? `aqua_service_model=oci_model_deployment.freeform_tags.get(AQUA_SERVICE_MODEL) is not None`

ads/aqua/deployment.py
```diff
@@ -297,11 +306,93 @@ def from_oci_model_deployment(cls, oci_model_deployment, region) -> "AquaDeploym
...
-                if instance_shape_config_details else None
+                if instance_shape_config_details
+                else None
             ),
-            console_link=CONSOLE_LINK_URL.format(oci_model_deployment.id, region)
-        ) 
+            console_link=CONSOLE_LINK_URL.format(oci_model_deployment.id, region),
+        )
+
+    def list_log_groups(self, **kwargs) -> List["LogGroupSummary"]:
```
mayoor, line 320: These pass through APIs that are not part of deployment service, can be kept in a separate file under `extension/aqua_ui_handler.py` or something like that. We dont need to show any of these in cli

ads/config.py
```diff
@@ -47,6 +47,9 @@
 NO_CONTAINER = os.environ.get("NO_CONTAINER")
 TMPDIR = os.environ.get("TMPDIR")
 
+# Aqua config
+ROOT_COMPARTMENT_OCID = os.environ.get("ROOT_COMPARTMENT_OCID")
```
mayoor, line 51: call it ODSC_MODEL_COMPARTMENT_OCID

ads/aqua/deployment.py
VipulMascarenhas: we're currently treating aqua_service_model as a string field (something like key=aqua_service_model val=llama2), which is also passed along when deployment is created.  This is due to an absence of a tag field in AquaModelSummary. Would it be better to treat this as boolean and have something like tags: Dict[str, str] in the dataclass?

ads/aqua/deployment.py
VipulMascarenhas: added in ads.common.utils

ads/aqua/deployment.py
VipulMascarenhas: moved the list log group, log ids and compartments to ui_handler.py.

ads/aqua/deployment.py
```diff
@@ -14,21 +14,28 @@
...
 from ads.config import COMPARTMENT_OCID
 
+
+# todo: move this to constants or have separate functions
 AQUA_SERVICE_MODEL = "aqua_service_model"
-CONSOLE_LINK_URL = (
-    "https://cloud.oracle.com/data-science/model-deployments/{}?region={}"
-)
 
 logger = logging.getLogger(__name__)
```
mrDzurb, line 25: Let's use one logger for AQUA, there is one specified in the __init__.py.

ads/aqua/deployment.py
```diff
@@ -37,10 +44,8 @@ class AquaDeployment(DataClassSerializable):
     created_on: str
     created_by: str
     endpoint: str
-    instance_shape: str
-    ocpus: float
-    memory_in_gbs: float
     console_link: str
+    shape_info: ShapeInfo
```
mrDzurb, line 48: NIT: ```shape_info: ShapeInfo = field(default_factory=ShapeInfo)```

ads/aqua/deployment.py
```diff
@@ -37,10 +44,8 @@ class AquaDeployment(DataClassSerializable):
     created_on: str
```
mrDzurb, line 44: It would be better to specify default values.

ads/aqua/deployment.py
```diff
@@ -241,15 +248,16 @@ def get(self, model_deployment_id: str, **kwargs) -> "AquaDeployment":
         # add error handler
```
mrDzurb, line 248: Needs to be cleaned up?

ads/aqua/deployment.py
```diff
@@ -259,9 +267,11 @@ def get(self, model_deployment_id: str, **kwargs) -> "AquaDeployment":
         return AquaDeploymentApp.from_oci_model_deployment(
             model_deployment, self.region
         )
-    
+
     @classmethod
-    def from_oci_model_deployment(cls, oci_model_deployment, region) -> "AquaDeployment":
+    def from_oci_model_deployment(
```
mrDzurb, line 272: Oh I was thinking that this method will be added to the ```AquaDeployment```?

ads/aqua/model.py
```diff
@@ -90,11 +89,10 @@ def get(self, model_id) -> "AquaModel":
...
         """
         # add error handler
-        oci_model = self.client.get_model(model_id).data
-
+        oci_model = self.ds_client.get_model(model_id).data
         # add error handler
-        # if not self._if_show(oci_model):
-        #     raise AquaClientError(f"Target model {oci_model.id} is not Aqua model.")
+        if not self._if_show(oci_model):
+            raise AquaClientError(f"Target model {oci_model.id} is not Aqua model.")
```
mrDzurb, line 95: Probably will not be able to use the AQUA name. I would suggest to have a custom error, something like ```NotAquaCompatibleError```.

ads/common/utils.py
```diff
@@ -1713,3 +1713,30 @@ def progress_callback(progress):
...
+        OCID of the resource
+    region: str
+        The Region Identifier that the client should connect to.
+
+    Returns
+    -------
+    console_link_url: str
+        a valid link to the console for the given resource
+    """
+    console_link_url = "https://cloud.oracle.com/data-science/{}/{}?region={}".format(
```
mrDzurb, line 1739: NIT: ```f""https://cloud.oracle.com/data-science/{resource}/{ocid}?region={region}""```

ads/aqua/deployment.py
VipulMascarenhas: updated

ads/aqua/deployment.py
VipulMascarenhas: updated

ads/aqua/deployment.py
VipulMascarenhas: updated

ads/aqua/deployment.py
VipulMascarenhas: yes, that makes sense. Moved to the AquaDeployment.

ads/aqua/model.py
VipulMascarenhas: yes, we'll update this once exception handler PR is finalized.

ads/common/utils.py
VipulMascarenhas: changed

ads/aqua/deployment.py
VipulMascarenhas: keeping for now, we have an action item to clean all this up once exception handler PR is finalized.

PR #586, opened 2024-02-09
[ODSC-50335] Model by Reference Implementation

VipulMascarenhas:
### Description

The Model Store team has enabled users to pass [Model Artifact by reference](https://confluence.oci.oraclecorp.com/display/ODSC/Model+by+Reference%3A+Design+Summary) when saving their models. This eliminates the need to copy model artifacts from a user's Object Storage bucket to the System Model Store bucket.
Relevant jira ticket: [ODSC-50335](https://jira.oci.oraclecorp.com/browse/ODSC-50335)

The implementation design for this PR is documented [here](https://confluence.oci.oraclecorp.com/display/ODSC/%5BDesign%5D+ADS+Model+By+Reference+Implementation).

### Testing

```
from ads.model.datascience_model import DataScienceModel
artifact_path = "oci://bucket_name@namespace/prefix"
display_name = "Test"

model = (DataScienceModel()
  .with_compartment_id("ocid1.compartment.oc1..xxx")
  .with_project_id("ocid1.datascienceproject.oc1.iad.xxx")
  .with_display_name(display_name)
  .with_artifact(artifact_path)
)
model.create(model_by_reference=True,)

```
This uploads a json file in the catalog entry with the file paths of the files in the artifact_path.

To download:
```
target_dir = 'local_dir'
 
# if model was already created using DataScienceModel().create(model_by_reference=True,)
model.download_artifact(target_dir)
 
# if loading using from_id for model_ocid which was created by reference
model =  DataScienceModel.from_id(model_ocid)
model.download_artifact(target_dir)

```

Unit Tests - WIP


ads/common/utils.py
```diff
@@ -1740,3 +1740,73 @@ def get_console_link(
...
+    """
+    # Add default signer if the uri is an object storage uri, and
+    # the user does not specify config or signer.
+    if uri.startswith("oci://") and "config" not in kwargs and "signer" not in kwargs:
+        kwargs.update(authutil.default_signer())
+    with fsspec.open(uri, "r", **kwargs) as f:
+        return f.read()
+
+
+def download_object_versions(
```
mrDzurb, line 1764: I think the name is a bit confusing, maybe - ```download_from_object_storage```? Can we instead of dictionary use the List[DataClass]? The dataclass can be placed in the object_storage_details module for example. 

Something like:
```
@dataclass
ObjectStoragePath:
     path: str
     version: str
```

then
```
def download_object_versions(paths: List[ObjectStoragePath], ...)
```

ads/common/utils.py
```diff
@@ -1740,3 +1740,73 @@ def get_console_link(
...
+            namespace_name=path_details.namespace,
+            bucket_name=path_details.bucket,
+            object_name=path_details.filepath,
+            version_id=version,
+        )
+
+        local_filepath = os.path.join(target_dir, path_details.filepath)
+        os.makedirs(os.path.dirname(local_filepath), exist_ok=True)
+
+        with open(local_filepath, "wb") as _file:
```
mrDzurb, line 1810: Probably this will be nery slow downloader. It would be nice to test the performance of this method. For example for LLama2 artifacts? I would suggest to use either AsyncIO or ThreadPoolExecutor to do parallel loading.

ads/model/artifact_downloader.py
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env python
 # -*- coding: utf-8; -*-
+import logging
```
mrDzurb, line 3: We have too many loggers :) Lets use common logger form ADS? We have a ticket to improve this in future. 

```
from ads.common import logger
```

ads/model/artifact_downloader.py
```diff
@@ -85,17 +87,31 @@ class SmallArtifactDownloader(ArtifactDownloader):
...
-        with open(zip_file_path, "wb") as zip_file:
-            zip_file.write(zip_content)
-        self.progress.update("Extracting model artifacts")
-        with ZipFile(zip_file_path) as zip_file:
-            zip_file.extractall(self.target_dir)
-        
-        utils.remove_file(zip_file_path)
+
+        file_name = (
+            "model_description" if file_extension == ".json" else str(uuid.uuid4())
```
mrDzurb, line 101: Can we make this validation as case insensitive?

ads/model/artifact_downloader.py
```diff
@@ -110,6 +126,7 @@ def __init__(
         bucket_uri: Optional[str] = None,
         overwrite_existing_artifact: Optional[bool] = True,
         remove_existing_artifact: Optional[bool] = True,
+        model_file_description: Optional[dict] = None,
```
mrDzurb, line 129: Could you add this attribute into the Params section?

ads/model/model_metadata.py
```diff
@@ -983,6 +983,29 @@ def from_dict(cls, data: Dict) -> "ModelMetadata":
         """
         pass
 
+    def contains(self, key: str) -> bool:
```
mrDzurb, line 986: I think ```get``` can be used instead?

ads/model/service/oci_datascience_model.py
```diff
@@ -304,18 +304,22 @@ def get_model_artifact_content(self) -> BytesIO:
...
         bytes_content: BytesIO
             Model artifacts to upload.
+        extension: str
+            File extension, defaults to zip
         """
         self.client.create_model_artifact(
             self.id,
             bytes_content,
-            content_disposition=f'attachment; filename="{self.id}.zip"',
+            content_disposition=f'attachment; filename="{self.id}{extension}"',
```
mrDzurb, line 322: This will fail if the extension will be passed as None?

ads/common/object_storage_details.py
```diff
@@ -138,3 +140,77 @@ def is_oci_path(uri: str = None) -> bool:
...
         return uri.lower().startswith("oci://")
+
+    def is_bucket_versioned(self) -> bool:
+        """Check if the given bucket is versioned.
+        Returns
+        -------
+        bool: return True if the bucket is versioned.
+
+        """
+        os_client = oci_client.OCIClientFactory(**self.auth).object_storage
```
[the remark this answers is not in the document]
VipulMascarenhas: added lazy creation of os_client in this class.

ads/common/utils.py
VipulMascarenhas: yea it was slow initially without the parallel loading. I was experimenting with ThreadPoolExecutor locally and it took approx. ~49 mins for Llama2-13b-hf for almost 104 Gb data, and ~13 mins for Phi-2 for almost 11 Gb of data downloaded from the OSS buckets. But I also have a gigabit connection, on VPN almost got around 500 Mbps download speed. Got to run more tests benchmarks on dev environment.

ads/model/artifact_downloader.py
VipulMascarenhas: ah yes, it was an unused import - removed it.

ads/model/artifact_downloader.py
VipulMascarenhas: updated.

ads/model/artifact_downloader.py
VipulMascarenhas: added

ads/model/model_metadata.py
VipulMascarenhas: yep - removed contains, get should work.

qiuosier:
Do we have a way to list all the referenced artifacts?

qiuosier:
I am seeing the progress is always 40%. Not sure if I am missing something?
image

VipulMascarenhas:
> I am seeing the progress is always 40%, then almost at the end it is bumping up to 80%. Not sure if I am missing something? image
> 
> image

I need to add a better progress bar based on total size of objects, the one you see is a current progress bar that keeps track of the artifact downloader steps.

VipulMascarenhas:
> Do we have a way to list all the referenced artifacts?

yes, there's a property called model_file_description in DataScienceModel, that should give you the list of all referenced artifacts.

ads/model/service/oci_datascience_model.py
VipulMascarenhas: changed this to always have zip as default extension.

PR #588, opened 2024-02-09
AQUA inference handler

kumar-shivam-ranjan:
This PR is intended to create ADS handler for hitting MD predict URL and returning back the response.


ads/aqua/deployment.py
```diff
@@ -410,3 +411,37 @@ def get_deployment_config(self, model_id: str) -> Dict:
             raise AquaServiceError(opc_request_id=None, status_code=500)
 
         return shape_config
+
+    @staticmethod
+    def get_model_deployment_inference(endpoint, prompt, model_params: Dict):
```
mayoor, line 416: Let us define a dataclass to encapsulate the  the model params and the prompt.

VipulMascarenhas (commented):
changes look good, let's add dataclass like Mayoor said. I tested the changes locally with a sample request (changed ocid), response looks good.

```
{
    "endpoint": "https://modeldeployment.us-ashburn-1.oci.customer-oci.com/ocid1.datasciencemodeldeployment.oc1.iad.xyz/predict",
    "prompt": "What is machine learning? Explain like I am five.",
    "model_params": {
                "max_tokens": 50,
                "temperature": 0.5,
                "top_k": 10,
                "top_p": 0.5
    }
}
```

```
{
    "text": [
        "What is machine learning? Explain like I am five.\n\nMachine learning is like teaching a computer to do something new by itself. Just like how you learn new things, a computer can also learn new things. Instead of telling the computer exactly what to do, we give it some examples and let it"
    ]
}
```

ads/aqua/deployment.py
kumar-shivam-ranjan: Done

VipulMascarenhas: something like this would be better:
```
@dataclass
class ModelParams:
    max_tokens: int = None
    temperature: float = None
    top_k: float = None
    top_p: float = None
    model: str = None

@dataclass
class MDInferenceResponse:
	model_params: field(default_factory=ModelParams) = None
	prompt: str = None
```
you can look at deployment.py for reference.

ads/aqua/deployment.py
```diff
@@ -410,3 +411,56 @@ def get_deployment_config(self, model_id: str) -> Dict:
             raise AquaServiceError(opc_request_id=None, status_code=500)
 
         return shape_config
+
+class AquaDeploymentInferenceRequest(AquaApp):
```
mrDzurb, line 415: Do we really need to put this class here? Do we want to allow user to send request via AQUA CLI?

mrDzurb (approved):

VipulMascarenhas (approved):
lgtm 👍

PR #592, opened 2024-02-14
[ODSC-52993] Set up caching for listing models and compartments

VipulMascarenhas:
To cache model list API results, we can use TTL cache from cachetools. [docs](https://cachetools.readthedocs.io/en/stable/#cachetools.TTLCache)

We have set up TTL cache for model listing (5 hrs) and compartment listing (2 hrs) APIs. Since we need an option to clear/invalidate the cache, we need to set `lock` as well. 

Note that caching is only enabled for models in the service compartment and not in the custom compartment.

To invalidate service model cache, user can call: 
```
DELETE http://localhost:8888/aqua/model/cache
```

To invalidate compartment listing cache, call:
```
DELETE http://localhost:8888/aqua/compartments/cache
```

Time to list models locally:
- Before caching ~700 ms
- After caching ~75ms

Time to list compartments locally:
- Before caching ~2.7 secs
- After caching ~60ms

To do:
- update THIRD_PARTY_LICENSES.txt if we proceed with this.


pyproject.toml
```diff
@@ -163,7 +163,7 @@ pii = [
   "spacy==3.6.1",
 ]
 llm = ["langchain>=0.0.295", "evaluate>=0.4.0"]
-aqua = ["fire"]
+aqua = ["fire", "cachetools"]
```
mrDzurb, line 166: We need to check license for this.

ads/aqua/extension/model_handler.py
```diff
@@ -24,6 +24,11 @@ def read(self, model_id):
         """Read the information of an Aqua model."""
         return self.finish(AquaModelApp().get(model_id))
 
+    @handle_exceptions
```
mrDzurb, line 27: Not sure if we need this :) User will not be able to reach out this end point. If only we add supporting refresh button in UI which might be a good idea.

ads/aqua/model.py
```diff
@@ -172,8 +176,16 @@ def get(self, model_id) -> "AquaModel":
             ),
         )
 
+    @cached(
```
mrDzurb, line 179: I think the problem with this approach is that we want different caching for the Service models and Custom models. Maybe we can start only with the service models?

VipulMascarenhas: In that case, the parameters values for the cache key can be set as None, that way results are cached only when the function is called without specific compartment id and project id. Would using a cached decorator still be a good idea or is it better to explicitly maintain a cache in the class?

mrDzurb: There are pros and cons. Does it support saving cache into the file? ALso does it support updating cache explicitly? For example if we return user the cashed data but in the parallel thread update the case with the fresh data?

pyproject.toml
VipulMascarenhas: did a search on slack and seems a lot of projects are using it, still need to confirm.

ads/aqua/model.py
VipulMascarenhas: having different cache for service vs. custom models isn't supported via the decorator directly. It's doable with custom wrappers but makes it more complex. Moved the cache outside and now we're caching service models and compartments only.
Need more testing on our end but UI team can use it in the meanwhile to that they don't have to wait longer times for testing.

mayoor: Let us start with the service models for now. The custom models anyway goes through RQS, which is makes it slightly faster.
Let us also add it to the compartment listing as well. I think these two will improve the overall experience.

ads/aqua/extension/model_handler.py
```diff
@@ -24,6 +25,16 @@ def read(self, model_id):
...
+    def delete(self, id=""):
+        """Handles DELETE request for clearing cache"""
+        url_parse = urlparse(self.request.path)
+        paths = url_parse.path.strip("/")
+        if paths.startswith("aqua/model/cache"):
```
mrDzurb, line 33: Wouldn't it be better to move the paths into the const variable? At least all of them will be in one place.

ads/aqua/ui.py
```diff
@@ -71,16 +79,10 @@ def list_logs(self, **kwargs) -> str:
             log_group_id=log_group_id, **kwargs
         ).data.__repr__()
 
-    def list_compartments(self, **kwargs) -> str:
+    def list_compartments(self) -> str:
```
mrDzurb, line 82: NIT: Typo?
```
Lists the compartments in a compartment
```

ads/aqua/ui.py
```diff
@@ -131,7 +140,10 @@ def list_compartments(self, **kwargs) -> str:
                     0,
                     Compartment(id=TENANCY_OCID, name=" ** Root - Name N/A **"),
```
mrDzurb, line 141: That what we want to show to user if they don't have permissions?

ads/aqua/ui.py
```diff
@@ -131,7 +140,10 @@ def list_compartments(self, **kwargs) -> str:
...
                 )
-            return compartments.__repr__()
+            res = compartments.__repr__()
```
mrDzurb, line 143: Can you add a small comment here, explaining why do we need this logic?

ads/aqua/ui.py
VipulMascarenhas: yea, should have been `Lists the compartments in a tenancy..`

ads/aqua/ui.py
VipulMascarenhas: added

PR #595, opened 2024-02-15
Aqua/list model v4

mingkang111:
# Description
This PR related to the following jira:
https://jira.oci.oraclecorp.com/browse/ODSC-53009
https://jira.oci.oraclecorp.com/browse/ODSC-52395

# Changed
* added `search_text` field in response.
* Added unit test for `list` api
* Fixed the CLI docstring for AquaModel and AquaModel.list

# Test
Screenshot 2024-02-16 at 5 41 32 PM

### DIFF
--- a/ads/aqua/model.py
+++ b/ads/aqua/model.py
@@ -247,7 +247,9 @@ class AquaModelApp(AquaApp):
             ),
         )
 
-        search_text = str(tags) if tags else UNKNOWN
+        search_text = (
+            self._build_search_text(tags, description="This is a") if tags else UNKNOWN
+        )
 
         return dict(
             compartment_id=model.compartment_id,
@@ -297,3 +299,10 @@ class AquaModelApp(AquaApp):
         return OCIResource.search(
             query, type=SEARCH_TYPE.STRUCTURED, tenant_id=TENANCY_OCID, **kwargs
         )
+
+    def _build_search_text(self, tags: dict, description: str = None) -> str:
+        """Constructs search_text field in response."""
+        description = description or ""
+        tags_text = ",".join(str(v) for v in tags.values())
+
+        return description + tags_text
--- a/ads/aqua/model.py
+++ b/ads/aqua/model.py
@@ -54,16 +54,6 @@ class AquaModelSummary(DataClassSerializable):
     console_link: str
     search_text: str
 
-    @staticmethod
-    def load_icon(model_name) -> str:
-        """Loads icon."""
-        # TODO: switch to the official logo
-        try:
-            return create_word_icon(model_name, return_as_datauri=True)
-        except Exception as e:
-            logger.debug(f"Failed to load icon for the model={model_name}.")
-            return None
-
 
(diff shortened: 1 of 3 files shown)


lu-ohai (approved):

VipulMascarenhas (approved):

ads/aqua/model.py
```diff
@@ -268,15 +276,31 @@ def _if_show(self, model: "AquaModel") -> bool:
...
-            tenant_id=TENANCY_OCID,
+            query, type=SEARCH_TYPE.STRUCTURED, tenant_id=TENANCY_OCID, **kwargs
         )
+
+    def _build_search_text(self, tags: dict, description: str = None) -> str:
+        """Constructs search_text field in response."""
+        description = description or ""
+        tags_text = ",".join(str(v) for v in tags.values())
+
+        return description + tags_text
```
mayoor, line 306: There should be space between description and tags_text

mayoor (approved):
I have left a minor comment to be taken care of

ads/aqua/model.py
mingkang111: Hi @mayoor , thanks for the review! The description and tags_text will be separated by "," (see test section above) may I know is that enough?

mrDzurb (approved):

ads/aqua/model.py
mayoor: The `search_text` in the screenshot looks like begins with `,`. I dont see description there.

ads/aqua/model.py
```diff
@@ -242,6 +247,8 @@ def process_model(cls, model, region) -> dict:
             ),
         )
 
+        search_text = self._build_search_text(tags) if tags else UNKNOWN
```
mayoor, line 250: How is the description passed in for constructing the search text?

mingkang111: This method reserves the space for description. Before rqs fixed the schema, we cannot directly get description from of rqs response. We will need extra `get_model` call to get it. I remembered that we decided not to go with extra call for now.

ads/aqua/model.py
mingkang111: Yes. We don't pass description for now. Once rqs fix the schema, I will update here.

ads/aqua/model.py
mayoor: Thanks @mingkang111. There should be space between the first comma and the tag, which I dont see in the test case output. Also, can you please add a TODO above this method call so that we dont miss it later. Feel free to merge it with the space fix.

mingkang111: Fixed and added todo. I also added one test for this method. It can help us to view the expected output easier.

test-requirements.txt
```diff
@@ -3,6 +3,7 @@ click
 coverage
 faker
 mock
+parameterized
```
mingkang111, line 6: BSD license. We only invoke its api and we won't distribute it.

PR #609, opened 2024-02-27
Update create evaluation api

lu-ohai:
### Update create evaluation api

- Updated `evaluation_description` to optional
- Add a checker to validate the model parameters
- Update the Model Hyperparameters with `shape`
- Added parameters in create response
- Refactored duplicated imports

### DIFF
--- a/ads/aqua/utils.py
+++ b/ads/aqua/utils.py
@@ -26,6 +26,12 @@ README = "README.md"
 UNKNOWN_JSON_STR = "{}"
 SUPPORTED_FILE_FORMATS = ["jsonl"]
 MODEL_BY_REFERENCE_OSS_PATH_KEY = "Object Storage Path"
+MODEL_PARAMETERS = [
+    "max_tokens",
+    "temperature",
+    "top_p",
+    "top_k"
+]
 
 
 def get_logger():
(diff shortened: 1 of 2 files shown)


ads/aqua/utils.py
```diff
@@ -95,6 +95,12 @@ def get_status(evaluation_status: str, job_run_status: str = None):
 }
 SUPPORTED_FILE_FORMATS = ["jsonl"]
 MODEL_BY_REFERENCE_OSS_PATH_KEY = "Object Storage Path"
+MODEL_PARAMETERS = [
```
mingkang111, line 98: Can we use `AquaEvalParams.__annotations__.keys()` instead to keep consistency with the field in return response? 

It will return
```
dict_keys(['shape', 'max_tokens', 'top_p', 'top_k', 'temperature'])
```

lu-ohai: Because the `AquaEvalParams` consists two parts, one is the `shape`, which is required and has been pre-checked in the handler, the other part is the model parameters. The model parameters is the thing that we want to check. If adding the `shape` in it, would it be very confusing for user?

mingkang111: Yes, you're right. Adding `shape` will be confusing. Lets hard code for now. I think these value probably won't change. 

Or you can try: 
```
valid_model_params = list(AquaEvalParams.__annotations__.keys())
valid_model_params.remove("shape")
```

PR #612, opened 2024-02-28
Cancel & Delete Evaluation

VipulMascarenhas:
### Description

This PR takes care of deleting and canceling evaluations. For cancel, only the evaluation job is canceled. For delete, first we delete the job and then we delete the associated model for evaluation.

Tested deleting a job can with active Job run, that would not work with this setup. If we need to delete the job, then we should cancel the jobs run first, and then delete the job. 

- Cancel API
```
ads aqua evaluation cancel --eval_id "ocid1.datasciencemodel.oc1.iad.<OCID>"
```
Response:
```
id:              ocid1.datasciencemodel.oc1.iad.<OCID>
lifecycle_state: CANCELING
time_accepted:   2024-03-04 16:58:54.229819
INFO:ads.aqua:Canceling Job Run: ocid1.datasciencejobrun.oc1.iad.<OCID> for evaluation ocid1.datasciencemodel.oc1.iad.<OCID>
```

- Delete API
```
ads aqua evaluation delete --eval_id "ocid1.datasciencemodel.oc1.iad.<OCID>"
```

Response:
```
id:              ocid1.datasciencemodel.oc1.iad.<OCID>
lifecycle_state: DELETING
time_accepted:   2024-03-04 17:01:07.564414
INFO:ads.aqua:Deleting Job: ocid1.datasciencejob.oc1.iad.<OCID> for evaluation ocid1.datasciencemodel.oc1.iad.<OCID>
INFO:ads.aqua:Deleting evaluation: ocid1.datasciencemodel.oc1.iad.<OCID>
```

The API requests would look like:
- Cancel : `PUT http://localhost:8888/aqua/evaluation/<OCID>/cancel`
- Delete: `DELETE http://localhost:8888/aqua/evaluation/<OCID>`

### DIFF
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -933,7 +933,7 @@ class AquaEvaluationApp(AquaApp):
 
         return report
 
-    def cancel_evaluation(self, eval_id) -> dict:
+    def cancel(self, eval_id) -> dict:
         """Cancels the job run for the given evaluation id.
         Parameters
         ----------
@@ -984,7 +984,7 @@ class AquaEvaluationApp(AquaApp):
             )
         return status
 
-    def delete_evaluation(self, eval_id):
+    def delete(self, eval_id):
         """Deletes the job and the associated model for the given evaluation id.
         Parameters
         ----------
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -14,6 +14,7 @@ from pathlib import Path
 from threading import Lock
 from typing import Any, Dict, List, Optional, Union
 
+
 import oci
 from cachetools import TTLCache
 from oci.data_science.models import (
@@ -41,6 +42,7 @@ from ads.aqua.utils import (
     UNKNOWN,
     is_valid_ocid,
     upload_file_to_os,
+    fire_and_forget,
 )
 from ads.common.auth import AuthType
 from ads.common.object_storage_details import ObjectStorageDetails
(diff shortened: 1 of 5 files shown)


ads/aqua/evaluation.py
```diff
@@ -933,6 +933,118 @@ def download_report(self, eval_id) -> AquaEvalReport:
...
+        except ValueError:
+            raise AquaMissingKeyError(
+                f"Custom metadata is missing {EvaluationCustomMetadata.EVALUATION_JOB_ID.value} key"
+            )
+
+        job = DataScienceJob.from_id(job_id)
+
+        # submit a coroutine for execution of job and model deletion task which runs in the background,
+        # no need to wait for result.
+        asyncio.create_task(self._delete_job_and_model(job, model))
```
VipulMascarenhas, line 1025: @mrDzurb needed some inputs on this. Currently, delete is a time consuming task if any of the job runs are in progress, hence we run this in the background for api call.
However, I realized that for the CLI case, we'll get a running event loop error as this call would be expected to be awaited. any thoughts on how we can handle this in a better way for both cases?

VipulMascarenhas: Update this logic -  previous code was:
```
asyncio.create_task(self._delete_job_and_model(job, model))
```

The option that seemed to work for both API and CLI was to define a wrapper for the _delete_job_and_model() function which runs this in the background. 

```
def fire_and_forget(func):
    @wraps(func)
    def wrapped(*args, **kwargs):
        return asyncio.get_event_loop().run_in_executor(None, func, *args, *kwargs)

    return wrapped
```

mrDzurb (approved):

PR #618, opened 2024-02-29
Automlx Version Upgrade

prasankh:
Automlx version upgrade from `23.2.3` to `23.4.1`

### DIFF
--- a/ads/opctl/operator/lowcode/forecast/model/automlx.py
+++ b/ads/opctl/operator/lowcode/forecast/model/automlx.py
@@ -57,7 +57,7 @@ class AutoMLXOperatorModel(ForecastOperatorBaseModel):
     @runtime_dependency(
         module="automl",
         err_msg=(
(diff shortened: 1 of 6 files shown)


ads/opctl/operator/lowcode/forecast/model/automlx.py
```diff
@@ -57,7 +57,7 @@ def preprocess(self, data, series_id=None):
     @runtime_dependency(
         module="automl",
         err_msg=(
-            "Please run `pip3 install oracle-automlx==23.2.3` to install the required dependencies for automlx."
```
ahosler, line 60: Don't we need to change this to automlx now?

https://docs.oracle.com/en-us/iaas/tools/automlx/latest/html/multiversion/latest/guides/quickstart.html

Concerning that the unit tests haven't flagged this

ahosler (changes requested):

PR #620, opened 2024-02-29
Update create evaluation with job runtime

lu-ohai:


lu-ohai:
Close as it's already merged.

PR #625, opened 2024-03-01
Mock APIs for Fine Tuning 

VipulMascarenhas:
This PR adds 3 APIs with mock results:
- List VCN with `GET http://localhost:8888/aqua/vcn`
- List Subnet with `GET http://localhost:8888/aqua/subnets?vcn_id=ocid1.vcn.oc1.iad.<OCID>`
- Create Fine Tuned Model with `POST http://localhost:8888/aqua/finetuning -d '{"id" : "value"}'`

This PR also modifies `job.py` to house base dataclasses that will be used both for evaluation and fine tuning. This PR only covers the changes for fine tuning, we can inherit the base classes in evaluation modules later on.


ads/aqua/job.py
```diff
@@ -6,68 +6,55 @@
...
+class DataScienceResource(Enum):
+    MODEL_DEPLOYMENT = "datasciencemodeldeployment"
+    MODEL = "datasciencemodel"
 
 
-@dataclass
-class AquaFineTuningJob(AquaJob):
-    """Represents an Aqua fine-tuning job."""
+@dataclass(repr=False)
+class AquaResourceIdentifier(DataClassSerializable):
```
mingkang111, line 29: Hi @VipulMascarenhas , this dataclass also used by evaluation, could you please put it into some common place and then other module can use from there as well?

ads/aqua/job.py
```diff
@@ -6,68 +6,55 @@
...
-    """Represents an Aqua job summary."""
-
-    id: str
-    compartment_id: str
-    project_id: str
-
-    model_id: str
-    task: str
-
+class Resource(Enum):
```
mingkang111, line 16: class `Resource` and  `DataScienceResource` are not unique in job. It would be better to put them into a common place shared by other module.

ads/aqua/job.py
```diff
@@ -6,68 +6,55 @@
...
+@dataclass(repr=False)
+class AquaResourceIdentifier(DataClassSerializable):
+    id: str = ""
+    name: str = ""
+    url: str = ""
 
-class AquaJobApp(AquaApp):
-    """Contains APIs for Aqua jobs."""
 
+class JobTypeTags(Enum):
```
mingkang111, line 42: Currently, our tags are dispersed across various locations, making it challenging to monitor their values. I would like to suggest consolidating all tags into a single location to ensure consistency in their values.

ads/aqua/finetune.py
```diff
@@ -0,0 +1,79 @@
...
+
+@dataclass(repr=False)
+class AquaFineTuningSummary(AquaJobSummary, DataClassSerializable):
+    parameters: AquaFineTuningParams = field(default_factory=AquaFineTuningParams)
+
+
+class AquaFineTuningApp(AquaApp):
+    """Contains APIs for Aqua fine-tuning jobs."""
+
+    def create(self, **kwargs) -> "AquaFineTuningSummary":
```
lu-ohai, line 28: If there is no explicit parameters for `create`, validating the input data would be very hard and can't be generalized

ads/aqua/ui.py
```diff
@@ -247,3 +247,92 @@ def list_job_shapes(self, **kwargs) -> list:
...
+            for `list_vcns <https://docs.oracle.com/iaas/api/#/en/iaas/20160918/Vcn/ListVcns>`_
+
+        Returns
+        -------
+            json representation of `oci.core.models.Vcn`."""
+
+        compartment_id = kwargs.pop("compartment_id", COMPARTMENT_OCID)
+        logger.info(f"Loading job shape summary from compartment: {compartment_id}")
+
+        # todo: add VirtualNetworkClient in oci_client.py
```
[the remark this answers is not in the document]
VipulMascarenhas: sure, I had left this for implementation by whoever works on the API, but I'll add it in this PR.

ads/aqua/job.py
VipulMascarenhas: I was thinking that job.py remains the base location for all common finetuning and evaluation classes, but moving to utils also makes sense.

ads/aqua/finetune.py
VipulMascarenhas: I had kept validation within the finetuning handler, we can move it herem right? That way validation works for both CLI as well as the inputs from the API call.

ads/aqua/ui.py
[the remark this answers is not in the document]
VipulMascarenhas: added

ads/aqua/finetune.py
lu-ohai: It would be inconsistent if move here as for create deployment and evaluation, handler checks if the required parameters are passed and reports corresponding http 400 error if missing.

VipulMascarenhas: agree, but this check will also have to be handled for CLI inputs. 
we have a few options here:
- use dict as input
- keep params as is and not have an object as input
- move all params within the function for all flows including deployment & evaluation and do validation here.
any preferences on what we can go with?

@mrDzurb, we discussed this last week - let us know if you prefer any particular method to go with here.

mingkang111: We can consider to use `pydantic` to help with validating input, if validation fails `pydantic` will raise an error with a breakdown of what was wrong, for example: 

```
from pydantic import BaseModel

class Info(BaseModel):
    conf_one: float
    conf_two: str
    conf_three: bool

    class Config:
        extra = 'forbid'


class ConfStructure(BaseModel):
    version: int
    info: Info


my_conf_wrong = {
    'version': 1,

    'info': {
        'conf_one': 2.5,
        'conf_two': 'foo',
        'conf_three': False,
        'optional_conf': 'bar'
    }
}
res = ConfStructure(**my_conf_wrong)
# pydantic.error_wrappers.ValidationError: 1 validation error for ConfStructure
#     info -> optional_conf
# extra fields not permitted (type=value_error.extra)
```

cc: @mrDzurb @mayoor @qiuosier

ads/aqua/model.py
```diff
@@ -268,7 +259,9 @@ def _process_model(
     ) -> dict:
         """Constructs required fields for AquaModelSummary."""
 
-        icon = self._load_icon(model.display_name)
+        # todo: revisit icon generation code
+        # icon = self._load_icon(model.display_name)
```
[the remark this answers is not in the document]
mingkang111: Thanks!

ads/aqua/ui.py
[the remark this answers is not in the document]
mingkang111: Thank you!

ads/aqua/job.py
mingkang111: Thank you!

ads/aqua/finetune.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2024 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
+from dataclasses import dataclass, field
+from ads.common.serializer import DataClassSerializable
+from ads.aqua.base import AquaApp
+from ads.aqua.job import AquaJobSummary
+from ads.aqua.data import Resource, AquaResourceIdentifier, Tags
```
mingkang111, line 10: Looks like there is no `data.py` under ads.aqua module.

VipulMascarenhas: my bad - missed pushing this file while switching local branches. added now.

ads/aqua/finetune.py
VipulMascarenhas: thanks Ming! yea, we can use pydantic or the custom validation Lu implemented - but I think Lu's question is more around where to do this, either in handler or within the code that handler calls. 

@lu we can go with Ming's idea here and use that to validate for CLI, but passing direct object to create() doesn't work (at least from what I read in the fire code/docs) - we'll have to allow user to pass in inputs with  dict or individual params. feel free to modify this code and we can review when we test the implementation of create API.

ads/aqua/finetune.py
mingkang111: No worries! Thank you!

ads/aqua/finetune.py
mingkang111: Can this work?
```
class CreateInputParam(BaseModel):
    expect_input_param1: list
    expect_input_param2: str
    kwargs: dict
    ...

class AquaXXXApp:
    def create(self, create_aqua_xxx_details = None, **kwargs) -> "AquaXXSummary":
        input_param = create_aqua_xxx_details if create_aqua_xxx_details and isinstance(create_aqua_xxx_details, CreateInputParam) else CreateInputParam(**kwargs)
```

In this case, API can still validate create_aqua_xxx_details from handler, and CLI can also use **kwargs. In this case, CLI can even work with dict as input.
for example:
```
input_param = CreateInputParam(**create_aqua_xxx_details) if isinstance(create_aqua_xxx_details, dict)
```

VipulMascarenhas: yes, this look good. 👍

mingkang111 (approved):

ads/aqua/finetune.py
qiuosier: I think we can put the validation here and let the UI/API handler to return 400 error if there is a validation error. Is there any `kwargs` that are not supposed to be part of the `create_aqua_xxx_details`? If not, then the entire `kwargs` can be treated as a `dict` and there is no need to have a separated object?
If we want to have CLI specific arguments in `kwargs`, we can have something like
```
# Extract the arguments for CLI
# Pop the CLI args from kwargs
cli_args = parse_cli_kwargs(kwargs)
# Now kwargs has been modified and is supposed to contain kwargs for creating the object only
# AquaDataModel can now do validation (with Pydantic or whatever).
AquaDataModel(**kwargs)
```

qiuosier (approved):

ads/aqua/extension/finetune_handler.py
```diff
@@ -0,0 +1,43 @@
...
+
+        if not input_data:
+            raise HTTPError(400, Errors.NO_INPUT_DATA)
+
+        # todo: create CreateAquaFineTuningDetails class and validate inputs
+        # validate_function_parameters(
+        #     data_class=CreateAquaFineTuningDetails, input_data=input_data
+        # )
+        try:
+            self.finish(AquaFineTuningApp().create(**input_data))
```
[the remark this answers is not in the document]
VipulMascarenhas: sounds good!

PR #627, opened 2024-03-01
Added uploading status.

lu-ohai:
### Update uploading dataset

- Moved uploading dataset to the front
- Added aqua evaluation tag after the job run is created
- Added size checker for local dataset to not exceed 50 MB
- Fixed a path bug in oci bucket


ads/aqua/evaluation.py
```diff
@@ -91,6 +92,11 @@ class EvaluationJobTags(Enum):
     EVALUATION_MODEL_ID = "evaluation_model_id"
 
 
+class EvaluationUploadStatus(Enum):
+    IN_PROGRESS = "IN_PROGRESS"
+    COMPLETED = "COMPLETED"
```
mingkang111, line 97: Hi @lu-ohai , may I know is there any new discussion regarding to this status? 
Based on current implementation, when jobrun=succeeded && model==active, our return status is `succeeded`. There is no more `COMPLETED` being used. You can check `LifecycleStatus.get_status` in ads/aqua/utils.py for more details.
In terms of `IN_PROGRESS`, we might not need to create this from ads. We can use oci's service status: `oci.data_science.models.JobRun.LIFECYCLE_STATE_IN_PROGRESS`.

mingkang111 (approved):
Testing successfully.

PR #629, opened 2024-03-04
Exclude the model version set OCID from the model details when updating the model.

mrDzurb:
## Description

I've observed that attempting to update a model already linked to a ```ModelVersionSet```, particularly when passing the same ```model_version_set_id``` as before, results in an error indicating the model is already associated with the Model Version Set.


mingkang111:
LGTM!

PR #630, opened 2024-03-04
Add Telemetry module to ADS

liudmylaru:
### Description

First step to include telemetry to aqua module is to move telemetry module into ADS.
https://jira.oci.oraclecorp.com/browse/ODSC-53773

### What was done

- move telemetry module from odsc_cli to ADS codebase

### Validation

- builded whl
- installed it in nb-session
- run code to record event in the conda bucket (I am not sure if aqua logs created in dev)
```
from ads import set_auth
from ads.telemetry.client import TelemetryClient

set_auth("resource_principal")
AQUA_BUCKET = "service-managed-models"
AQUA_BUCKET_NS = "ociodscdev"
telemetry = TelemetryClient(bucket=AQUA_BUCKET, namespace=AQUA_BUCKET_NS)
category = "ads/telemetry/module"
action = "test"
telemetry.record_event_async(category=f"{category}", action=action)
```
- checked logs records. Saw an expected entry with. check the ticket (https://jira.oci.oraclecorp.com/browse/ODSC-53773) comment section fro a screenshot.

### DIFF
--- a/ads/telemetry/base.py
+++ b/ads/telemetry/base.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2024 Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+
+import logging
+
+from ads import set_auth
+from ads.common import oci_client as oc
+from ads.common.auth import default_signer
+from ads.config import OCI_RESOURCE_PRINCIPAL_VERSION
+
+
+logger = logging.getLogger(__name__)
+class TelemetryBase:
+    """Base class for Telemetry Client."""
+
+    def __init__(self, bucket: str, namespace: str = None) -> None:
+        """Initializes the telemetry client.
+
+        Parameters
+        ----------
+        bucket : str
+            OCI object storage bucket name storing the telemetry objects.
+        namespace : str, optional
+            Namespace of the OCI object storage bucket, by default None.
+        """
+        if OCI_RESOURCE_PRINCIPAL_VERSION:
+            set_auth("resource_principal")
+        self._auth = default_signer()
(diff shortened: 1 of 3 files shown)


ads/telemetry/base.py
```diff
@@ -0,0 +1,62 @@
...
+        Parameters
+        ----------
+        bucket : str
+            OCI object storage bucket name storing the telemetry objects.
+        namespace : str, optional
+            Namespace of the OCI object storage bucket, by default None.
+        """
+        if OCI_RESOURCE_PRINCIPAL_VERSION:
+            set_auth("resource_principal")
+        self._auth = default_signer()
```
VipulMascarenhas, line 30: I think we need to pass service endpoint here. Before, when trying to hit int endpoints, auth was point to prod when this wasn't set.

`self._auth = default_signer({"service_endpoint": OCI_ODSC_SERVICE_ENDPOINT})`

VipulMascarenhas (approved):
changes look good to me 👍

ads/telemetry/base.py
```diff
@@ -0,0 +1,62 @@
...
+        Parameters
+        ----------
+        bucket : str
+            OCI object storage bucket name storing the telemetry objects.
+        namespace : str, optional
+            Namespace of the OCI object storage bucket, by default None.
+        """
+        if OCI_RESOURCE_PRINCIPAL_VERSION:
+            set_auth("resource_principal")
+        self._auth = default_signer({"service_endpoint": OCI_ODSC_SERVICE_ENDPOINT})
```
qiuosier, line 30: We should not use ODSC endpoint here as this is for object storage.

ads/telemetry/base.py
VipulMascarenhas: just saw QQ's comment, I didn't know this only works for DS client and not other clients. :)

VipulMascarenhas (approved):

PR #633, opened 2024-03-05
AQUA/FT/Update the response based on the new requirements

mingkang111:
# Description
Based on the discussion, we added the following fields into response:
* jobrun's info (ocid,name,console uri)
* both val_loss and train_loss will be returned in metrics
* final_train_loss added into returning metrics

### DIFF
--- a/ads/aqua/model.py
+++ b/ads/aqua/model.py
@@ -37,6 +37,7 @@ from ads.model.datascience_model import DataScienceModel
 @dataclass(repr=False)
 class AquaFineTuningMetric(DataClassSerializable):
     name: str
+    category: str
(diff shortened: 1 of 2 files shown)


ads/aqua/model.py
```diff
@@ -37,6 +37,7 @@
 @dataclass(repr=False)
 class AquaFineTuningMetric(DataClassSerializable):
     name: str
+    category: str
```
VipulMascarenhas, line 40: nit: type or dataset_type seems clearer to understand  for user than category. Fine with either.

VipulMascarenhas (approved):

PR #638, opened 2024-03-06
Fix error type when file already exists

lu-ohai:
### Fix error type when file already exists

- Added new class `AquaFileExistsError`

### DIFF
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -353,7 +353,7 @@ class AquaEvaluationApp(AquaApp):
             except FileExistsError:
                 raise AquaFileExistsError(
                     f"Dataset {dataset_file} already exists in {create_aqua_evaluation_details.report_path}. "
-                    "Please use a new dataset file name."
+                    "Please use a new dataset file name or report path."
                 )
             logger.debug(
                 f"Uploaded local file {evaluation_dataset_path} to object storage {dst_uri}."
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -26,6 +26,7 @@ from oci.data_science.models import (
 from ads.aqua import logger, utils
 from ads.aqua.base import AquaApp
 from ads.aqua.exception import (
+    AquaFileExistsError,
     AquaFileNotFoundError,
     AquaMissingKeyError,
     AquaRuntimeError,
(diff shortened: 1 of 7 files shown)


ads/aqua/evaluation.py
```diff
@@ -340,13 +341,20 @@ def create(
...
+            try:
+                upload_local_to_os(
+                    src_uri=evaluation_dataset_path,
+                    dst_uri=dst_uri,
+                    auth=self._auth,
+                    force_overwrite=False,
+                )
+            except FileExistsError:
+                raise AquaFileExistsError(
+                    f"Dataset {dataset_file} already exists in {create_aqua_evaluation_details.report_path}. "
```
mingkang111, line 355: Provides anther report path or a new dataset file name will be both work?

ads/aqua/extension/evaluation_handler.py
```diff
@@ -59,7 +60,10 @@ def post(self, *args, **kwargs):
                 )
             )
         except Exception as ex:
-            raise HTTPError(500, str(ex))
+            if isinstance(ex, AquaError):
```
mingkang111, line 63: We don't need this. Our @handle_exception handler will handle the error for us. 
I would suggest to remove the try catch around `self.finish`:
```

    @handle_exceptions
    def post(self, *args, **kwargs):
        """Handles post request for the evaluation APIs

        Raises
        ------
        HTTPError
            Raises HTTPError if inputs are missing or are invalid.
        """
        try:
            input_data = self.get_json_body()
        except Exception:
            raise HTTPError(400, Errors.INVALID_INPUT_DATA_FORMAT)

        if not input_data:
            raise HTTPError(400, Errors.NO_INPUT_DATA)

        validate_function_parameters(
            data_class=CreateAquaEvaluationDetails, input_data=input_data
        )

        self.finish(
                # TODO: decide what other kwargs will be needed for create aqua evaluation.
                AquaEvaluationApp().create(
                    create_aqua_evaluation_details=(
                        CreateAquaEvaluationDetails(**input_data)
                    )
                )
          )
```

lu-ohai: Thanks, I didn't know that 😄

mingkang111:
Verify the results successfully.
Screenshot 2024-03-06 at 1 13 09 PM

mingkang111 (approved):

PR #642, opened 2024-03-07
Fixed subnet id while creating evaluation/ft job

lu-ohai:
### Fixed subnet id while creating evaluation/ft job

- Added new env variable `AQUA_JOB_SUBNET_ID` to set the custom egress for evaluation/fine tuning jobs.
- Fixed subnet issue when running job on notebook session.


ads/aqua/evaluation.py
```diff
@@ -494,6 +494,18 @@ def create(
                     memory_in_gbs=create_aqua_evaluation_details.memory_in_gbs,
                     ocpus=create_aqua_evaluation_details.ocpus,
                 )
+            if AQUA_JOB_SUBNET_ID:
+                evaluation_job.infrastructure.with_subnet_id(
+                    AQUA_JOB_SUBNET_ID
+                )
+            else:
+                if NB_SESSION_IDENTIFIER in os.environ:
```
mayoor, line 502: Do we need this if statement?

mayoor: I think we can always set to default networking

lu-ohai: As per document: https://accelerated-data-science.readthedocs.io/en/latest/user_guide/jobs/infra_and_runtime.html#networking. Job using other resource's default networking only applies to NB, so I have a checker here specially for NB. @qiuosier

qiuosier: Both of you are right :) This if does not really matter.

mayoor (approved):

PR #645, opened 2024-03-07
improve apis by requirements

mingkang111:
# TODO
improve unit test later

### DIFF
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -177,6 +177,15 @@ class AquaEvaluationSummary(DataClassSerializable):
     parameters: AquaEvalParams = field(default_factory=AquaEvalParams)
 
 
+@dataclass(repr=False)
+class AquaEvaluationDetail(DataClassSerializable):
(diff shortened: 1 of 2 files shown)


ads/aqua/evaluation.py
```diff
@@ -700,12 +709,65 @@ def get(self, eval_id) -> AquaEvaluationSummary:
...
+            loggroup = utils.query_resource(log_id, return_all=False)
+            loggroup_name = loggroup.display_name if loggroup else ""
+
+        try:
+            introspection = json.loads(
+                self._get_attribute_from_model_metadata(resource, "ArtifactTestResults")
+            )
+        except:
+            # TODO: remove this hardcode value later
+            introspection = {
```
mrDzurb, line 751: we can return just empty dictionary in this case

mrDzurb (approved):

ads/aqua/evaluation.py
mingkang111: Sure. This is for demo purpose in case any bug happen. I will remove it today.

PR #646, opened 2024-03-08
Deployment and Finetuning config

VipulMascarenhas:
This PR covers the following:

1. MODEL_BY_REFERENCE_OSS_PATH_KEY has been changed from `Object Storage Path` to `artifact_location`.
2. load_config moved to utils so that it can be reused for other config files as well.
3. Finetuning config has been added.  To call, use `GET http://localhost:8888/aqua/finetuning/config/<OCID>`

`finetuning_config.json`
```
{
"shape": {
      "VM.GPU.A10.1": {
        "replica": 1,
        "batch_size": 2
    }
}
```

5. Deployment config will now have nested parameters required for deployment, instead of common one at the root level. The config will look like this:

`deployment_config.json`
```
{
    "shape": [
        "VM.GPU.A10.1",
        "VM.GPU.A10.2"
    ]
}
```

### DIFF
--- a/ads/aqua/deployment.py
+++ b/ads/aqua/deployment.py
@@ -276,14 +276,19 @@ class AquaDeploymentApp(AquaApp):
         model_path_prefix = os_path.filepath.rstrip("/")
 
         # todo: revisit this after new image is built
-        env_var.update({"MODEL": f"{AQUA_MODEL_DEPLOYMENT_FOLDER}{model_path_prefix}"})
+        env_var.update({"MODEL": f"{model_path_prefix}"})
         env_var.update(
             {"PARAMS": f"--served-model-name {AQUA_MODEL_DEPLOYMENT_FOLDER}"}
         )
         env_var.update({"MODEL_DEPLOY_PREDICT_ENDPOINT": "/v1/completions"})
-        cards = int(instance_shape.split(".")[-1])
-        if cards > 1:
-            env_var.update({"TENSOR_PARALLELISM": str(cards)})
+
+        # todo: added for testing, remove when TENSOR_PARALLELISM will be set inside the container
+        try:
+            cards = int(instance_shape.split(".")[-1])
+            if cards > 1:
+                env_var.update({"TENSOR_PARALLELISM": str(cards)})
+        except:
+            pass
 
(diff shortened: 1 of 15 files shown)


ads/aqua/extension/finetune_handler.py
```diff
@@ -39,5 +55,13 @@ def post(self, *args, **kwargs):
         except Exception as ex:
             raise HTTPError(500, str(ex))
```
lu-ohai, line 56: This is a bug here. `create` inside will raise 400 error, but adding this `try catch` will make all client errors from `create` as 500. This needs to be removed as the `handle_exceptions` will take care of this. Same for the deployment handler to create model deployment.

ads/aqua/deployment.py
```diff
@@ -252,10 +248,27 @@ def create(
...
+        env_var.update({"MODEL": f"{AQUA_MODEL_DEPLOYMENT_FOLDER}{model_path_prefix}"})
+
+        deployment_config = load_config(
+            artifact_path,
+            config_file_name=AQUA_MODEL_DEPLOYMENT_CONFIG,
+            auth=self._auth,
+        )
+        shape_dict = deployment_config["shape"]
+        if instance_shape not in shape_dict:
+            raise AquaValueError(
```
lu-ohai, line 266: Same as I mentioned below, if 400 client error is raised in create model deployment, the `try catch` must be removed.

ads/aqua/extension/finetune_handler.py
VipulMascarenhas: yes, you're right. post method hasn't been implemented yet, we can address this change when in the create FT PR.

lu-ohai: Same for the `deployment_handler.py` as there will be 400 error raised in `create`: https://github.com/oracle/accelerated-data-science/blob/shape_config_api/ads/aqua/extension/deployment_handler.py#L111

VipulMascarenhas: updated this PR for all apis within deployment_handler, can you check if anything is missing?

ads/aqua/deployment.py
VipulMascarenhas: updated

PR #662, opened 2024-03-12
Telemetry for Aqua

VipulMascarenhas:
### Description
This PR covers the following:
1. Adds telemetry to aqua APIs.
2. Reuse auth and clients, moved default_signer() to base to avoid repetitive code.
3. Telemetry for errors - Since error handling is done at the handler level, we'll only be able to track errors for the Aqua UI and not CLI. We should move the error handling process one level down in the future.
 
### Testing
In Progress.


ads/telemetry/client.py
```diff
@@ -66,14 +70,18 @@ def record_event(self, category: str = None, action: str = None, **kwargs) -> No
         if not category or not action:
             raise ValueError("Please specify the category and the action.")
         endpoint = f"{self.service_endpoint}/n/{self.namespace}/b/{self.bucket}/o/telemetry/{category}/{action}"
+        if detail:
+            endpoint = f"{endpoint}/{detail}"
```
liudmylaru, line 74: Telemetry data parser will cut of last word and put into "action" column in database, then it will get middle section between "telemetry" and last word - and save it into "category" column.
This if will make "detail" got into "action" column when data will be harvested. We can't have smart parser, as we do not know which telemetry will/will not have detail after "action".

"detail" will bring confusion.
any "detail" we would like to have, this has to be saved into value dict.

ads/aqua/finetune.py
```diff
@@ -445,6 +448,19 @@ def create(
             ),
         )
 
+        # tracks the shape used for fine-tuning the service models
+        self.telemetry.record_event_async(
+            category="aqua/finetune",
+            action="create/shape",
```
liudmylaru, line 454: when records will be parsed, only 1 word will become an "action". To not bring confusion, I would suggest to use same approach when recording event. Category can include several words, but action is one last word always.

ads/aqua/model.py
```diff
@@ -600,11 +612,22 @@ def list(
         List[AquaModelSummary]:
             The list of the `ads.aqua.model.AquaModelSummary`.
         """
+
         models = []
         if compartment_id:
+            # tracks number of times custom model listing was called
+            self.telemetry.record_event_async(
```
liudmylaru, line 619: record events here invoked without try catch blocks. To be safe that our code for telemetry will never break, I would suggest to add try-except, like in Example in the description class TelemetryClient(TelemetryBase).

OR, we can add this try-catch inside TelemetryClient().record_event() method. (like this option better)

ads/telemetry/client.py
liudmylaru: I think I forgot to add value also into headers. Here is initial document with value - https://confluence.oci.oraclecorp.com/display/ODSC/%5BTelemetry%5D+Python+Client. For some reason it was not implemented in odsc_cli.
@qiuosier should we add "value" parameter to record_event() method? Or this will be handled with kwargs. And We should just add one example into description for record_envent() on how to record value to Logs.

ads/aqua/model.py
VipulMascarenhas: second option is good or we can use a try-except wrapper like this:
```
def try_except_wrapper(func):
    def wrapper(*args, **kwargs):
        try:
            return func(*args, **kwargs)
       except Exception as e:
           logger.error("There is an error recording telemetry : {e}")
    return wrapper
```
Then use it as:
```
@try_except_wrapper
def record_event_async()
    ...
```

ads/aqua/finetune.py
VipulMascarenhas: updated.

ads/aqua/deployment.py
```diff
@@ -383,6 +382,23 @@ def create(
...
+            # tracks unique deployments that were created in the user compartment
+            self.telemetry.record_event_async(
+                category="aqua/custom/deployment", action="create", detail=model_name
+            )
+        else:
+            # tracks unique deployments that were created in the user compartment
+            self.telemetry.record_event_async(
+                category="aqua/service/deployment", action="create", detail=model_name
+            )
+        # tracks the shape used for fine-tuning the service models
```
liudmylaru, line 395: split and put into is-else above, so that:
(add proper comment)

aqua/service/deployment/create/shape/{shape_1}
aqua/service/deployment/create/shape/{shape_2}
aqua/service/deployment/create/shape/{shape_3}

aqua/custom/deployment/create/shape/{shape_1}
aqua/custom/deployment/create/shape/{shape_2}
aqua/custom/deployment/create/shape/{shape_3}

liudmylaru (approved):

PR #664, opened 2024-03-12
Update shapes for Fine Tuning

VipulMascarenhas:
This PR adds additional logic to look up configs for fine tuning creation and deployment. If config for fine tuned models need to be looked up, then we'll use the base name instead of the model name.

### DIFF
--- a/ads/aqua/base.py
+++ b/ads/aqua/base.py
@@ -11,7 +11,13 @@ from oci.data_science.models import UpdateModelDetails, UpdateModelProvenanceDet
 from ads import set_auth
 from ads.aqua import logger
 from ads.aqua.data import Tags
-from ads.aqua.utils import UNKNOWN, is_valid_ocid, load_config, logger
+from ads.aqua.utils import (
+    UNKNOWN,
+    is_valid_ocid,
+    load_config,
+    logger,
+    get_base_model_from_tags,
+)
 from ads.common import oci_client as oc
 from ads.common.auth import default_signer
 from ads.common.utils import extract_region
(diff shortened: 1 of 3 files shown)


ads/aqua/base.py
```diff
@@ -224,6 +230,14 @@ def get_config(self, model_id: str, config_file_name: str) -> Dict:
         oci_model = self.ds_client.get_model(model_id).data
         model_name = oci_model.display_name
 
+        is_fine_tuned_model = (
```
lu-ohai, line 233: This can be simplified as:
```
is_fine_tuned_model = Tags.AQUA_FINE_TUNED_MODEL_TAG.value in oci_model.freeform_tags
```

VipulMascarenhas: updated.

ads/aqua/utils.py
```diff
@@ -572,3 +573,17 @@ def wrapped(*args, **kwargs):
...
     return wrapped
+
+
+def get_base_model_from_tags(tags):
+    base_model_ocid = None
+    base_model_name = None
+    if Tags.AQUA_FINE_TUNED_MODEL_TAG.value in tags:
+        tag = tags[Tags.AQUA_FINE_TUNED_MODEL_TAG.value]
+        base_model_ocid, base_model_name = tag.split("#")
+        if not (is_valid_ocid(base_model_ocid) and base_model_name):
```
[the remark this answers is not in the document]
VipulMascarenhas: I thought first that the caller should handle it, but raising error here makes more sense. Updated this.

mingkang111 (approved):

PR #672, opened 2024-03-13
Cleanup evaluation API constants.

mrDzurb:
## Description
Cleanup evaluation API constants.


lu-ohai (approved):

mingkang111 (approved):
May I know have we tested the creation of eval after these change?

PR #675, opened 2024-03-14
Add API to get model license 

mingkang111:
# Usage
```
GET  http://localhost:8888/aqua/model/<ocid>/license
```

# Response
```
{
"id": <ocid>,
"license": "xxxxxxx"
}
```


ads/aqua/model.py
```diff
@@ -667,55 +682,6 @@ def clear_model_list_cache(
                 }
         return res
 
-    def ModelCustomMetadata_process_model(
```
mingkang111, line 670: Remove the unintended commit before.

ads/aqua/extension/model_handler.py
```diff
@@ -43,7 +43,17 @@ def list(self):
         project_id = self.get_argument("project_id", default=None)
         return self.finish(AquaModelApp().list(compartment_id, project_id))
 
+class AquaModelLicenseHandler(AquaAPIhandler):
```
VipulMascarenhas, line 46: do we need an extra handler here? Can we move this within the current handler and manage the url within the current get()?

lu-ohai (approved):

ads/aqua/extension/model_handler.py
mingkang111: I am adding extra handler because I would like to avoid adding too much if else in one get method.

PR #683, opened 2024-03-15
[Do not merge] Updated fetching config path in model artifact

lu-ohai:
### Updated fetching config path in model artifact

- Merge this pr only after service models have all updated the config path


ads/aqua/base.py
```diff
@@ -213,29 +213,29 @@ def if_artifact_exist(self, model_id: str, **kwargs) -> bool:
...
+    ) -> Dict:
         """Gets the config for the given Aqua model.
+
         Parameters
         ----------
         model_id: str
             The OCID of the Aqua model.
         config_file_name: str
             name of the config file
+        default_shapes: Union[Dict, List]
```
VipulMascarenhas, line 230: get_config can be generic and don't need to return just shapes list. It would be better to let get_config not take in default_shapes, but return empty from get_config if we don't find it in the json. Then get_deployment_config can check if get_config returned empty, and return "shape": default_shapes list. 

`Using default {default_shapes} instead.` message can be logged inside get_deployment_config.

ads/aqua/model.py
```diff
@@ -736,12 +735,20 @@ def load_license(self, model_id: str) -> AquaModelLicense:
...
+        if not (artifact_path and commit):
+            raise AquaMissingKeyError(
+                "Failed to get artifact path or commit from custom metadata.",
+                500
+            )
 
         content = str(
-            read_file(file_path=f"{artifact_path}/{LICENSE_TXT}", auth=default_signer())
+            read_file(
+                file_path=f"{artifact_path.rstrip('/')}/{commit}/{LICENSE_TXT}",
```
VipulMascarenhas, line 749: nit: we could do rstrip('/') within get_artifact_path_and_commit so that you don't have to do it outside.

ads/aqua/utils.py
```diff
@@ -64,6 +64,13 @@
 JOB_INFRASTRUCTURE_TYPE_DEFAULT_NETWORKING = "ME_STANDALONE"
 NB_SESSION_IDENTIFIER = "NB_SESSION_OCID"
 LIFECYCLE_DETAILS_MISSING_JOBRUN = "The asscociated JobRun resource has been deleted."
+DEFAULT_DEPLOYMENT_SHAPE_LIST = [
```
VipulMascarenhas, line 67: fine to hardcode here now, we should think of moving it outside the code later on.

VipulMascarenhas (approved):
overall looks good 👍  added a few comments, but nothing is a blocker.

ads/aqua/base.py
lu-ohai: Yes, but we need to log why we're using the default for users in `get_deployment_config` api, this could either due to, config missing, invalid config, or missing artifact path, which could cause duplicate logging as some of them are already be logged in details while in `get_config` api. We can keep this in discuss in future if there's any more type of return value added.

ads/aqua/utils.py
lu-ohai: Do you have any suggestion where to put this?

VipulMascarenhas: either in conda bucket or service models bucket. Fine to keep it here for the first iteration.

VipulMascarenhas:
we can deprecate this PR, we took the changes in a different branch and merged it. thank you!

lu-ohai:
Closing as it's merged at https://github.com/oracle/accelerated-data-science/pull/692

PR #688, opened 2024-03-15
FT/EVAL improvement regarding MVS

mingkang111:
# Description
The model version that we use in FT should not show the version set from evaluation. We need to add tags to the version set that is create for evaluation and FT to distinguish them and show them accordingly.

# APIs
* List experiments for evaluation:
```
GET  http://localhost:8888/aqua/experiment/
```

* List model version sets for FT
```
GET  http://localhost:8888/aqua/finetuning/modelversionsets
```

### DIFF
--- a/ads/aqua/base.py
+++ b/ads/aqua/base.py
@@ -11,12 +11,14 @@ from oci.data_science.models import UpdateModelDetails, UpdateModelProvenanceDet
 from ads import set_auth
 from ads.aqua import logger
 from ads.aqua.data import Tags
+from ads.aqua.exception import AquaRuntimeError, AquaValueError
 from ads.aqua.utils import (
     UNKNOWN,
+    _is_valid_mvs,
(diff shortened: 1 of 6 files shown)


ads/aqua/base.py
```diff
@@ -132,24 +132,54 @@ def create_model_version_set(
...
+        -------
+        tuple: (model_version_set_id, model_version_set_name)
+        """
         if not model_version_set_id:
             try:
                 model_version_set = ModelVersionSet.from_name(
                     name=model_version_set_name,
                     compartment_id=compartment_id,
                 )
+                # TODO: tag should be selected based on which operation (eval/FT) invoke this method
```
mingkang111, line 160: Currently, this method only used in the creation of FT. So hardcode the tag for now.

lu-ohai (approved):

VipulMascarenhas (approved):
changes look good!

PR #696, opened 2024-03-18
adding series_id in inliers, outliers

ahosler:
Outliers
image
image

Inliers
image
image

### DIFF
--- a/ads/opctl/operator/lowcode/common/data.py
+++ b/ads/opctl/operator/lowcode/common/data.py
@@ -24,7 +24,6 @@ class AbstractData(ABC):
         self.data = None
         self._data_dict = dict()
         self.name = name
-        self.data_with_all_cols = None
         self.load_transform_ingest_data(spec)
 
     def get_raw_data_by_cat(self, category):
--- a/ads/opctl/operator/lowcode/common/data.py
+++ b/ads/opctl/operator/lowcode/common/data.py
@@ -27,9 +27,18 @@ class AbstractData(ABC):
         self.data_with_all_cols = None
         self.load_transform_ingest_data(spec)
 
-
-    def get_data_with_all_cols(self):
-        return self.data_with_all_cols.reset_index(drop=False)
+    def get_raw_data_by_cat(self, category):
+        import pandas as pd
+        mapping = self._data_transformer.get_target_category_columns_map()
+        # For given category, mapping gives the target_category_columns and it's values.
+        # condition filters raw_data based on the values of target_category_columns for the given category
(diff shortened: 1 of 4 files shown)


ads/opctl/operator/lowcode/common/data.py
```diff
@@ -24,8 +24,14 @@ def __init__(self, spec: dict, name="input_data"):
         self.data = None
         self._data_dict = dict()
         self.name = name
+        self.data_with_all_cols = None
```
ahosler, line 27: We should either have the data object be mutable or immutable. Since the rest of the code is mutable, let's go with that (unless I'm missing something, which is likely).
Let's remove this attirbute and just call the method directly when we need it

ads/opctl/operator/lowcode/common/data.py
```diff
@@ -65,7 +71,10 @@ def _load_data(self, data_spec, **kwargs):
     def _transform_data(self, spec, raw_data, **kwargs):
         transformation_start_time = time.time()
         self._data_transformer = self.Transformations(spec, name=self.name)
-        data = self._data_transformer.run(raw_data)
+        self.data_with_all_cols = self._data_transformer.run(raw_data)
+        data = self.data_with_all_cols
+        if spec.target_category_columns:
+            data = data.drop(spec.target_category_columns, axis=1)
```
ahosler, line 77: Can we use more specific variable names? it's hard to follow what's happening here. (I know that you mostly inherited this problem)

ads/opctl/operator/lowcode/common/transformations.py
```diff
@@ -85,7 +85,7 @@ def _set_series_id_column(self, df):
             df[DataColumns.Series] = merge_category_columns(
                 df, self.target_category_columns
             )
-            df = df.drop(self.target_category_columns, axis=1)
+            #df = df.drop(self.target_category_columns, axis=1)
```
ahosler, line 88: Can we safely remove this? Why have we left it commented?

ads/opctl/operator/lowcode/common/data.py
```diff
@@ -24,8 +24,14 @@ def __init__(self, spec: dict, name="input_data"):
         self.data = None
         self._data_dict = dict()
         self.name = name
+        self.data_with_all_cols = None
         self.load_transform_ingest_data(spec)
 
+
+    def get_data_with_all_cols(self):
```
ahosler, line 31: Maybe we could name this: "get_original_data"? or "get_raw_data"? 
Is this just "raw data" or do we need this data to be label encoded? I think we can dig a bit deeper here.

PR #697, opened 2024-03-18
Telemetry: Track shape name, replica and ocid along with the model.

qiuosier:
This would help understanding how the different shape and replica combinations are used for specific models.
Also the last a few characters of the OCID is saved so that we might use it to match with other telemetry records (in the future).


VipulMascarenhas (approved):
lgtm 👍

liudmylaru (approved):

PR #704, opened 2024-03-18
Evaluation APIs Improvements

mingkang111:
# Description
This PR contains improvements on the following APIs:
* List eval
* Get status

# Major change
* Fix the logic for determine termination state
* Add cache for evaluation in terminal state
* Still use rqs to get each jobrun but use threadpool with 10 threads and send all possible requests in parallel

# Test 
```
GET http://localhost:8888/aqua/evaluation
```

### DIFF
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -356,7 +356,7 @@ class AquaEvaluationApp(AquaApp):
 
     _report_cache = TTLCache(maxsize=10, ttl=timedelta(hours=5), timer=datetime.now)
     _metrics_cache = TTLCache(maxsize=10, ttl=timedelta(hours=5), timer=datetime.now)
-    _eval_cache = TTLCache(maxsize=100, ttl=timedelta(hours=5), timer=datetime.now)
+    _eval_cache = TTLCache(maxsize=200, ttl=timedelta(hours=10), timer=datetime.now)
     _cache_lock = Lock()
 
     @telemetry(entry_point="plugin=evaluation&action=create", name="aqua")
@@ -1642,8 +1642,13 @@ class AquaEvaluationApp(AquaApp):
             oci.resource_search.models.ResourceSummary, oci.data_science.models.JobRun
         ] = None,
     ) -> dict:
-        """Builds evaluation status based on the model status and job run status."""
+        """Builds evaluation status based on the model status and job run status.
+        When detect `aqua_evaluation_error` in custom metadata, the jobrun is failed.
+        However, if jobrun failed before saving this meta, we need to check the existance
+        of the evaluation artifact.
 
+        """
+        # TODO: revisit for CANCELED evaluation
         job_run_status = (
             JobRun.LIFECYCLE_STATE_FAILED
             if self._get_attribute_from_model_metadata(
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -7,6 +7,7 @@ import json
 import os
 import re
 import tempfile
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import asdict, dataclass, field
 from datetime import datetime, timedelta
 from enum import Enum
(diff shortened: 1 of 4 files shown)


ads/aqua/evaluation.py
```diff
@@ -1569,8 +1619,17 @@ def _get_status(
         ] = None,
     ) -> dict:
         """Builds evaluation status based on the model status and job run status."""
-        model_status = model.lifecycle_state
+
```
mrDzurb, line 1622: TO not forget what we are doing here, could you add more description?

ads/aqua/evaluation.py
```diff
@@ -1380,6 +1429,7 @@ def _get_source(
             if not source_name:
                 resource_type = utils.get_resource_type(source_id)
 
+                # TODO: adjust resource principal mapping
```
mrDzurb, line 1432: What is this comment about?

ads/aqua/evaluation.py
```diff
@@ -1569,8 +1619,17 @@ def _get_status(
...
         job_run_status = (
+            JobRun.LIFECYCLE_STATE_FAILED
+            if self._get_attribute_from_model_metadata(
+                model, EvaluationCustomMetadata.EVALUATION_ERROR.value
+            )
+            else None
+        )
+
+        model_status = model.lifecycle_state
+        job_run_status = job_run_status or (
```
mrDzurb, line 1632: I would prefer to avoid using such constructions :) it is really hard to understand what's going on here. 
Wouldn't it be better to have something like this?

```
if self._get_attribute_from_model_metadata(
            model, EvaluationCustomMetadata.EVALUATION_ERROR.value
        ):
            job_run_status = JobRun.LIFECYCLE_STATE_FAILED
        elif jobrun and not jobrun.lifecycle_state == JobRun.LIFECYCLE_STATE_DELETED:
            job_run_status = jobrun.lifecycle_state
        elif self._if_eval_artifact_exist(model):
            job_run_status = JobRun.LIFECYCLE_STATE_SUCCEEDED
        else:
             job_run_status = JobRun.LIFECYCLE_STATE_FAILED
```

However for the last else condition it would make sense to me to add UNKNOWN status, because we really don't know the actual status. What is the status==CANCELED, but we will return FAILED instead. For now I think we can move forward, but add TODO to revisit this.

ads/aqua/evaluation.py
```diff
@@ -67,6 +62,11 @@
 from ads.model.model_version_set import ModelVersionSet
 from ads.telemetry import telemetry
 
+EVAL_TERMINATION_STATE = [
+    JobRun.LIFECYCLE_STATE_SUCCEEDED,
+    JobRun.LIFECYCLE_STATE_FAILED,
```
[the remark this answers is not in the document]
mingkang111: CANCELED will show as CANCELED. However, if the jobrun being deleted, it will convert to DELETED. Added todo to revisit a way to keep as CANCELED state.

ads/aqua/evaluation.py
```diff
@@ -349,6 +350,7 @@ class AquaEvaluationApp(AquaApp):
 
     _report_cache = TTLCache(maxsize=10, ttl=timedelta(hours=5), timer=datetime.now)
     _metrics_cache = TTLCache(maxsize=10, ttl=timedelta(hours=5), timer=datetime.now)
+    _eval_cache = TTLCache(maxsize=100, ttl=timedelta(hours=5), timer=datetime.now)
```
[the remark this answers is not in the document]
mingkang111: Yes. changed to 200 and 10hrs.

ads/aqua/evaluation.py
```diff
@@ -894,29 +896,76 @@ def list(
 
         mapping = self._prefetch_resources(compartment_id)
```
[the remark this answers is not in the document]
mingkang111: yes. Based on current size of resources, getting ~400 resources takes <1s.

ads/aqua/evaluation.py
mingkang111: Here we are try to retrieve source model name based on the id (when name is not in custometa). However, `resource_type` can have `int` as suffix for the case of datasciencemodeldeployment. We need to revisit the mapping later. Currently will have no impact as source model name always in metadata in INT environment, and in production, it will no have `int` suffix.

ads/aqua/evaluation.py
mingkang111: Added

ads/aqua/evaluation.py
mingkang111: As discussed, I will keep the current logic and add todo to revisit the logic when refactoring the code.

mrDzurb (approved):

PR #708, opened 2024-03-19
adding all original columns in anomaly outputs

govarsha:
Outliers
image
image

Inliers
image
image


ads/opctl/operator/lowcode/common/data.py
```diff
@@ -26,6 +26,20 @@ def __init__(self, spec: dict, name="input_data"):
         self.name = name
         self.load_transform_ingest_data(spec)
 
+    def get_raw_data_by_cat(self, category):
+        import pandas as pd
```
ahosler, line 30: Can we move this import outside of the method?

govarsha: Done

ahosler (approved):

ads/opctl/operator/lowcode/common/transformations.py
```diff
@@ -189,3 +195,25 @@ def _check_historical_dataset(self, df):
...
+             "Product Category 2__A__2": {
+                "PPG_Code": "Product Category 2",
+                "Class": "A",
+                "Num": 2
+            },
+            
+        }
+    """
+    def get_target_category_columns_map(self):
+        return self._target_category_columns_map
```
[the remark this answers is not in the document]
govarsha: I wanted to create the map when we are merging the columns and use that map only. Since the merging is done in transformations, I had to keep this function so that I can use the map in data.py
 We can think of some other way also, if this seems like an issue.

PR #713, opened 2024-03-19
Use common threadpool - improvements

VipulMascarenhas:
For the two places where we run queries in parallel, we can use common thread pool executor. Updated for list_buckets and list_evaluations API.

```
http://localhost:8888/aqua/buckets/?versioned=True
http://localhost:8888/aqua/evaluation/
```


mrDzurb (approved):

VipulMascarenhas:
closing older PR, will have a separate one to cover this in subsequent refactor efforts.

PR #725, opened 2024-03-20
Aqua/fix/not copy license tag

mingkang111:


ads/aqua/finetune.py
```diff
@@ -433,6 +431,9 @@ def create(
             for metadata in ft_model_custom_metadata.to_dict()["data"]
         ]
 
+        if source.freeform_tags:
+            source.freeform_tags.pop(Tags.LICENSE)
```
lu-ohai, line 435: `source.freeform_tags.pop(Tags.LICENSE.value)` ?

lu-ohai (changes requested):

ads/aqua/finetune.py
mingkang111: Good catch! Fixed.

ads/aqua/finetune.py
```diff
@@ -433,6 +431,9 @@ def create(
             for metadata in ft_model_custom_metadata.to_dict()["data"]
         ]
 
+        if source.freeform_tags:
```
VipulMascarenhas, line 434: can we also add `Tags.LICENSE.value in source.freeform_tags` check here? Service models will have this all the time, but if some users are trying this using CLI, they might try to create a FT model which doesn't have this tag.

mingkang111: I added the check. But we don't need to check if value in tags. `pop` with default value will handle it.

lu-ohai (approved):

VipulMascarenhas (approved):

PR #727, opened 2024-03-20
Load compartments from mapping file

VipulMascarenhas:
This PR covers the following:

- Load ODSC_MODEL_COMPARTMENT_OCID constant from config file, else default to env var.
- remove unused constants - AQUA_MODEL_DEPLOYMENT_IMAGE

### DIFF
--- a/ads/config.py
+++ b/ads/config.py
@@ -8,7 +8,7 @@ import contextlib
 import inspect
 import os
 from typing import Dict, Optional
-
+import json
 from ads.common.config import DEFAULT_CONFIG_PATH, DEFAULT_CONFIG_PROFILE, Config, Mode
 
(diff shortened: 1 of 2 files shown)


ads/config.py
```diff
@@ -86,6 +84,15 @@
 AQUA_SERVICE_NAME = "aqua"
 DATA_SCIENCE_SERVICE_NAME = "data-science"
 
+with open(os.path.join(AQUA_CONFIG_FOLDER, "compartment_mapping.json")) as f:
+    compartment_mapping = json.load(f)
+default_compartment = compartment_mapping.get(CONDA_BUCKET_NS)
```
mayoor, line 89: Let us not get rid of the logic to fetch from `ODSC_MODEL_COMPARTMENT_OCID`. We should always fall back to `ODSC_MODEL_COMPARTMENT_OCID`. That way, we will not need a code change when new realm support is added as they will have different namespace.

PR #732, opened 2024-03-20
Update md tag name

VipulMascarenhas:


ads/aqua/deployment.py
```diff
@@ -122,14 +121,13 @@ def from_oci_model_deployment(
         aqua_service_model_tag = freeform_tags.get(
             Tags.AQUA_SERVICE_MODEL_TAG.value, None
         )
+        aqua_model_name = freeform_tags.get(Tags.AQUA_MODEL_NAME_TAG.value, None)
```
lu-ohai, line 124: Remember to be told to use empty string instead of `None`

ads/aqua/deployment.py
```diff
@@ -264,6 +254,8 @@ def create(
             if tag in aqua_model.freeform_tags:
                 tags[tag] = aqua_model.freeform_tags[tag]
 
+        tags.update({Tags.AQUA_MODEL_NAME_TAG: aqua_model.display_name})
```
lu-ohai, line 257: ```tags.update({Tags.AQUA_MODEL_NAME_TAG.value: aqua_model.display_name})```

lu-ohai (changes requested):

ads/aqua/deployment.py
VipulMascarenhas: ah yes. updated.

ads/aqua/deployment.py
VipulMascarenhas: yep, realized while testing. fixed.

lu-ohai (approved):

PR #736, opened 2024-03-21
Fix unit test

VipulMascarenhas:


tests/unitary/with_extras/aqua/test_model.py
```diff
@@ -75,6 +75,7 @@ def setUp(self):
 
     @classmethod
     def setUpClass(cls):
+        os.environ["CONDA_BUCKET_NS"] = "test-namespace"
```
mingkang111, line 78: I would suggest to remove the env setup after finishing the test to avoid having impact on other tests.

mingkang111 (approved):

liudmylaru (approved):

PR #746, opened 2024-03-25
Compatibility check for Aqua extension and update compartment mapping config

VipulMascarenhas:
Compartment mapping will now be loaded from container_index.json.

```
{
  "odsc-llm-evaluate": [
    {
      "name": "dsmc://odsc-llm-evaluate",
      "version": "x.x.x"
    }
  ],
  "odsc-llm-fine-tuning": [
    {
      "name": "dsmc://odsc-llm-fine-tuning",
      "version": "x.x.x"
    }
  ],
  "odsc-vllm-serving": [
    {
      "name": "dsmc://odsc-vllm-serving",
      "version": "x.x.x"
    }
  ],
  "service-model-compartment": {
    "namespace1": "ocid1.compartment.oc1..<OCID>",
    "namespace2": "ocid1.compartment.oc1..<OCID>"
  }
}
```

### DIFF
--- a/ads/aqua/__init__.py
+++ b/ads/aqua/__init__.py
@@ -6,7 +6,20 @@
 
 import logging
 import sys
+import os
+from ads.aqua.utils import fetch_service_compartment
+
 
 logger = logging.getLogger(__name__)
 handler = logging.StreamHandler(sys.stdout)
 logger.setLevel(logging.INFO)
+
+
+ODSC_MODEL_COMPARTMENT_OCID = os.environ.get("ODSC_MODEL_COMPARTMENT_OCID")
+if not ODSC_MODEL_COMPARTMENT_OCID:
+    try:
+        ODSC_MODEL_COMPARTMENT_OCID = fetch_service_compartment()
+    except:
+        logger.error(
+            "ODSC_MODEL_COMPARTMENT_OCID environment variable is not set for Aqua."
+        )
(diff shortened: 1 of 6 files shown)


ads/aqua/__init__.py
```diff
@@ -6,7 +6,20 @@
...
 handler = logging.StreamHandler(sys.stdout)
 logger.setLevel(logging.INFO)
+
+
+ODSC_MODEL_COMPARTMENT_OCID = os.environ.get("ODSC_MODEL_COMPARTMENT_OCID")
+if not ODSC_MODEL_COMPARTMENT_OCID:
+    try:
+        ODSC_MODEL_COMPARTMENT_OCID = fetch_service_compartment()
+    except:
+        logger.error(
```
mrDzurb, line 23: The ```ODSC_MODEL_COMPARTMENT_OCID``` will be None in this case, right? Is it fine?

mayoor (dismissed):

mrDzurb (approved):

mrDzurb (approved):

ads/aqua/__init__.py
VipulMascarenhas: yes - API calls will fail in that case, but for now we assume that either we set ODSC_MODEL_COMPARTMENT_OCID explicitly or index file will have it. Is there a way to enforce this?

mayoor (approved):

PR #753, opened 2024-04-05
Adds support for mlforecast in the forecasting operator

codeloop:
Add support for mlforecast in the forecasting operator
- Feature for modelling using recursive and per step model for forecast horizon
- Add arguments for the data preprocessing and feature engineering
- Add the report for model with historical fitted values
- Add the third party library license information
- Migrate from datapane to report_creator


ads/opctl/operator/lowcode/forecast/environment.yaml
```diff
@@ -8,6 +8,7 @@ dependencies:
       - oracle-ads>=2.9.0
       - prophet
       - neuralprophet
+      - mlforecast
```
mrDzurb, line 11: Do we need BA for this?

mrDzurb (approved):

ads/opctl/operator/lowcode/forecast/model/forecast_datasets.py
```diff
@@ -158,6 +159,15 @@ def get_all_data_long(self, include_horizon=True):
             on=[self._datetime_column_name, ForecastOutputColumns.SERIES],
         ).reset_index()
 
+    def get_all_data_long_test(self):
```
ahosler, line 162: What's the meaning of "test" here?

ads/opctl/operator/lowcode/forecast/environment.yaml
codeloop: I have applied for LT & BA, awaiting legal review to complete.

ads/opctl/operator/lowcode/forecast/model/forecast_datasets.py
codeloop: I was referring to the just the forecast horizon part here, updated the function signature to better name. Thanks!

darenr:
What additional features does the `mlforecast` library add to the operator? Is there a benchmark where this shows improvement over `prophet` - I'm just curious what the rationale was for adding this. Not saying it's not a good idea, just curious what the reasons are?

codeloop:
> What additional features does the `mlforecast` library add to the operator? Is there a benchmark where this shows improvement over `prophet` - I'm just curious what the rationale was for adding this. Not saying it's not a good idea, just curious what the reasons are?

Yes, we do have time and metric benchmark for mlforecast against prophet, it offers an improvement over it in both the measures,
`lgbm with MLForecast has a speedup of 20.95 compared with prophet`

`We can see that lgbm with MLForecast was able to provide metrics at least twice as good as Prophet as seen in the column improvement above, and way faster.`
more detailed analysis for a commonly benchmarked electricity load forecasting dataset @ https://nixtlaverse.nixtla.io/mlforecast/docs/tutorials/electricity_load_forecasting.html#comparison-with-prophet.
We want to add a boosting-based forecasting model, mlforecast offers this along with support for feature engineering and scalability options.

ahosler (approved):
Awesome work. Good job pushing this through

darenr (approved):

PR #757, opened 2024-04-08
Websocket handlers for AQUA

harsh97:
This PR adds websocket support for AQUA communication with jupyterlab. We are starting with initial support for listing of evaluations


ads/aqua/decorator.py
```diff
@@ -67,32 +74,32 @@ def inner_function(self: AquaAPIhandler, *args, **kwargs):
             MissingEndpointForNonRegionalServiceClientError,
             RequestException,
         ) as error:
-            self.write_error(
+            return self.write_error(
```
mrDzurb, line 77: If we add return here, why don't we do similar for the line 66?

ads/aqua/extension/__init__.py
```diff
@@ -14,6 +14,7 @@
 from ads.aqua.extension.finetune_handler import __handlers__ as __finetune_handlers__
 from ads.aqua.extension.model_handler import __handlers__ as __model_handlers__
 from ads.aqua.extension.ui_handler import __handlers__ as __ui_handlers__
+from ads.aqua.extension.aqua_web_socket_handler import __handlers__ as __ws_handlers__
```
mrDzurb, line 17: I think we should keep ui_ in the file name, maybe - ```ui_websocket_handler```

ads/aqua/extension/aqua_web_socket_handler.py
```diff
@@ -0,0 +1,55 @@
...
+)
+from tornado.websocket import WebSocketHandler
+
+from ads.aqua.extension.models.WSModels import BaseRequest, RequestResponseType
+
+
+class AquaWebSocketHandler(WebSocketHandler):
+    """Handler for Aqua Websocket."""
+
+    _handlers_: List[type(AquaWSMsgHandler)] = [AquaEvaluationWSMsgHandler]
```
mrDzurb, line 25: Why do we need to use type here? Just curious, wouldn't it work - ```List[AquaWSMsgHandler]```?

ads/aqua/extension/aqua_web_socket_handler.py
```diff
@@ -0,0 +1,55 @@
...
+
+    thread_pool: ThreadPoolExecutor
+
+    def open(self, *args, **kwargs):
+        self.thread_pool = concurrent.futures.ThreadPoolExecutor(max_workers=20)
```
mrDzurb, line 30: NIT: ```self.thread_pool = ThreadPoolExecutor(max_workers=20)```

mrDzurb: Please move the ```max_workers```  to the constant variable.

ads/aqua/extension/aqua_web_socket_handler.py
```diff
@@ -0,0 +1,55 @@
...
+        logger.info("WebSocket opened")
+
+    def on_message(self, message: Union[str, bytes]):
+        logging.debug(f"Received message: {message}")
+        request = BaseRequest.from_json(message)
+        for handler in self._handlers_:
```
mrDzurb, line 36: Could you add some description for this loop, what we are doing here?

ads/aqua/extension/aqua_web_socket_handler.py
```diff
@@ -0,0 +1,55 @@
...
+            if request.kind in handler.get_message_types():
+                message_handler = handler(message)
+                future: asyncio.futures.Future = self.thread_pool.submit(
+                    message_handler.process
+                )
+                future.add_done_callback(self.on_message_processed)
+                break
+        print(request)
```
mrDzurb, line 44: probably should go either to debug or should be removed?

ads/aqua/extension/aqua_web_socket_handler.py
```diff
@@ -0,0 +1,55 @@
...
+    def on_message(self, message: Union[str, bytes]):
+        logging.debug(f"Received message: {message}")
+        request = BaseRequest.from_json(message)
+        for handler in self._handlers_:
+            if request.kind in handler.get_message_types():
+                message_handler = handler(message)
+                future: asyncio.futures.Future = self.thread_pool.submit(
+                    message_handler.process
+                )
+                future.add_done_callback(self.on_message_processed)
```
mrDzurb, line 42: What is some error occurs during the execution feature?

ads/aqua/extension/models/WSModels.py
mrDzurb, line 1: let's follow one convention for the naming files :)

pyproject.toml
```diff
@@ -180,7 +180,7 @@ pii = [
   "spacy==3.6.1",
 ]
 llm = ["langchain>=0.1.10", "evaluate>=0.4.0"]
-aqua = ["jupyter_server"]
+aqua = ["jupyter_server", "dataclass-wizard"]
```
mrDzurb, line 183: I guess we will need to submit BA for this. However in ADS we are already use ```DataClassSerializable``` class allowing us to serialize/deserialize dataclasses to/from JSON/YAML/DICT

ads/aqua/extension/aqua_web_socket_handler.py
harsh97: I added type here since we are just adding the class instead of instance of the class.

ads/aqua/extension/aqua_web_socket_handler.py
harsh97: Added check for the same!! Thanks!

pyproject.toml
harsh97: I have removed the need for htis library. This library helped us in keeping snake case naming in python and camelCase naming in typescript.

ads/aqua/extension/aqua_web_socket_handler.py
harsh97: To clarify: The check is in the similar way like we have for our other apis with @handle_exceptions decorator. If the function still throws error we aren't handling that which is similar to the current handling logic.

harsh97: I removed our error handling logic from "handle_exceptions" decorator and discovered that with normal REST calls we are still able to get "internal server error" as response. This was because we have inbuiilt error handling coming from tornado for this purpose. I have incorporated this logic in our websocket handler and now we are getting similar behaviour to our REST APIs.

VipulMascarenhas (approved):
lgtm 👍

PR #759, opened 2024-04-08
Update cli import

VipulMascarenhas:
While using ADS CLI, if ODSC_MODEL_COMPARTMENT_OCID is not set and if the aqua service compartment is not accessible, ads writes the following in the logs. This PR only reads aqua modules when aqua cli is invoked.

```
ERROR:ODSC_AQUA:Failed to read file oci://service-managed-models@id19sfcrra6z/service_models/config/container_index.json.
ERROR:ads.aqua:ODSC_MODEL_COMPARTMENT_OCID environment variable is not set for Aqua, due to .
```


mrDzurb:
DO we consider this as a hot fix and want to publish this version ASAP?

mrDzurb (approved):

darenr (approved):

PR #761, opened 2024-04-08
bugfix/aqua.evaluation 

mingkang111:
# Description
This PR aims to fix the following problem:
* potential NPE
* AttributeError cannot be caught correctly

### DIFF
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -1043,7 +1043,6 @@ class AquaEvaluationApp(AquaApp):
                 "Please check if the OCID is correct."
             )
 
-        # TODO: add job_run_id as input param to skip the query below
         model_provenance = self.ds_client.get_model_provenance(eval_id).data
 
         jobrun_id = model_provenance.training_id
@@ -1394,7 +1393,6 @@ class AquaEvaluationApp(AquaApp):
 
     def load_evaluation_config(self, eval_id):
         """Loads evaluation config."""
-        # TODO
         return {
             "model_params": {
                 "max_tokens": 500,
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -833,13 +833,13 @@ class AquaEvaluationApp(AquaApp):
         logger.info(f"Fetching evaluation: {eval_id} details ...")
 
         resource = utils.query_resource(eval_id)
-        model_provenance = self.ds_client.get_model_provenance(eval_id).data
-
         if not resource:
             raise AquaRuntimeError(
                 f"Failed to retrieve evalution {eval_id}."
                 "Please check if the OCID is correct."
             )
+        model_provenance = self.ds_client.get_model_provenance(eval_id).data
+
         jobrun_id = model_provenance.training_id
         job_run_details = self._fetch_jobrun(
             resource, use_rqs=False, jobrun_id=jobrun_id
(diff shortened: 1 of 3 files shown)


VipulMascarenhas (approved):

ads/aqua/evaluation.py
```diff
@@ -1265,7 +1273,10 @@ def cancel(self, eval_id) -> dict:
             raise AquaRuntimeError(
                 f"Failed to get evaluation details for model {eval_id}"
             )
-        job_run_id = model.provenance_metadata.training_id
+
+        job_run_id = (
```
mingkang111, line 1277: This is the fix for the existing potential bug in production: when `model.provenance_metadata` is None, the function will break.

ads/aqua/evaluation.py
```diff
@@ -1328,7 +1339,7 @@ def delete(self, eval_id):
             job_id = model.custom_metadata_list.get(
                 EvaluationCustomMetadata.EVALUATION_JOB_ID.value
             ).value
-        except ValueError:
+        except Exception:
```
mingkang111, line 1342: `model.custom_metadata_list` can be None for DataScienceModel, which will give 500 `AttributeError` from ads. We need to make it as `AquaMissingKeyError` as well.

ads/aqua/evaluation.py
```diff
@@ -1538,20 +1550,6 @@ def _build_resource_identifier(
             )
             return AquaResourceIdentifier()
 
-    def _get_jobrun(
```
mingkang111, line 1541: The function has not been used. So it is safe to remove it.

VipulMascarenhas (changes requested):
can we revert the changes in evaluation.py and add the changes separately later on? This PR will change the current behavior in production even if the intention is to fix bugs. All these changes can go as a part of the next set of improvements/bug fixes later on.

mingkang111:
> can we revert the changes in evaluation.py and add the changes separately later on? This PR will change the current behavior in production even if the intention is to fix bugs. All these changes can go as a part of the next set of improvements/bug fixes later on.

These changes as I commented, are worthwhile to be fixed. They didn't touch the main logic. I believe it is the reason for writing test: to find the bug and fix them. I will keep this PR open to be merged later. 
I also prepared another PR: https://github.com/oracle/accelerated-data-science/pull/763 which is the version that adding test without the change in `evaluation.py`. In case @mayoor prefer to merge those first.

VipulMascarenhas (approved):
thanks for this PR, changes look good.

ads/aqua/evaluation.py
```diff
@@ -833,13 +833,13 @@ def get(self, eval_id) -> AquaEvaluationDetail:
...
 
         resource = utils.query_resource(eval_id)
-        model_provenance = self.ds_client.get_model_provenance(eval_id).data
-
         if not resource:
             raise AquaRuntimeError(
                 f"Failed to retrieve evalution {eval_id}."
                 "Please check if the OCID is correct."
             )
+        model_provenance = self.ds_client.get_model_provenance(eval_id).data
```
mrDzurb, line 841: Can we guarantee that the ```.data``` will be available always?  What will happen in negative scenario?

ads/aqua/evaluation.py
```diff
@@ -1294,7 +1294,10 @@ def cancel(self, eval_id) -> dict:
             raise AquaRuntimeError(
                 f"Failed to get evaluation details for model {eval_id}"
             )
-        job_run_id = model.provenance_metadata.training_id
+
+        job_run_id = (
+            model.provenance_metadata.training_id if model.provenance_metadata else None
```
mrDzurb, line 1299: Can't the ```model.provenance_metadata``` be None by itself?

mingkang111: Hi @mrDzurb , the model here is ads `DataScienceModel`, which is guaranteed to have `provenance_metadata` property.

ads/aqua/evaluation.py
mingkang111: Based on how oci python sdk implement it, as long as `self.ds_client.get_model_provenance(eval_id)` gives response, it will have `data` property. See https://github.com/oracle/oci-python-sdk/blob/master/src/oci/base_client.py#L747 and https://github.com/oracle/oci-python-sdk/blob/master/src/oci/response.py#L8

We have verify this before as well: https://github.com/oracle/accelerated-data-science/blob/main/ads/model/service/oci_datascience_model.py#L240

PR #765, opened 2024-04-09
Add AI quick actions link

mayoor:
image


darenr (approved):
> Deploy, Fine Tune, and

ideally a comma after Tune

Also you prefer linking to the markdown docs rather than service docs?

PR #779, opened 2024-04-16
Fix telemetry event record to better report output

liudmylaru:
### Description

Telemetry data related to shape model combination for MD and finetune has a bad design so that visualized report not easily build out for this metrics. Suggesting an update, so that we will do:

```
aqua/custom/deployment/create/shape/{shape}
{value=model_name}
aqua/service/deployment/create/shape/{shape}
{value=model_name}
aqua/service/finetune/create/shape/{shape}
{value=model_name}
```

instead:
```
aqua/custom/{model_name}/deployment/create/shape/{shape}
aqua/service/{model_name}/deployment/create/shape/{shape}
aqua/service/{model_name}/finetune/create/shape/{shape}
```

Also dropping:
```aqua/service/deployment/create/shape/{shape}```, this will duplicate suggested change above, if used without value
```aqua/custom/deployment/create/shape/{shape}```, this will duplicate suggested change above, if used without value

### DIFF
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -673,6 +673,14 @@ class AquaEvaluationApp(AquaApp):
             ),
         )
 
+        # tracks shapes used in evaluation that were created for the given evaluation source
+        self.telemetry.record_event_async(
+            category="aqua/evaluation/create",
+            action="shape",
+            detail=create_aqua_evaluation_details.shape_name,
+            value=evaluation_source.display_name
+        )
+
         # tracks unique evaluation that were created for the given evaluation source
         self.telemetry.record_event_async(
             category="aqua/evaluation",
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -688,7 +688,7 @@ class AquaEvaluationApp(AquaApp):
             category="aqua/evaluation/create",
             action="shape",
             detail=create_aqua_evaluation_details.shape_name,
-            value=evaluation_source.display_name,
+            value=self._get_service_model_name(evaluation_source),
         )
 
         # tracks unique evaluation that were created for the given evaluation source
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by qiuosier — DISMISSED** [review]

File: ads/aqua/evaluation.py [hunk 1]
```diff
@@ -673,6 +673,14 @@ def create(
             ),
         )
 
+        # tracks shapes used in evaluation that were created for the given evaluation source
+        self.telemetry.record_event_async(
+            category="aqua/evaluation/create",
+            action="shape",
+            detail=create_aqua_evaluation_details.shape_name,
+            value=evaluation_source.display_name
```
VipulMascarenhas, line 681: no action needed in this PR, `evaluation_source.display_name` will be replaced with base model name. Lu will be updating this in this PR later this week.

ads/aqua/finetune.py
```diff
@@ -470,6 +471,14 @@ def create(
             category="aqua/service/finetune",
             action="create",
             detail=source.display_name,
+            **telemetry_kwargs,
+        )
+        # track combination of model and shape used for fine-tune creation
+        self.telemetry.record_event_async(
```
VipulMascarenhas, line 477: I'm thinking if this will duplicate the telemetry data we have in line 464 while doing post processing. Not a dealbreaker, we have to ensure this is bucketed into different category. 

`aqua/service/finetune/create/shape/{create_fine_tuning_details.shape_name}x{create_fine_tuning_details.replica}", value=**telemetry_kwargs`

`aqua/service/finetune/create/shape/{create_fine_tuning_details.shape_name}x{create_fine_tuning_details.replica}", value=source.display_name`

liudmylaru: no, it is not a duplicate. the idea is to have 3 vents recorded:

1. FT create with shape (and 6 chars of ocid for future track)
2. FT create with model_name (and 6 chars of ocid for future track)
3. FT create with shape and model to see which models used with which shapes (no ocid chars, as we looking into combination of model/shape)

So they are seems to be the same.

VipulMascarenhas (approved):

mrDzurb (approved):

PR #781, opened 2024-04-17
Updated source name for telemetry

lu-ohai:


ads/aqua/evaluation.py
```diff
@@ -767,6 +768,34 @@ def _build_evaluation_runtime(
...
+            The service model name of source.
+        """
+        if isinstance(source, ModelDeployment):
+            fine_tuned_model_tag = source.freeform_tags.get(
+                Tags.AQUA_FINE_TUNED_MODEL_TAG.value, UNKNOWN
+            )
+            if not fine_tuned_model_tag:
+                return source.freeform_tags.get(Tags.AQUA_MODEL_NAME_TAG.value)
+            else:
+                return extract_id_and_name_from_tag(fine_tuned_model_tag)[1]
```
VipulMascarenhas, line 796: since extract_id_and_name_from_tag function is not used anywhere else in the code, can we update that 
function to remove the is_valid_ocid check call?

VipulMascarenhas (commented):
added one minor comment to remove is_valid_ocid check. Also, you can update this PR to merge to feature/aquav1.0.1 branch.

ads/aqua/evaluation.py
lu-ohai: Since this part is not blocking, it may help us better find out the issue if there is any mis format happened.

qiuosier:
Could you do the same update for fine-tuning? We may want to support fine-tuning custom models in the future.

lu-ohai:
> Could you do the same update for fine-tuning? We may want to support fine-tuning custom models in the future.

Do we have a design for this already? If we support custom model, does it also copied from service model? Also, will the custom model has a specific tag to tell from service model? @qiuosier

VipulMascarenhas:
> > Could you do the same update for fine-tuning? We may want to support fine-tuning custom models in the future.
> 
> Do we have a design for this already? If we support custom model, does it also copied from service model? Also, will the custom model has a specific tag to tell from service model? @qiuosier

@lu-ohai let's add fine-tuning one in a separate PR.

VipulMascarenhas (approved):

PR #782, opened 2024-04-17
Bugfix/aqua error handle

mingkang111:
# Description
This PR aims to improve handling error. 

# Changed 
* Handle HTTPError in `handle_exception` decorator
* Add test for decorator
* Move decorator to GET/POST/PUT/DELETE method in handler
* Add name in each test for better tracing what being test
* As we are using tornado framework, tornado's HTTPError should be used instead of requests.HTTPError

# Test
Screenshot 2024-04-17 at 15 53 37
Screenshot 2024-04-17 at 15 53 45

### DIFF
--- a/ads/aqua/extension/base_handler.py
+++ b/ads/aqua/extension/base_handler.py
@@ -11,11 +11,12 @@ from dataclasses import asdict, is_dataclass
 from typing import Any
 
 from notebook.base.handlers import APIHandler
-from tornado.web import HTTPError, Application
 from tornado import httputil
-from ads.telemetry.client import TelemetryClient
-from ads.config import AQUA_TELEMETRY_BUCKET, AQUA_TELEMETRY_BUCKET_NS
+from tornado.web import Application, HTTPError
+
 from ads.aqua import logger
+from ads.config import AQUA_TELEMETRY_BUCKET, AQUA_TELEMETRY_BUCKET_NS
+from ads.telemetry.client import TelemetryClient
 
 
 class AquaAPIhandler(APIHandler):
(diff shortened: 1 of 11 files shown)


ads/aqua/extension/base_handler.py
```diff
@@ -66,12 +68,15 @@ def finish(self, payload=None):  # pylint: disable=W0221
 
     def write_error(self, status_code, **kwargs):
         """AquaAPIhandler errors are JSON, not human pages."""
-
         self.set_header("Content-Type", "application/json")
         reason = kwargs.get("reason")
         self.set_status(status_code, reason=reason)
         service_payload = kwargs.get("service_payload", {})
-        message = self.get_default_error_messages(service_payload, str(status_code))
+        default_msg = responses.get(status_code, "Unknown HTTP Error")
```
mayoor, line 75: Why are we using **kwargs for write_error instead of putting in the parameters with default value?

mayoor (approved):

VipulMascarenhas (approved):
will test the remaining handlers once this is merged.

ads/aqua/extension/base_handler.py
mingkang111: Hi @mayoor , this is also doable. The `write_error` here is try to keep the same argument with the parent class. Like how `APIHandler` overide `IpythonHandler` and how `IpythonHandler` override Tornado's `RequestHandler`.  
If you prefer the way to use explicit argument with default value at here, we can change it too.

qiuosier (approved):

PR #785, opened 2024-04-18
[ODSC-56061] Update Model deployment telemetry

VipulMascarenhas:
### Description

This PR takes care of the following:
1. Simplify and remove redundant code in create method.
2. Add "ocid=last_8_digits_of_MD_ID" as telemetry value when creating MD. This gets us logs like:
```
{"datetime": 1712689489086, "category": "aqua/service/deployment/create", "action": "model-name", "value": "ocid=3kmulq12", ...
```
or 
```
{"datetime": 1712689489086, "category": "aqua/custom/deployment/create", "action": "ft-base-model-name", "value": "ocid=3kmulq12", ...
```

3. Add ocid and lifecycle state when listing deployments. This logs telemetry information only if state is active or failed to avoid unnecessary logging of other states that won't be much useful to us.This gives us logs like:
```
{"datetime": 1712689489086, "category": "aqua/deployment/list", "action": "3kmulq12", "value": "ACTIVE", ...
{"datetime": 1712689489086, "category": "aqua/deployment/list", "action": "3kmulq12", "value": "FAILED", ...
```

### DIFF
--- a/ads/aqua/deployment.py
+++ b/ads/aqua/deployment.py
@@ -391,40 +391,30 @@ class AquaDeploymentApp(AquaApp):
             .with_runtime(container_runtime)
         ).deploy(wait_for_completion=False)
 
-        if is_fine_tuned_model:
-            # tracks unique deployments that were created in the user compartment
-            self.telemetry.record_event_async(
-                category="aqua/custom/deployment", action="create", detail=model_name
-            )
-            # tracks the shape used for deploying the custom models
-            self.telemetry.record_event_async(
-                category="aqua/custom/deployment/create",
-                action="shape",
-                detail=instance_shape,
-            )
-            # tracks the shape used for deploying the custom models by name
-            self.telemetry.record_event_async(
-                category=f"aqua/custom/{model_name}/deployment/create",
-                action="shape",
-                detail=instance_shape,
-            )
-        else:
-            # tracks unique deployments that were created in the user compartment
-            self.telemetry.record_event_async(
-                category="aqua/service/deployment", action="create", detail=model_name
-            )
-            # tracks the shape used for deploying the service models
-            self.telemetry.record_event_async(
-                category="aqua/service/deployment/create",
-                action="shape",
-                detail=instance_shape,
-            )
-            # tracks the shape used for deploying the service models by name
-            self.telemetry.record_event_async(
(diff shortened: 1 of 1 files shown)


ads/aqua/deployment.py
```diff
@@ -391,40 +391,30 @@ def create(
...
+            {"ocid": deployment_id[-8:]} if len(deployment_id) > 8 else {}
+        )
+        # tracks unique deployments that were created in the user compartment
+        self.telemetry.record_event_async(
+            category=f"aqua/{model_type}/deployment",
+            action="create",
+            detail=model_name,
+            **telemetry_kwargs,
+        )
+        # tracks the shape used for deploying the custom or service models
```
qiuosier, line 406: I think, for these two events:
```
aqua/{model_type}/deployment/create/shape/{shape_name}
aqua/{model_type}/{model_name}/deployment/create/shape/{shape_name}
```
Some additional post processing in the logs of the second event can give the counts for the first one?

VipulMascarenhas: right, for our internal report creator we can do that, but Liuda mentioned that we'd probably need this separate for the way the data is needed for horizon dashboard. I can remove that, @liudmylaru can you confirm if we still need the first one?

ads/aqua/deployment.py
```diff
@@ -391,40 +391,30 @@ def create(
...
-            # tracks the shape used for deploying the service models by name
-            self.telemetry.record_event_async(
-                category=f"aqua/service/{model_name}/deployment/create",
-                action="shape",
-                detail=instance_shape,
-            )
+        model_type = "custom" if is_fine_tuned_model else "service"
+        deployment_id = deployment.dsc_model_deployment.id
+        telemetry_kwargs = (
+            {"ocid": deployment_id[-8:]} if len(deployment_id) > 8 else {}
```
mrDzurb, line 397: Just Curious, what we are checking here? Why if the length less than 8 we don't send ocid at all?

mrDzurb (approved):

ads/aqua/deployment.py
VipulMascarenhas: it's just a sanity check to make sure the string slicing won't fail. I wasn't sure if we'll ever have a case where deployment_id is ever going to be different than 60 characters like we have now.

PR #796, opened 2024-04-19
Updated error message.

lu-ohai:


ads/aqua/__init__.py
```diff
@@ -26,3 +26,8 @@
         logger.error(
```
VipulMascarenhas, line 26: logging this inside Exception wouldn't cover the case if `fetch_service_compartment` returns None. Can we do something like:
```
except Exception as e:
    pass
if not ODSC_MODEL_COMPARTMENT_OCID:
     logger.error(
        f"ODSC_MODEL_COMPARTMENT_OCID environment variable is not set for AQUA."
    )
    if NB_SESSION_OCID:
        logger.error(
            "AQUA is not available for this notebook session."
        )
    exit()
```

lu-ohai: Oh I was updating it the same as you suggested in the ticket. Could you be more specific or we can have a zoom call on this?

mingkang111:
Can you please rebase the `feature/aquav1.0.1` ? The `feature/aquav1.0.1` just sync with `main`.

ads/aqua/__init__.py
lu-ohai: Just updated. Could you take a look now?

VipulMascarenhas (approved):
changes look good, thank you!

PR #797, opened 2024-04-19
Avoid deactivate/activate when feature flag for Aqua is enabled.

mayoor:


darenr (approved):

VipulMascarenhas (commented):
@mayoor updated the PR with exception handling inside fetch_service_compartment.

ads/aqua/utils.py
```diff
@@ -584,14 +584,19 @@ def fetch_service_compartment():
         f"oci://{AQUA_SERVICE_MODELS_BUCKET}@{CONDA_BUCKET_NS}/service_models/config"
     )
 
-    config = load_config(
-        file_path=config_file_name,
-        config_file_name=CONTAINER_INDEX,
-    )
+    try:
```
mrDzurb, line 587: Could you add typing for this method and add mode details in the description?

mrDzurb (approved):

ads/aqua/utils.py
VipulMascarenhas: updated.

PR #802, opened 2024-04-22
Graceful exit for aqua cli

VipulMascarenhas:
### Description

Calling `exit()` in `__init__.py` of AquaCommand causes issues while running independent tests and loading modules, hence `exit()` is moved to AquaCommand.

### Testing
1. Env var is set 
```
> ads aqua model list
{
    "compartment_id": "ocid1.compartment.oc1..<OCID>",
    "icon": "",
...
...
```

2. Env var is not set but `fetch_default_compartment()` picks up default compartment id in the backend.
```
> ODSC_MODEL_COMPARTMENT_OCID="" ads aqua model list
{
    "compartment_id": "ocid1.compartment.oc1..<OCID>",
    "icon": "",
...
...
```

3. Env var is not set and service bucket is set incorrectly, i.e. ODSC_MODEL_COMPARTMENT_OCID is `None`.
```
> ODSC_MODEL_COMPARTMENT_OCID="" AQUA_SERVICE_MODELS_BUCKET="test" ads aqua model list
ERROR:ads.aqua:ODSC_MODEL_COMPARTMENT_OCID environment variable is not set for Aqua.
```

### Unit Tests
```
> python3 -m pytest tests/unitary/with_extras/aqua/test_cli.py
===================================================== test session starts ======================================================
platform darwin -- Python 3.8.18, pytest-7.4.0, pluggy-1.0.0 -- /Users/vmascarenhas/miniconda3/envs/ads-local-v38/bin/python3
cachedir: .pytest_cache
rootdir: /Users/vmascarenhas/workspace/git/accelerated-data-science
configfile: pytest.ini
collected 4 items

tests/unitary/with_extras/aqua/test_cli.py::TestAquaCLI::test_aqua_command_without_compartment_env_var PASSED            [ 25%]
tests/unitary/with_extras/aqua/test_cli.py::TestAquaCLI::test_aquacommand_0_default PASSED                               [ 50%]
tests/unitary/with_extras/aqua/test_cli.py::TestAquaCLI::test_aquacommand_1_set_logging_level PASSED                     [ 75%]
tests/unitary/with_extras/aqua/test_cli.py::TestAquaCLI::test_entrypoint PASSED                                          [100%]

====================================================== 4 passed in 4.80s =======================================================
```

### DIFF
--- a/ads/aqua/cli.py
+++ b/ads/aqua/cli.py
@@ -5,7 +5,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 import os
 
-from ads.aqua import ENV_VAR_LOG_LEVEL, set_log_level
+from ads.aqua import ENV_VAR_LOG_LEVEL, set_log_level, ODSC_MODEL_COMPARTMENT_OCID
 from ads.aqua.deployment import AquaDeploymentApp
 from ads.aqua.evaluation import AquaEvaluationApp
 from ads.aqua.finetune import AquaFineTuningApp
@@ -41,3 +41,6 @@ class AquaCommand:
             'WARNING', 'ERROR', and 'CRITICAL'.
         """
         set_log_level(log_level)
+        # gracefully exit if env var is not set
+        if not ODSC_MODEL_COMPARTMENT_OCID:
+            exit()
--- a/tests/unitary/with_extras/aqua/test_cli.py
+++ b/tests/unitary/with_extras/aqua/test_cli.py
@@ -4,13 +4,17 @@
 # Copyright (c) 2024 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 
+import os
 import logging
 import subprocess
+import pytest
 from unittest import TestCase
 from unittest.mock import patch
-
+from importlib import reload
 from parameterized import parameterized
 
+import ads.aqua
+import ads.config
 from ads.aqua.cli import AquaCommand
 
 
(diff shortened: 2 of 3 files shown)


darenr:
```python

sys.exit: @mock.patch('sys.exit')

```

That's often used when a function calls `exit()`

tests/unitary/with_extras/aqua/test_cli.py
```diff
@@ -24,6 +28,12 @@ class TestAquaCLI(TestCase):
         datefmt="%m/%d/%Y %I:%M:%S %p",
         level=logging.INFO,
     )
+    SERVICE_COMPARTMENT_ID = "ocid1.compartment.oc1..<OCID>"
+
+    def setUp(self):
+        os.environ["ODSC_MODEL_COMPARTMENT_OCID"] = TestAquaCLI.SERVICE_COMPARTMENT_ID
```
mrDzurb, line 34: Modifying the real env variables can potentially bring the problems to the other tests. It would be better to patch them or use fixture that could take care of restoring the original values for the env variables

```
@mock.patch.dict(os.environ, {"key": "value"})
def test_case(self):
  ....
```

OR

```
@pytest.fixture(autouse=True, scope="class")
def set_env_vars():
     with patch.dict(os.environ, clear=True):
           ....
```

ads/aqua/cli.py
```diff
@@ -41,3 +41,6 @@ def __init__(
             'WARNING', 'ERROR', and 'CRITICAL'.
         """
         set_log_level(log_level)
+        # gracefully exit if env var is not set
+        if not ODSC_MODEL_COMPARTMENT_OCID:
+            exit()
```
mrDzurb: Shouldn't be exit called with some code? Can't be the error thrown here instead of exit()?

VipulMascarenhas:
> ```python
> @mock.patch('sys.exit')
> ```
> 
> That's often used when a function calls `exit()`

nice catch, removed the assert raise and patched exit() instead.

ads/aqua/cli.py
VipulMascarenhas: I updated this with sys.exit(0) thinking we would want to do a graceful exit since we log the below error before exiting.
```
ERROR:ads.aqua:ODSC_MODEL_COMPARTMENT_OCID environment variable is not set for Aqua.
```

But now that i think of it, would it make sense to log exit 1 instead and print an additional message here instead of only logging it. I'll update the PR.

mayoor (approved):

PR #806, opened 2024-04-24
Add substring util

VipulMascarenhas:
Added util function as per [review comments](https://github.com/oracle/accelerated-data-science/pull/803#discussion_r1578210330).

### DIFF
--- a/ads/aqua/utils.py
+++ b/ads/aqua/utils.py
@@ -743,7 +743,7 @@ def known_realm():
     return os.environ.get("CONDA_BUCKET_NS") in AQUA_GA_LIST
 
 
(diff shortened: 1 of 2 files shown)


ads/aqua/utils.py
```diff
@@ -741,3 +741,9 @@ def known_realm():
 
     """
     return os.environ.get("CONDA_BUCKET_NS") in AQUA_GA_LIST
+
+
+def get_ocid_substring(ocid: str, key_len: int = 8) -> str:
```
mayoor, line 746: What is the significance of 8 characters in ocid?

VipulMascarenhas: no significance, I removed the defaults.

mayoor (approved):

PR #809, opened 2024-04-25
Signer exception fix v2.11.9

VipulMascarenhas:
### Description

The jupyter build fails as the default_signer() throws an `oci.exceptions.ConfigFileNotFound`. The `load_config` inside `fetch_service_compartment` has been updated to catch all generic exceptions.

### DIFF
--- a/ads/aqua/utils.py
+++ b/ads/aqua/utils.py
@@ -582,7 +582,7 @@ def fetch_service_compartment() -> Union[str, None]:
             file_path=config_file_name,
             config_file_name=CONTAINER_INDEX,
         )
-    except AquaFileNotFoundError:
+    except Exception:
         logger.error(
             f"Config file {config_file_name}/{CONTAINER_INDEX} to fetch service compartment OCID could not be found."
         )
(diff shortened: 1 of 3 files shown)


ads/aqua/utils.py
```diff
@@ -582,7 +582,7 @@ def fetch_service_compartment() -> Union[str, None]:
             file_path=config_file_name,
             config_file_name=CONTAINER_INDEX,
         )
-    except AquaFileNotFoundError:
+    except Exception:
```
darenr, line 585: If there's any possibility of the exception being something other than not being able to find the file we'd never know. What about get the exception "as e" and then add str(e) to the message just in case?

darenr (approved):

ads/aqua/utils.py
VipulMascarenhas: yes, will add the message to the logger..

VipulMascarenhas: added

mrDzurb (approved):

mayoor (approved):

PR #811, opened 2024-04-29
[WIP] Support for BYOM models

mayoor:
# Bring your own model Support

This support entails support for "any" huggingface model on AI quick actions. There are following scenarios that needs to be addressed to facilitate this feature - 
1. User chooses a huggingface model that has been already tested and certified on AI Quick Actions by the service. To support such models, the user need not specify which runtime container is required for inference or finetuning
2. User chooses a model that has not been certified, but user chooses from the list of readymade runtimes from the service. To support such models, the user will specify the service managed container name as available on the documentation. The documentation will list out the key libraries and their versions across different service managed container so that the user can choose the right image. 
3. Use chooses a model that has no supporting container in service managed container list. The user builds the inference/finetuning container in their tenancy and provides the the container URI while registering the model

The PR also brings in a notion of "regstering" model. What it means is that user is importing the huggingface model into model catalog within user tenancy. 

# Assumptions
1. Model registration requires internet connection. It is assumed that the surface from where the user "registers" the model has internet connection
2. For the gated model model user will have authorized huggingface token.

# Huggingface token setup
Run `huggingface-cli login` command and follow the on screen instruction

# Usage

1. Scenario 1[Verified Model]- ```ads aqua model register --model meta-llama/Meta-Llama-3-8B  --os_path oci://mayoor-dev-versioned@ociodscdev/cached-models --local_dir `pwd`/cache-models 
2. Scenario 2[Unverified Model with SMC] - ```ads aqua model register --model meta-llama/Meta-Llama-3-8B  --os_path oci://mayoor-dev-versioned@ociodscdev/cached-models --local_dir /Users/mayorao/dev/odsc/accelerated-data-science/cache-models --odsc-vllm-container --inference_container_type_smc --finetuning_container odsc-finetuning-llm-container --finetuning_container_type_smc```
3. Scenario 3[Unverified Model with Custom Container] - ```ads aqua model register --model meta-llama/Meta-Llama-3-8B  --os_path oci://mayoor-dev-versioned@ociodscdev/cached-models --local_dir /Users/mayorao/dev/odsc/accelerated-data-science/cache-models --inference_container iad.ocir.io/my/custom:1.0 --finetuning_container iad.ocir.io/my/custom-ft:1.0```

# TODO
* Test cases.


darenr (approved):

darenr:
Why'd it take so long? :)

PR #813, opened 2024-05-01
Improves the extraction of custom metadata items by key.

mrDzurb:
## Description
Improves the extraction of custom metadata items by key.
now it supports providing a value to return if the specified key does not exist.

```
assert metadata.get("NoneExistingKey", None) == None
```

### DIFF
--- a/ads/model/model_metadata.py
+++ b/ads/model/model_metadata.py
@@ -8,10 +8,10 @@ import json
 import logging
 import os
 import sys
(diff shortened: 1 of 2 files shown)


darenr (approved):

darenr:
I had to lookup the subtle difference between abstract method and abstract class method.

VipulMascarenhas (approved):

PR #814, opened 2024-05-02
Updates on Job Run API

qiuosier:
This PR adds the following for job run:

- `wait()` method, which will wait until the job run finishes without printing the logs (while `watch()`) prints the logs.
- `exit_code` property, which extracts the exit code from the lifecycle details.
- `force_delete` option to the `delete()` method, which will cancel the job run when it is set to `True`.

This PR also fixes the bug that `force_delete` is available in job.delete() but it is not actually used.

### DIFF
--- a/ads/jobs/builders/infrastructure/dsc_job.py
+++ b/ads/jobs/builders/infrastructure/dsc_job.py
@@ -9,6 +9,7 @@ import datetime
 import inspect
 import logging
 import os
+import re
 import time
 import traceback
 import uuid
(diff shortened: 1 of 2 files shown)


VipulMascarenhas (approved):

ads/jobs/builders/infrastructure/dsc_job.py
```diff
@@ -655,6 +676,22 @@ def _check_and_print_status(self, prev_status) -> str:
             print(f"{timestamp} - {status}")
         return status
 
+    def wait(self, interval: float = SLEEP_INTERVAL):
```
mrDzurb, line 679: Just curious, how this would be different with  ```job.run(wait=True)```?

mrDzurb (approved):

darenr (approved):

mingkang111 (approved):

ads/jobs/builders/infrastructure/dsc_job.py
qiuosier: The extra wait method will allow users to run multiple jobs at the same time and then wait for the job runs to finish. This will be useful when running parallel jobs.

PR #818, opened 2024-05-03
Odsc-56003/fix eval delete

mingkang111:
# Description
This PR aims to fix status for DELETING/DELETED evaluation:

# Changed
* Improved `_get_status` to return correct status under different cases.
* DELETE Eval will return the real status if can be fetched or `DELETING` if real status not available. 
* DELETE Eval remove target eval from listing cache, so that when UI invoke LIST can get correct status. 

# TODO 
* UI show banner/or something else that deletion in progress while waiting for backend response.

# Test
Screenshot 2024-05-03 at 17 01 26

### DIFF
--- a/ads/aqua/evaluation.py
+++ b/ads/aqua/evaluation.py
@@ -1395,6 +1395,11 @@ class AquaEvaluationApp(AquaApp):
             )
 
         try:
+            jobrun_id = model.provenance_metadata.training_id
+        except Exception:
+            raise AquaMissingKeyError(f"Associated Job Run OCID is missing.")
+
+        try:
             job_id = model.custom_metadata_list.get(
                 EvaluationCustomMetadata.EVALUATION_JOB_ID.value
             ).value
@@ -1407,9 +1412,12 @@ class AquaEvaluationApp(AquaApp):
 
         self._delete_job_and_model(job, model)
 
+        self._eval_cache.pop(key=eval_id, default=None)
+        jobrun = utils.query_resource(jobrun_id, return_all=False)
+
         status = dict(
             id=eval_id,
-            lifecycle_state="DELETING",
+            lifecycle_state=jobrun.lifecycle_state if jobrun else "DELETING",
             time_accepted=datetime.now().strftime("%Y-%m-%d %H:%M:%S.%f%z"),
         )
         return status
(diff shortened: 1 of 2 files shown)


ads/aqua/evaluation.py
```diff
@@ -367,6 +367,9 @@ class AquaEvaluationApp(AquaApp):
     _report_cache = TTLCache(maxsize=10, ttl=timedelta(hours=5), timer=datetime.now)
     _metrics_cache = TTLCache(maxsize=10, ttl=timedelta(hours=5), timer=datetime.now)
     _eval_cache = TTLCache(maxsize=200, ttl=timedelta(hours=10), timer=datetime.now)
+    _deletion_cache = TTLCache(
```
VipulMascarenhas, line 370: any reason we've selected 10 mins, or this is arbitrary?

ads/aqua/evaluation.py
```diff
@@ -1407,9 +1411,21 @@ def delete(self, eval_id):
...
+            jobrun_id = model.custom_metadata_list.get(
+                EvaluationCustomMetadata.EVALUATION_JOB_RUN_ID.value
+            ).value
+            jobrun = utils.query_resource(jobrun_id, return_all=False)
+        except Exception:
+            logger.debug("Associated Job Run OCID is missing.")
+            jobrun = None
+
+        self._eval_cache.pop(key=eval_id, default=None)
+        self._deletion_cache.__setitem__(key=eval_id, value="")
```
VipulMascarenhas, line 1424: curious about this: would this cache be set incorrectly if the `_delete_job_and_model` comes across ServiceError while deleting resources?  In that case, will the list API will show status as DELETED for 10 mins (TTL duration) and and then get updated with the correct status?

ads/aqua/evaluation.py
```diff
@@ -1697,30 +1713,33 @@ def _get_status(
...
-            JobRun.LIFECYCLE_STATE_FAILED
-            if self._get_attribute_from_model_metadata(
+        model_status = model.lifecycle_state
+        job_run_status = None
+
+        if jobrun:
+            job_run_status = jobrun.lifecycle_state
+
+        if jobrun is None:
+            if model.identifier in self._deletion_cache.keys():
```
VipulMascarenhas, line 1731: wouldn't this show DELETED status when list API is called even if the job run is still in the process of being deleted?

mingkang111: Hi @VipulMascarenhas , jobrun status is tracked by rqs in LIST. So if job run is still active (i.e. jobrun status != deleted), the jobrun will not be None, and the real status will be used.  For jobrun has been deleted by DELETE Button in AQUA, RQS won't be able to track that jobrun, and the jobrun status should show as DELETED in this case.

VipulMascarenhas (approved):
lgtm 👍

PR #827, opened 2024-05-09
Updated Fine Tuning Parameters

lu-ohai:
### Updated Fine Tuning Parameters

The following fine tuning parameters have been added/updated and provided as options for user:
```
epochs: int
learning_rate: Optional[float] = None
sample_packing: Optional[bool] = "auto"
batch_size: Optional[int] = None
sequence_len: Optional[int] = None
pad_to_sequence_len: Optional[bool] = None
lora_r: Optional[int] = None
lora_alpha: Optional[int] = None
lora_dropout: Optional[float] = None
lora_target_linear: Optional[bool] = None
lora_target_modules: Optional[List] = None
```
### Unit test
Screenshot 2024-05-09 at 5 10 41 PM


VipulMascarenhas (approved):
lgtm 👍

ads/aqua/finetune.py
```diff
@@ -540,7 +553,9 @@ def create(
                 finetuning_source=source.id,
                 finetuning_experiment_id=experiment_model_version_set_id,
             ),
-            parameters=ft_parameters,
+            parameters={
+                key: value for key, value in asdict(ft_parameters).items() if value
```
qiuosier, line 557: same as comment in ` _build_oci_launch_cmd()`, we should check `if value is None`.

ads/aqua/finetune.py
```diff
@@ -591,6 +608,30 @@ def _build_fine_tuning_runtime(
...
+        dataset_path: str,
+        report_path: str,
+        val_set_size: float,
+        parameters: AquaFineTuningParams,
+        finetuning_params: str = None,
+    ) -> str:
+        """Builds the oci launch cmd for fine tuning container runtime."""
+        oci_launch_cmd = f"--training_data {dataset_path} --output_dir {report_path} --val_set_size {val_set_size} "
+        for key, value in asdict(parameters).items():
+            if value:
```
qiuosier, line 622: I think we need to check explicitly if value is None as we may have some parameter as False or 0. For example, if user specify False, this line will drop it from the parameters, then in the FT image it will use auto instead of False since this is not passed into it.

lu-ohai: Oh yes, thanks for pointing this. Updated

qiuosier (approved):

PR #829, opened 2024-05-10
Add support for CLI generation

mayoor:
* Added support for "ads aqua model register" CLI generation
* Fixes to list imported models from user tenancy


ads/aqua/extension/ui_handler.py
```diff
@@ -175,6 +187,46 @@ def is_bucket_versioned(self):
...
+        try:
+            input_data = self.get_json_body()
+        except Exception:
+            raise HTTPError(400, Errors.INVALID_INPUT_DATA_FORMAT)
+
+        if not input_data:
+            raise HTTPError(400, Errors.NO_INPUT_DATA)
+
+        validate_function_parameters(data_class=CLIDetails, input_data=input_data)
+        command_details = CLIDetails(**input_data)
```
mrDzurb, line 217: NIT: This potentially can break the code, if the input data contains some extra fields?  In ```DataClassSerializable``` we support method - ```from_dict``` which has ```ignore_unknown``` attribute.

ads/aqua/extension/ui_handler.py
```diff
@@ -175,6 +187,46 @@ def is_bucket_versioned(self):
...
+
+        interface = AquaCLIHandler.command_interface_map[
```
mrDzurb, line 219: NIT: wouldn't it be better to use ```.get()``` and throw some error or return UNKNOW in case if the command not in map?

mrDzurb (approved):

ads/aqua/model.py
```diff
@@ -835,12 +853,18 @@ def _create_model_catalog_entry(
...
 
         Returns:
             DataScienceModel: Returns Datascience model
         """
         model_info = None
         model = DataScienceModel()
+        if compartment_id:
+            model.with_compartment_id(compartment_id)
+        if project_id:
+            model.with_project_id(project_id)
```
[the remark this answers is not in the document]
mrDzurb: Probably there is a bug somewhere in the implementation, in case of compartment equal None it should be auto extracted form the env vars. Need to check this. For now we can use approach suggested by Vipul.

ads/aqua/extension/ui_handler.py
mayoor: `validate_function_parameters` function just above that line will validate and throw error

ads/aqua/extension/ui_handler.py
mayoor: I wanted the line to fail. The exception will get caught inside the `handle_exception` decorator and return invalid data.

VipulMascarenhas (approved):

ads/aqua/base.py
```diff
@@ -319,3 +320,19 @@ def telemetry(self):
...
             )
         return self._telemetry
+
+
+class CLIBuilderMixin:
+    """
+    CLI builder from API interface. To be used with the DataClass only.
+    """
+
+    def build_cli(self) -> str:
```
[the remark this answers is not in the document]
mayoor: Changed to for comprehension

PR #832, opened 2024-05-10
BUG: Fallback model referencing target series rather than the target column

codeloop:
Fixed the references to the target column.


qiuosier (approved):

ahosler (approved):
LGTM

PR #848, opened 2024-05-15
Aqua refactor: Move all enums under `enums.py` for centralized management

mingkang111:
This PR aims to move all enums under `enums.py` for centralized management

# test
Screenshot 2024-05-15 at 17 30 43

### DIFF
--- a/ads/aqua/app.py
+++ b/ads/aqua/app.py
@@ -12,6 +12,7 @@ from oci.data_science.models import UpdateModelDetails, UpdateModelProvenanceDet
 
 from ads import set_auth
 from ads.aqua import logger
+from ads.aqua.common.enums import Tags
 from ads.aqua.common.errors import AquaRuntimeError, AquaValueError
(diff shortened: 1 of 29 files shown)


VipulMascarenhas (approved):

mrDzurb:
I would recommend maintaining only the enums that are commonly used across multiple packages in ``common/enums.py``. For enums specific to particular functionalities like evaluation or fine-tuning, it would be more appropriate to create dedicated enum files within their respective packages.

mingkang111:
> I would recommend maintaining only the enums that are commonly used across multiple packages in `common/enums.py`. For enums specific to particular functionalities like evaluation or fine-tuning, it would be more appropriate to create dedicated enum files within their respective packages.

Hi @mrDzurb , yes, this is actually the next step. When working on each module, we should consider to move dedicated resources into corresponding module. Here is moving all top-level into `common/enums.py`. Since you mention now, I just moved the one used only in model into model module.

PR #853, opened 2024-05-16
Add source model tag to imported models

VipulMascarenhas:
### Description

This PR covers the following:
1. Adds a `pipeline_tag` check as the API fails where this tag is None.
2. Add AQUA_SERVICE_MODEL_TAG with source model id as value when registered models are verified.

### Unit Tests
```
> python -m pytest -q tests/unitary/with_extras/aqua/test_model.py
rootdir: /Users/vmascarenhas/workspace/git/accelerated-data-science
configfile: pytest.ini
collected 16 items

tests/unitary/with_extras/aqua/test_model.py ................                                                          [100%]

===================================================== 16 passed in 3.31s =====================================================
```

### DIFF
--- a/ads/aqua/extension/model_handler.py
+++ b/ads/aqua/extension/model_handler.py
@@ -179,7 +179,10 @@ class AquaHuggingFaceHandler(AquaAPIhandler):
             )
 
         # Check pipeline_tag, it should be `text-generation`
-        if hf_model_info.pipeline_tag.lower() != ModelTask.TEXT_GENERATION:
+        if (
+            not hf_model_info.pipeline_tag
+            or hf_model_info.pipeline_tag.lower() != ModelTask.TEXT_GENERATION
+        ):
             raise AquaRuntimeError(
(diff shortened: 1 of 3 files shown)


ads/aqua/extension/model_handler.py
```diff
@@ -179,7 +179,10 @@ def post(self, *args, **kwargs):
             )
 
         # Check pipeline_tag, it should be `text-generation`
-        if hf_model_info.pipeline_tag.lower() != ModelTask.TEXT_GENERATION:
+        if (
+            not hf_model_info.pipeline_tag
```
mrDzurb, line 183: If we add this validation, then error message should be adjusted as well?

mrDzurb (approved):

ads/aqua/extension/model_handler.py
VipulMascarenhas: not really, it just prints:
`"Unsupported pipeline tag for the chosen model: 'None'. AQUA currently supports the following tasks only: text-generation. Please select a model with a compatible pipeline tag."`

PR #858, opened 2024-05-20
Copy verified configs for imported models

VipulMascarenhas:
### Description

This PR covers the following:
1. CONSOLE_LINK_RESOURCE_TYPE_MAPPING was missing in utils.py after code refactor.
2. is_valid_ocid now checks if input begins with ocid instead of regex check.
3. If container_index is not accessible, then print debug logs instead of error.
4. `copy_model_config` util function added which copies config json from artifact path of verified models to user bucket.

### Tests

```
> python -m pytest -q tests/unitary/with_extras/aqua/test_model.py
rootdir: /Users/user/workspace/git/accelerated-data-science
configfile: pytest.ini
collected 17 items

tests/unitary/with_extras/aqua/test_model.py .................                                                         [100%]

===================================================== 17 passed in 4.42s =====================================================

```

### DIFF
--- a/ads/aqua/common/utils.py
+++ b/ads/aqua/common/utils.py
@@ -100,23 +100,6 @@ LIFECYCLE_DETAILS_MAPPING = {
     JobRun.LIFECYCLE_STATE_NEEDS_ATTENTION: "Missing jobrun information.",
 }
 
-CONSOLE_LINK_RESOURCE_TYPE_MAPPING = dict(
-    datasciencemodel="models",
-    datasciencemodeldeployment="model-deployments",
-    datasciencemodeldeploymentdev="model-deployments",
-    datasciencemodeldeploymentint="model-deployments",
-    datasciencemodeldeploymentpre="model-deployments",
-    datasciencejob="jobs",
-    datasciencejobrun="job-runs",
-    datasciencejobrundev="job-runs",
-    datasciencejobrunint="job-runs",
-    datasciencejobrunpre="job-runs",
-    datasciencemodelversionset="model-version-sets",
-    datasciencemodelversionsetpre="model-version-sets",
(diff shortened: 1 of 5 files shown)


darenr (approved):

ads/aqua/common/utils.py
```diff
@@ -824,3 +839,51 @@ def get_combined_params(params1: str = None, params2: str = None) -> str:
     ]
 
     return " ".join(combined_params)
+
+
+def copy_model_config(artifact_path: str, os_path: str, auth: dict = None):
```
mrDzurb, line 844: We might need to think of adding ```copy_folder``` function, to copy all the files from one folder to another, if we don't have such function yet. But implementing such function might take some time. In evaluation i added ```copy_files``` method, but it needs to be adjusted for copying entire folder. Also current implementation of ```copy_model_config``` will not take into account the nested folders if we have such in future, right?

mrDzurb (approved):

ads/aqua/common/utils.py
VipulMascarenhas: Yes, agree. I started with that implementation, but there were quite a few boundary conditions we need to account for and need time for testing. I'll add a todo here to revisit this.

ads/aqua/common/utils.py
```diff
@@ -100,6 +100,23 @@ def get_status(evaluation_status: str, job_run_status: str = None):
     JobRun.LIFECYCLE_STATE_NEEDS_ATTENTION: "Missing jobrun information.",
 }
 
+CONSOLE_LINK_RESOURCE_TYPE_MAPPING = dict(
```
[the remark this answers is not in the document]
VipulMascarenhas: yes, I somehow missed that. Moved to constants.

PR #860, opened 2024-05-21
Bug: mlforecast not taking datetime column from the spec

prasankh:
Recently added MLforecast model was not using the datetime column from the spec.


codeloop (approved):
LGTM.

ahosler (approved):
Great catch

darenr (approved):

PR #862, opened 2024-05-21
Accept container family as input for model creation and param validation

VipulMascarenhas:
### Description

This PR includes the changes for:
1. When user registers the model and inference container is not set, then set default as `odsc-tgi-serving`.
2. When deployment is created, accept container_family as input when user deploys unverified models. 
3. When model params are being validated, accept container_family as input when user validates params for unverified models.

### Unit Tests

```
> python -m pytest -q tests/unitary/with_extras/aqua/test_deployment*.py
rootdir: /Users/user/workspace/git/accelerated-data-science
configfile: pytest.ini
collected 29 items

tests/unitary/with_extras/aqua/test_deployment.py ....................                                                 [ 68%]
tests/unitary/with_extras/aqua/test_deployment_handler.py ..s......                                                    [100%]

=============================================== 28 passed, 1 skipped in 7.69s ================================================

```

### DIFF
--- a/ads/aqua/common/enums.py
+++ b/ads/aqua/common/enums.py
@@ -40,6 +40,10 @@ class Tags(str, metaclass=ExtendedEnumMeta):
     AQUA_EVALUATION_MODEL_ID = "evaluation_model_id"
 
 
+class HuggingFaceTags(str, metaclass=ExtendedEnumMeta):
+    TEXT_GENERATION_INFERENCE = "text-generation-inference"
+
+
 class RqsAdditionalDetails(str, metaclass=ExtendedEnumMeta):
     METADATA = "metadata"
     CREATED_BY = "createdBy"
--- a/ads/aqua/common/enums.py
+++ b/ads/aqua/common/enums.py
@@ -40,6 +40,21 @@ class Tags(str, metaclass=ExtendedEnumMeta):
     AQUA_EVALUATION_MODEL_ID = "evaluation_model_id"
 
 
+class InferenceContainerType(str, metaclass=ExtendedEnumMeta):
+    CONTAINER_TYPE_VLLM = "vllm"
+    CONTAINER_TYPE_TGI = "tgi"
+
+
+class InferenceContainerTypeFamily(str, metaclass=ExtendedEnumMeta):
+    AQUA_VLLM_CONTAINER_FAMILY = "odsc-vllm-serving"
+    AQUA_TGI_CONTAINER_FAMILY = "odsc-tgi-serving"
+
+
+class InferenceContainerParamType(str, metaclass=ExtendedEnumMeta):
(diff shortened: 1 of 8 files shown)


ads/aqua/model/model.py
```diff
@@ -621,7 +622,15 @@ def _create_model_catalog_entry(
                 )
             else:
                 logger.warn(
-                    f"Require Inference container information. Model: {model_name} does not have associated inference container defaults. Check docs for more information on how to pass inference container. Proceeding with model registration without the fine-tuning container information. This model will not be available for fine tuning."
+                    f"Proceeding with model registration without the fine-tuning container information. "
+                    f"This model will not be available for fine tuning."
+                )
+
+            if not inference_container:
+                inference_container = InferenceContainerTypeKey.AQUA_TGI_CONTAINER_KEY
```
mayoor, line 630: This should be based on the check if this model has text_generation_inference tag. If not, we should default to vllm

darenr (approved):

ads/aqua/model/model.py
VipulMascarenhas: updated this to get the right image family based on the text_generation_inference tag. I thought all text-generation models will have the tag anyway, so didn't add an additional check here.

ads/aqua/common/enums.py
```diff
@@ -40,6 +40,25 @@ class Tags(str, metaclass=ExtendedEnumMeta):
     AQUA_EVALUATION_MODEL_ID = "evaluation_model_id"
 
 
+class InferenceContainerType(str, metaclass=ExtendedEnumMeta):
```
VipulMascarenhas, line 43: moved this from datasciencemodeldeployment.enums to here to avoid circular imports.

ads/aqua/modeldeployment/deployment.py
```diff
@@ -227,9 +233,17 @@ def create(
                 AQUA_DEPLOYMENT_CONTAINER_METADATA_NAME
             ).value
         except ValueError:
-            raise AquaValueError(
-                f"{AQUA_DEPLOYMENT_CONTAINER_METADATA_NAME} key is not available in the custom metadata field for model {aqua_model.id}"
+            message = (
+                f"{AQUA_DEPLOYMENT_CONTAINER_METADATA_NAME} key is not available in the custom metadata field "
+                f"for model {aqua_model.id}."
             )
+            logger.debug(message)
```
mrDzurb, line 240: NIT: shouldn't we log here the actual error, not the one that we generated?

mrDzurb (approved):

ads/aqua/modeldeployment/deployment.py
VipulMascarenhas: not for this one. We want this to be a user friendly message when this entry is not available in custom metadata.

PR #869, opened 2024-05-27
Feature/multi model artifact handler

YashPandit4u:
Adding a new class to easily handle multi-model artifact creation in ADS.


ads/model/model_description.py
```diff
@@ -0,0 +1,173 @@
...
+    def __init__(self, model_ocid=None):
+
+        self.auth()
+
+        if model_ocid == None: 
+            # if no model given then start from scratch
+            self.modelDescriptionJson = self.empty_json
+        else:
+            # if model given then get that as the starting reference point
+            print("Getting model details from backend")
```
darenr, line 39: all the print statements should be moved to logger.info/debug please

ads/model/model_description.py
```diff
@@ -0,0 +1,173 @@
...
+            destination_file_path = "downloaded_artifact.json"
+            get_model_artifact_content_response = self.data_science_client.get_model_artifact_content(
+                model_id=model_ocid,
+            )
+            try:
+                with open(destination_file_path, "wb") as f:
+                    f.write(get_model_artifact_content_response.data.content)
+                with open(destination_file_path, 'r') as f:
+                    self.modelDescriptionJson = json.load(f)
+            except FileNotFoundError:
```
darenr, line 49: we should just let the exception propagate up - this is too much like Java :)

ads/model/model_description.py
```diff
@@ -0,0 +1,173 @@
...
+
+        def checkIfFileExists(fileName):
+            isExists = False
+            try:
+                headResponse = self.object_storage_client.head_object(namespace, bucket, object_name=fileName)
+                if headResponse.status == 200:
+                    isExists = True
+            except Exception as e:
+                if hasattr(e, 'status') and e.status == 404:
+                    print(f"File not found in bucket: {fileName}")
```
darenr, line 70: logger.error

ads/model/model_description.py
```diff
@@ -0,0 +1,173 @@
...
+            objectStorageList = []
+            has_next_page, opc_next_page = True, None
+            while has_next_page:
+                response = self.object_storage_client.list_object_versions(
+                    namespace_name=namespace,
+                    bucket_name=bucket,
+                    prefix=prefix,
+                    fields="name,size",
+                    page = opc_next_page
+                    )
```
darenr, line 86: this code doesn't look like it was run through the pre-commit ruff/black formatter?

ads/model/model_description.py
```diff
@@ -0,0 +1,173 @@
...
+
+        objects = [{
+                "name": obj.name,
+                "version": obj.version_id,
+                "sizeInBytes": obj.size
+            } for obj in objectStorageList if obj.size > 0]
+        
+        if len(objects) == 0:
+            print("No files to add in the bucket: ", bucket, " with namespace: ", namespace, " and prefix: ", prefix, " file names: ", files)
+            return
```
darenr, line 114: is this an error condition?

ads/model/model_description.py
```diff
@@ -0,0 +1,173 @@
...
+import oci
+import pytz
+import datetime
+import os
+from oci.data_science.models import Metadata
+import ads
+
+class ModelDescription:
+
+    region = ''
```
darenr, line 12: should't this be in the __init__ ctor?

darenr (changes requested):

ads/model/model_description.py
YashPandit4u: Yes thanks, done.

ads/model/model_description.py
YashPandit4u: Done

ads/model/model_description.py
YashPandit4u: Yes, thanks for the comment. Now have raised errors everywhere needed.

Note for this code piece: Now I have changed the code a bit, instead of populating the file with API response and then reading the file contents into a local variable, I have skipped the file part.

ads/model/model_description.py
YashPandit4u: Done

ads/model/model_description.py
YashPandit4u: Yes, it is, now I have raised a "ValueError" in this case.

ads/model/model_description.py
YashPandit4u: Manually ran black.

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+import pytz
+import datetime
+import os
+from oci.data_science.models import Metadata
+import ads
+
+# from ads.common import logger
+import logging
+
+logger = logging.getLogger("ads.model_description")
```
VipulMascarenhas, line 13: can we reuse the logger from `ads.common`?

VipulMascarenhas (commented):
left a few comments, I think we should be reusing most of the implementation from DataScienceModel class instead of recreating it in here. A high level design would look like: 

```
class DataScienceModelCollection(DataScienceModel):

    def __init__():
        super().__init__(**kwargs)
        # no need to add model self.modeldescriptionjson since this attribute is already present in DataScienceModel
    
    def add():
        ...
        
    def remove():
        ...
        
    def show():
        ...     
```

The usage would look like:
```
    >>> model = (DataScienceModelCollection()
    ...    .with_compartment_id()
    ...    .with_project_id()
    ...    .with_display_name()
    ...    .with_description()
    ...    .with_freeform_tags()
    ...    .with_artifact("model1_path", "model2_path")

    >>> model.add(...)
    >>> model.remove(...)
    >>> model.create(model_by_reference=True)
```
any thoughts on this?

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+        a default display name indicating the creation timestamp is used. Defaults to None.
+
+        Returns:
+        - str: The OCID of the created model.
+
+        Note:
+        - The display name defaults to a string indicating the creation timestamp if not provided.
+        """
+        display_name = (
+            "Created by MMS SDK on "
```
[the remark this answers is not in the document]
Abh1navKumar: Yes it should be ADS SDK now.

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+    - save(project_ocid, compartment_ocid, display_name=None): Saves the model to the Model Catalog of OCI Data Science service.
+    """
+
+    empty_json = {
+        "version": "1.0",
+        "type": "modelOSSReferenceDescription",
+        "models": [],
+    }
+
+    def auth(self):
```
Abh1navKumar, line 35: do we need to have this Auth() func ? 
Customer usually set the Auth using ads.setAuth() once which is used in all the ADS SDK interactions.

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+        - None
+
+        Returns:
+        - None
+
+        Note:
+        - The JSON representation of the model description is formatted with an indentation
+        of 4 spaces.
+        """
+        logger.info(json.dumps(self.modelDescriptionJson, indent=4))
```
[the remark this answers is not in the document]
Abh1navKumar: Agree. It should return the modelDescriptionJson to user. They can print it themselves if needed or reuse it.

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+                f"Error writing to file '{file_path}': {e}"
+            )  # Handle the exception accordingly, e.g., log the error, retry writing, etc.
+        except Exception as e:
+            logger.error(
+                f"An unexpected error occurred: {e}"
+            )  # Handle other unexpected exceptions
+        logger.info("Model Artifact stored successfully.")
+        return os.path.abspath(file_path)
+
+    def save(self, project_ocid, compartment_ocid, display_name=None):
```
Abh1navKumar, line 290: Please define types for type checking.

Abh1navKumar (changes requested):

ads/model/model_description.py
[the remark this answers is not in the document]
YashPandit4u: Done, changed to ADS

mrDzurb (commented):
I may have misunderstood what we're trying to achieve in this PR, but wouldn't a user experience like the following be more streamlined?

```python
model_description = ModelDescription().add("oci://bucket_name@namespace/prefix1", "oci://bucket_name@namespace/person_prefix2")
```

Here, `ModelDescription` acts as a builder for the JSON configuration needed.
It still can contain all the introduced methods in this PR: add, remove, build, print...

I would prefer  instead of ```.build to use``` ```.to_json(uri="path/to/the/json")``` where if user don't provide uri then JSON should be returned as a result of the function. Also adding ```.to_dict()``` might be useful as well.
In this case we don't need ```.show()``` method.

We could then enhance the `DatascienceModel` class with:

```python
DatascienceModel().with_model_description(model_description).create()
```

This approach leverages the existing `DatascienceModel` class, which already supports CRUD operations for models, so we wouldn't need to replicate these in the `ModelDescription` class. The purpose of `ModelDescription` would be to simplify the preparation of the correct JSON configuration for users. 

Perhaps we could consider a clearer name for clarity, as "ModelDescription" might be somewhat confusing. How about using "ModelArtifactConfigBuilder" to more accurately reflect its purpose as a tool for constructing model artifact configurations?

Then user experience would be 

```python

model_artifact_config = ModelArtifactConfigBuilder().add().add()

DatascienceModel().with_model_artifact(model_artifact_config).create()
```

cc: @mayoor @VipulMascarenhas

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Note:
+        - This method removes information about objects in the specified bucket from the
+        instance of the ModelDescription.
+        - If a matching model (with the specified namespace, bucket name, and prefix) is found
+        in the model description JSON, it is removed.
+        - If no matching model is found, the method returns without making any changes.
+        """
+
+        def findModelIdx():
```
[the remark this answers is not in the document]
YashPandit4u: Done

ads/model/model_description.py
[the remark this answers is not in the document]
YashPandit4u: Added return too

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+                content = get_model_artifact_content_response.data.content
+                self.modelDescriptionJson = json.loads(content)
+            except json.JSONDecodeError as e:
+                logger.error(f"Error decoding JSON: {e}")
+                raise e
+            except Exception as e:
+                logger.error(f"An unexpected error occurred: {e}")
+                raise e
+
+    def add(self, namespace, bucket, prefix=None, files=None):
```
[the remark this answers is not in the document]
YashPandit4u: Added type hints.

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+        Returns:
+        - str: The OCID of the created model.
+
+        Note:
+        - The display name defaults to a string indicating the creation timestamp if not provided.
+        """
+        display_name = (
+            "Created by MMS SDK on "
+            + datetime.datetime.now(pytz.utc).strftime("%Y-%m-%d %H:%M:%S %Z")
+            if (display_name == None)
```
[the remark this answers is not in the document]
YashPandit4u: The suggested statement will trigger even when we have blank string.

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+            else display_name
+        )
+        customMetadataList = [Metadata(key="modelDescription", value="true")]
+        model_details = oci.data_science.models.CreateModelDetails(
```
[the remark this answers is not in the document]
YashPandit4u: Did this now by extending this class from DataScienceModel

ads/model/model_description.py
```diff
@@ -0,0 +1,328 @@
...
+        Note:
+        - The display name defaults to a string indicating the creation timestamp if not provided.
+        """
+        display_name = (
+            "Created by MMS SDK on "
+            + datetime.datetime.now(pytz.utc).strftime("%Y-%m-%d %H:%M:%S %Z")
+            if (display_name == None)
+            else display_name
+        )
+        customMetadataList = [Metadata(key="modelDescription", value="true")]
```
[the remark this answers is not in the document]
YashPandit4u: As we are now using create from the parent class this code piece is removed.

ads/model/model_description.py
YashPandit4u: This internal _auth() is used in __init__() ans is used to authenticate object storage client based on details given by the ads auth.

ads/model/model_description.py
YashPandit4u: Thanks, did this for all methods.

YashPandit4u:
Hi  @VipulMascarenhas ,

I have now extended from the class already present in ads and the usage of the utility looks like this now:

`import sys
import ads.common
from ads.model.model_collection import DataScienceModelCollection
import ads
import json

# Auth
ads.set_auth(auth="security_token")


mdes = DataScienceModelCollection() \
    .with_ref_model_id(model_ocid="ocid1.datasciencemodel.oc1.iad.amaaaaaav66vvnialtuoq5y4ltdz26bxhlow4uh5w544ullowe3jl4vwuayq") \
    .with_display_name("New Design 8") \
    .with_compartment_id("ocid1.tenancy.oc1..aaaaaaaahzy3x4boh7ipxyft2rowu2xeglvanlfewudbnueugsieyuojkldq") \
    .with_project_id("ocid1.datascienceproject.oc1.iad.amaaaaaav66vvniaqsyu2nufljutkn4rzth2nz4q3zqslirct7eayl5ojpma")

mdes.add(namespace="ociodscdev", bucket="unzip-multi-model", prefix="model-linear-1")
mdes.remove(namespace="ociodscdev", bucket="unzip-multi-model", prefix="model-linear-1")

mdes.create()

sys.exit(0)

`

Please take a look when free and give your comments.

Thanks and Regards
Yash Pandit

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+                f"Error writing to file '{file_path}': {e}"
+            )  # Handle the exception accordingly, e.g., log the error, retry writing, etc.
+        except Exception as e:
+            logger.error(
+                f"An unexpected error occurred: {e}"
+            )  # Handle other unexpected exceptions
+        logger.info("Model Artifact stored successfully.")
+        return os.path.abspath(file_path)
+
+    def show(self):
```
Abh1navKumar, line 273: we need to add return type as str also.

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+        Note:
+        - If `files` is not provided, it retrieves information about all objects in the bucket.
+        If `files` is provided, it only retrieves information about objects with matching file names.
+        - If no objects are found to add to the model description, a ValueError is raised.
+        """
+
+        # Remove if the model already exists
+        self.remove(namespace, bucket, prefix)
+
+        def check_if_file_exists(fileName):
```
Abh1navKumar, line 108: Wrong indentation.

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+from typing import List, Optional
+import logging
+
+logger = logging.getLogger("ads.model_description")
+logger.setLevel(logging.INFO)
+
+
+class DataScienceModelCollection(DataScienceModel):
+
+    def _auth(self):
```
mrDzurb, line 16: I don't think that we need this method here. The all operations with the service take place within the ```OCIDataScienceModel``` which is already has such method. Please take a look how the ```DatascienceModel.create``` method works.

```python
# Create model in the model catalog
logger.info("Saving model to the Model Catalog.")
self.dsc_model = self._to_oci_dsc_model(**kwargs).create()
```

However if we still need to create a client, then it would be better to use ```OCIClientFactory```. 
Something like:

```
def create_os_client(self):
        return oci_client.OCIClientFactory(**self.auth).object_storage
```

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
+import json
+import ads.common
+import oci
+import os
+import ads
+from ads.model.datascience_model import DataScienceModel
+from typing import List, Optional
+import logging
+
+logger = logging.getLogger("ads.model_description")
```
mrDzurb, line 10: Wouldn't this be better - ```logger = logging.getLogger(__name__)```?

ads/model/model_collection.py
mrDzurb, line 108: Please double check if the existing utils methods are already implemented such functionality.

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+        self = self.with_artifact(uri=tmp_file_path)
+        created_model = super().create()
+        try:
+            os.remove(tmp_file_path)
+        except Exception as e:
+            logger.error(f"Error occurred while cleaning file: {e}")
+            raise e
+        return created_model.id
+
+    def build(self) -> str:
```
VipulMascarenhas, line 244: we had reservations around calling this `build` since conventionally build function should return self/object. User can just do `json.dump(model.model_file_description)` but if we really need this functionality, it would be good to have different function name.

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+                f"Error writing to file '{file_path}': {e}"
+            )  # Handle the exception accordingly, e.g., log the error, retry writing, etc.
+        except Exception as e:
+            logger.error(
+                f"An unexpected error occurred: {e}"
+            )  # Handle other unexpected exceptions
+        logger.info("Model Artifact stored successfully.")
+        return os.path.abspath(file_path)
+
+    def show(self) -> str:
```
VipulMascarenhas, line 273: this is covered by `model.model_file_description` property.

VipulMascarenhas (commented):
@YashPandit4u suggested a few changes, just needs to be re-organized to make it consistent with ads usage.

ads/model/model_collection.py
YashPandit4u: Yes removing this.

ads/model/model_collection.py
YashPandit4u: Okay

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+            return -1
+
+        modelSearchIdx = findModelIdx()
+        if modelSearchIdx == -1:
+            return
+        else:
+            # model found case
+            self.model_file_description["models"].pop(modelSearchIdx)
+
+    def create(self):
```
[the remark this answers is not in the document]
YashPandit4u: Yes removed this.

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+        Note:
+        - This method removes information about objects in the specified bucket from the
+        instance of the ModelDescription.
+        - If a matching model (with the specified namespace, bucket name, and prefix) is found
+        in the model description JSON, it is removed.
+        - If no matching model is found, the method returns without making any changes.
+        """
+
+        def findModelIdx():
+            for idx, model in enumerate(self.model_file_description["models"]):
```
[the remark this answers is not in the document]
YashPandit4u: Thanks, added this condition.

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+            {
+                "namespace": namespace,
+                "bucketName": bucket,
+                "prefix": prefix,
+                "objects": objects,
+            }
+        )
+        self.set_spec(self.CONST_MODEL_FILE_DESCRIPTION, tmp_model_file_description)
+
+    def remove(self, namespace: str, bucket: str, prefix: Optional[str] = None):
```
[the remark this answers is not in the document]
YashPandit4u: Yes done.

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+            self.set_spec(self.CONST_MODEL_FILE_DESCRIPTION, json.loads(content))
+        except json.JSONDecodeError as e:
+            logger.error(f"Error decoding JSON: {e}")
+            raise e
+        except Exception as e:
+            logger.error(f"An unexpected error occurred: {e}")
+            raise e
+        return self
+
+    def add(
```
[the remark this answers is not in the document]
YashPandit4u: The user won't have uri handy, hence we have parameters like namespace, and bucket to make it easy. 
Will it be possible to keep these parameters? The function signature would look like:

`def add_artifact(
        self,
        namespace: str,
        bucket: str,
        prefix: Optional[str] = None,
        files: Optional[List[str]] = None,
    )`

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+import ads
+from ads.model.datascience_model import DataScienceModel
+from typing import List, Optional
+import logging
+
+logger = logging.getLogger("ads.model_description")
+logger.setLevel(logging.INFO)
+
+
+class DataScienceModelCollection(DataScienceModel):
```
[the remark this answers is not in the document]
YashPandit4u: Yes, now I have added the add_artifact and remove_artifact methods in the DataScienceModel class itself.

ads/model/model_collection.py
```diff
@@ -0,0 +1,288 @@
...
+        tmp_model_file_description = self.model_file_description
+        tmp_model_file_description["models"].append(
+            {
+                "namespace": namespace,
+                "bucketName": bucket,
+                "prefix": prefix,
+                "objects": objects,
+            }
+        )
+        self.set_spec(self.CONST_MODEL_FILE_DESCRIPTION, tmp_model_file_description)
```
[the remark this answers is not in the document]
YashPandit4u: Thanks for this suggestion.
Will it be okay if we do this in the next PR due to customer deadlines?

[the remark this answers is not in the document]
VipulMascarenhas: sure, let's mark a todo somewhere to make sure of this.

ads/model/model_collection.py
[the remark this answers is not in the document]
VipulMascarenhas: isn't the uri same as `f"oci://{bucketName}@{namespace}/{prefix}"`, all of which are added as input to this function. My concern is that this input makes it incosistent with all other `DataScienceModel` or `GenericModel` operations in ads and we wouldn't be able to change it later on. 
@mrDzurb what do you think?

VipulMascarenhas:
@YashPandit4u thanks for making all the changes here, looks good overall. One concern I have is around the path inputs, it would be good to have the oci path like I mentioned in the comments above as user can construct the path easily with the params (bucket, namespace and prefix).  Also, could you please add a couple unit tests in `test_datascience_model.py` for this change? 

cc: @mrDzurb

ads/model/datascience_model.py
```diff
@@ -1466,3 +1468,159 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
             bucket_uri.append(uri)
 
         return bucket_uri[0] if len(bucket_uri) == 1 else bucket_uri, artifact_size
+    
+    def add_artifact(
+        self,
+        namespace: str,
```
mrDzurb, line 1474: For consistency it would be better to accept OCIFS "oci://" path.

ads/model/datascience_model.py
```diff
@@ -1466,3 +1468,159 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
...
+        bucket: str,
+        prefix: Optional[str] = None,
+        files: Optional[List[str]] = None,
+    ):
+        """
+        Adds information about objects in a specified bucket to the model description JSON.
+
+        Parameters:
```
mrDzurb, line 1482: NIT: for the pydocs we have a bit different format

```
Parameters
------------
namespace (str): The namespace of the object storage.

Returns
--------
None

Raises
------
 ValueError: If no files are found to add to the model description.

```

ads/model/datascience_model.py
```diff
@@ -1466,3 +1468,159 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
...
+        if self.model_file_description == None:
+            self.empty_json = {
+                "version": "1.0",
+                "type": "modelOSSReferenceDescription",
+                "models": [],
+            }
+            self.set_spec(self.CONST_MODEL_FILE_DESCRIPTION, self.empty_json)
+        
+        # Get object storage client
+        authData = default_signer()
```
mrDzurb, line 1509: Technically the auth object is already constructed and accessible by 
```
self.dsc_model.auth
```

YashPandit4u: Thanks, using this now.

ads/model/datascience_model.py
YashPandit4u: Changed docs to this format.

ads/model/datascience_model.py
YashPandit4u: Have changed both functions add_artifact and remove_artifact with this uri type input.

VipulMascarenhas (dismissed):
@YashPandit4u thanks for the update. There are a few improvements to make for the next update: 
- check_if_file_exists and list_obj_versions_unpaginated should reuse existing methods.
- _extract_oci_uri_components wasn't required, as mentioned in previous comments we could just use `ObjectStorageDetails.from_path(uri)` to get bucket name, namespace and prefix. If we want to add validation ObjectStorageDetails has a few static methods like `is_valid_uri` to do this.

YashPandit4u:
Hi @VipulMascarenhas , thanks for telling the improvement items. 
* I have now removed the "_extract_oci_uri_components" method and used the already present ObjectStorageDetails.from_path(uri) to get the namespace, prefix, and filepath. Thanks for the suggestion.

VipulMascarenhas (dismissed):
lgtm 👍

YashPandit4u:
Hi @VipulMascarenhas ,
Based on the management decision, I have reverted the method signatures to the ones suggested by the customer.

YashPandit4u:
Hi @VipulMascarenhas ,
 
 I have now added options for both uri and (namespace, bucket, prefix) in add_artifact, remove_artifact.

ads/model/datascience_model.py
```diff
@@ -1466,3 +1467,226 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
             bucket_uri.append(uri)
 
         return bucket_uri[0] if len(bucket_uri) == 1 else bucket_uri, artifact_size
+
+    def add_artifact(
```
mrDzurb, line 1471: NIT: Could you please add an examples section for the add and delete artifacts feature?

ads/model/datascience_model.py
```diff
@@ -1466,3 +1467,226 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
...
+        if uri:
+            object_storage_details = ObjectStorageDetails.from_path(uri)
+            bucket = object_storage_details.bucket
+            namespace = object_storage_details.namespace
+            prefix = (
+                None
+                if object_storage_details.filepath == ""
+                else object_storage_details.filepath
+            )
+        if (not namespace) or (not bucket):
```
mrDzurb, line 1527: NIT: ```if not (namespace and bucket):```

ads/model/datascience_model.py
```diff
@@ -1466,3 +1467,226 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
...
+            raise ValueError("Both 'namespace' and 'bucket' must be provided.")
```
mrDzurb, line 1528: NIT: Wouldn't it be better to say something like - ```Artifacts cannot be added. Both 'namespace' and 'bucket' parameters must be provided.```. otherwise it would be hard for user to understand where they need to provide the namespace and bucket.

ads/model/datascience_model.py
```diff
@@ -1466,3 +1467,226 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
...
+            - If no files are found to add to the model description.
+
+        Note
+        ----
+        - If `files` is not provided, it retrieves information about all objects in the bucket.
+        - If `files` is provided, it only retrieves information about objects with matching file names.
+        - If no objects are found to add to the model description, a ValueError is raised.
+        """
+
+        if uri and (namespace or bucket):
```
mrDzurb, line 1514: Maybe we can do something like this?

```
if uri and (namespace or bucket):
    raise ValueError("Artifacts cannot be added. Please provide either a 'uri' alone, or 'namespace' and 'bucket' together, but not 'uri' with 'namespace' or 'bucket'.")

elif not (uri or (namespace and bucket)):
    raise ValueError("Artifacts cannot be added. You must provide either a 'uri' or both 'namespace' and 'bucket'.")

```

ads/model/datascience_model.py
```diff
@@ -1466,3 +1467,226 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
...
+
+        # Get object storage client
+        self.object_storage_client = oc.OCIClientFactory(
+            **(self.dsc_model.auth)
+        ).object_storage
+
+        # Remove if the model already exists
+        self.remove_artifact(namespace=namespace, bucket=bucket, prefix=prefix)
+
+        def check_if_file_exists(fileName):
```
mrDzurb, line 1552: NIT: I would not recommend to create nested functions, it would be hard to write unit tests for such methods. I would rather recommend to move such methods into utils or outside the class.

ads/model/datascience_model.py
```diff
@@ -1466,3 +1467,226 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
...
+                    )
+
+        objects = [
+            {"name": obj.name, "version": obj.version_id, "sizeInBytes": obj.size}
+            for obj in objectStorageList
+            if obj.size > 0
+        ]
+
+        if len(objects) == 0:
+            error_message = (
```
mrDzurb, line 1607: NIT: 

```
error_message = (
        f"No files found to add in the bucket '{bucket}' within the namespace '{namespace}' "
        f"and prefix '{prefix}'. Expected file names: {files}"
    )
```

ads/model/datascience_model.py
```diff
@@ -1466,3 +1467,226 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
...
+            {
+                "namespace": namespace,
+                "bucketName": bucket,
+                "prefix": "" if not prefix else prefix,
+                "objects": objects,
+            }
+        )
+        self.set_spec(self.CONST_MODEL_FILE_DESCRIPTION, tmp_model_file_description)
+
+    def remove_artifact(
```
mrDzurb, line 1625: Please add the examples section.

ads/model/datascience_model.py
```diff
@@ -1466,3 +1467,226 @@ def _download_file_description_artifact(self) -> Tuple[Union[str, List[str]], in
...
+
+        Raises
+        ------
+        ValueError
+            - If both 'uri' and ('namespace' and 'bucket') are provided.
+            - If neither 'uri' nor both 'namespace' and 'bucket' are provided.
+            - If the model description JSON is None.
+        """
+
+        if uri and (namespace or bucket):
```
mrDzurb, line 1658: See the comments for the add_artifact method.

mrDzurb (approved):

VipulMascarenhas (approved):

PR #880, opened 2024-06-11
Update version to 2.11.11 with bug fixes

VipulMascarenhas:
### Description

This PR has code changes for the following:
- Update version and release notes.
- Integration test bug fix: 
     - Remove validation for byom FT model deployment and fine-tuning.
     - Update metadata to replace artifact location if already present.
     - Pop base custom model flag from FT model to ensure it is not listed when model_type is set as BASE.

### DIFF
--- a/ads/aqua/common/utils.py
+++ b/ads/aqua/common/utils.py
@@ -669,7 +669,7 @@ def get_model_by_reference_paths(model_file_description: dict):
     models = model_file_description["models"]
 
     if models:
-        if len(models) == 1:
+        if len(models) > 0:
             base_model_artifact = models[0]
             base_model_path = f"oci://{base_model_artifact['bucketName']}@{base_model_artifact['namespace']}/{base_model_artifact['prefix']}".rstrip(
                 "/"
--- a/ads/aqua/common/utils.py
+++ b/ads/aqua/common/utils.py
@@ -668,17 +668,25 @@ def get_model_by_reference_paths(model_file_description: dict):
     fine_tune_output_path = UNKNOWN
     models = model_file_description["models"]
 
-    if models:
-        if len(models) > 0:
-            base_model_artifact = models[0]
-            base_model_path = f"oci://{base_model_artifact['bucketName']}@{base_model_artifact['namespace']}/{base_model_artifact['prefix']}".rstrip(
-                "/"
-            )
-        if len(models) == 2:
-            ft_model_artifact = models[1]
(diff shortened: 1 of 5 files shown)


darenr (approved):

mrDzurb (approved):

mayoor (dismissed):

ads/aqua/common/utils.py
```diff
@@ -668,23 +668,18 @@ def get_model_by_reference_paths(model_file_description: dict):
...
-            model["prefix"],
-        )
-        bucket_uri = f"oci://{bucket_name}@{namespace}/{prefix}".rstrip("/")
-        if bucket_name == AQUA_SERVICE_MODELS_BUCKET:
-            base_model_path = bucket_uri
-        else:
-            fine_tune_output_path = bucket_uri
+    if models:
+        if len(models) > 0:
+            base_model_artifact = models[0]
```
mayoor, line 673: Can you put a comment here why index the first model?

ads/aqua/common/utils.py
```diff
@@ -668,23 +668,18 @@ def get_model_by_reference_paths(model_file_description: dict):
...
-            model["namespace"],
-            model["bucketName"],
-            model["prefix"],
-        )
-        bucket_uri = f"oci://{bucket_name}@{namespace}/{prefix}".rstrip("/")
-        if bucket_name == AQUA_SERVICE_MODELS_BUCKET:
-            base_model_path = bucket_uri
-        else:
-            fine_tune_output_path = bucket_uri
+    if models:
```
mrDzurb, line 671: I guess we can simplify a logic a bit here, something like: 

```
if not models:
     rise AquaValueError
 if len(models)>0:
      base_model_artifact = models[0]
 if len(models)>1:
        ft_model_artifact = models[1]
```

mrDzurb (approved):

ads/aqua/common/utils.py
VipulMascarenhas: yes, will simplify this. I have trouble sleeping when we raise AquaValueError anywhere now.

VipulMascarenhas: updated.

ads/aqua/common/utils.py
VipulMascarenhas: updated.

mrDzurb (approved):

mayoor (approved):

PR #885, opened 2024-06-19
Added ruff for pre-commit check

liudmylaru:
### Description

Ruff can be used to replace Flake8 (plus dozens of plugins), Black, isort, pydocstyle, and more, all while executing tens or hundreds of times faster than any individual tool.

Added ruff to run mainly default settings with little extra and exclude. While developing our code we will add/exclude more. Please check on rules and suggest what can be added to default (section 'select' in pyproject.toml) and what can be excluded (section 'ignore' in pyproject.toml)- https://docs.astral.sh/ruff/rules/.

### What was done

- added repo to .pre-commit-config.yaml
- added rules to pyproject.toml
- removed black, because ruff has drop-in parity with it

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,10 +4,12 @@ repos:
       rev: v0.4.9
       hooks:
           - id: ruff
+            types_or: [ python, pyi, jupyter ]
             args: [ --fix ]
             files: ^ads
             exclude: ^docs/
           - id: ruff-format
+            types_or: [ python, pyi, jupyter ]
             exclude: ^docs/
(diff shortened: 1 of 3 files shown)


mrDzurb (dismissed):

darenr (approved):
With the defaults do you find that ads builds fine? I found that I had to use the ruff conf that I shared to exclude various things otherwise it would flag a number or issues. If you don't find that then go ahead. Thanks for finishing this.

VipulMascarenhas (approved):

mrDzurb (approved):

PR #887, opened 2024-06-25
Python 3.11 support

liudmylaru:
### Description

ADS to support Python3.11. Jira: https://jira.oci.oraclecorp.com/browse/ODSC-54710

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,7 @@ onnx = [
   "lightgbm",
   "onnx>=1.12.0",
   "onnxmltools>=1.10.0",
-  "onnxruntime>=1.10.0,<1.16.0", # avoid v1.16 https://github.com/microsoft/onnxruntime/issues/17631, revealed by unit tests
+  "onnxruntime==1.15.1,!=1.16.0",  # use 1.15.1 (same as Oracle Database https://docs.oracle.com/en/database/oracle/machine-learning/oml4sql/23/mlsql/onnx.html); avoid v1.16 https://github.com/microsoft/onnxruntime/issues/17631, revealed by unit tests
   "oracle_ads[viz]",
   "protobuf",
   "skl2onnx>=1.10.4",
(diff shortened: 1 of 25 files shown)


pyproject.toml
```diff
@@ -182,27 +199,25 @@ aqua = ["jupyter_server"]
...
-  "dask==2023.10.1; python_version>='3.9'", # oci-cli depends on click==8.0.4, dask>2023.10.1 depends on "click>=8.1"
   "arff",
   "category_encoders==2.6.3",  # set version to avoid backtracking
   "cohere==4.53",  # set version to avoid backtracking
-  "dask==2023.10.1; python_version>='3.9'", # oci-cli depends on click==8.0.4, dask>2023.10.1 depends on "click>=8.1"
-  "dask==2023.5.0; python_version=='3.8'",
   "faiss-cpu",
   "fastparquet==2024.2.0",  # set version to avoid backtracking
   "imbalanced-learn",
+  "kubernetes",
```
darenr, line 208: Do we absolutely need to pull this in?

pyproject.toml
```diff
@@ -89,18 +90,24 @@ data = [
...
-huggingface = ["transformers"]
 notebook = ["ipython>=7.23.1, <8.0", "ipywidgets~=7.6.3"]
 onnx = [
-  "lightgbm<4.0.0",            # relax when the official releases of skl2onnx (v1.16.0) and onnxmltools (1.11.3), https://github.com/sdpython/mlprodict/issues/488
-  "onnx>=1.12.0",
+  "lightgbm",
+  "onnx>=1.12.0,<=1.14.0",  # v 1.14.0 set base on onnxrutime version and onnx opset support - https://onnxruntime.ai/docs/reference/compatibility.html#onnx-opset-support
   "onnxmltools>=1.10.0",
-  "onnxruntime>=1.10.0,<1.16", # v1.16 introduced issues https://github.com/microsoft/onnxruntime/issues/17631, revealed by unit tests
+  "onnxruntime~=1.15.1,!=1.16.0",  # v1.15.1 used in Oracle Database https://docs.oracle.com/en/database/oracle/machine-learning/oml4sql/23/mlsql/onnx.html); avoid v1.16 https://github.com/microsoft/onnxruntime/issues/17631, revealed by unit tests
```
darenr, line 108: I've been told that 23ai uses 1.17 of the `onnnxruntime`, but it's possible I'm remembering it wrong.  Will this mean we are compatible with 23ai because that would be great if we are.

darenr (dismissed):
Couple of inline comments that I noticed. This is super fast work, thanks.

pyproject.toml
liudmylaru: I moved it from test-requirements.txt. I checked, there is no need to have it in requirements.txt and here in [testsuite], because this is a dependency for a [feature-store-marketplace]. I will remove it from here, you are right, this should not be a dependency for ads test suite, but for feature-store feature.

pyproject.toml
liudmylaru: set onnxruntime  version ~=1.17.0

mrDzurb (dismissed):

mrDzurb (approved):
When we merge this PR, can we smash the commits?

VipulMascarenhas (approved):
lgtm 👍

PR #892, opened 2024-06-30
Websocket changes for AQUA (1.0.3)

kumar-shivam-ranjan:


.pre-commit-config.yaml
```diff
@@ -1,16 +1,4 @@
 repos:
-    # ruff
```
darenr, line 2: ADS has switched from black to ruff

darenr (changes requested):

.pre-commit-config.yaml
kumar-shivam-ranjan: yes , some of the files have outdated changes.  will resolve these.

PR #897, opened 2024-07-03
Optimised report loading for anomaly operator

govarsha:
**Made following changes to optimise report loading for anomaly detector:**
1. If non-anomolous data points > 1000, it downsamples them to 1000. I chose the threshold as 1000 because visually a plot with more than 1000 datapoints is too crowded.
2. All anomolous data points are included
3. In the report, we show the whole dataset which leads to very large file size. Hence, we are showing 1000 data points at max here as well.
4. The optimisation can be turned off by passing optimize_report = False in spec

**Results:**
I have used [cpu_utilization_asg_misconfiguration.csv](https://github.com/numenta/NAB/blob/master/data/realKnownCause/cpu_utilization_asg_misconfiguration.csv) from NAB which has 18051 data points.

**Report without optimisation:**
Loading time - 15 seconds
Size - 6.2MB
Screenshot 2024-07-03 at 12 27 33 PM
Screenshot 2024-07-03 at 12 27 42 PM

**Report with optimisation:**
Loading time - 2/3 seconds
Size - 411KB
[
Screenshot 2024-07-03 at 12 27 04 PM
](url)
Screenshot 2024-07-03 at 12 27 18 PM


anomaly_out/anomali_2_test.csv
```diff
@@ -0,0 +1,8 @@
+PPG_Code,last_day_of_week,anomaly
```
ahosler, line 1: Is this file used?

ads/opctl/operator/lowcode/anomaly/schema.yaml
```diff
@@ -377,4 +377,9 @@ spec:
       type: dict
       required: false
 
+    optimize_report:
```
ahosler, line 380: IMO "optimize_report" will be too vague. 
In v2, we should do "report_options" and have "subsample_data" as one option (along with "theme", "name", etc).

In that spirit, maybe we do "subsample_report_data"? Otherwise looks good (default to true, not required)

anomaly_out/anomali_2_test.csv
govarsha: no, came by mistake. removed it.

ahosler (dismissed):
This looks great, thanks!

VipulMascarenhas (approved):

ads/opctl/operator/lowcode/anomaly/model/base_model.py
```diff
@@ -79,7 +79,7 @@ def generate_report(self):
                 anomaly_output, test_data, elapsed_time
             )
         table_blocks = [
-            rc.DataTable(df, label=col, index=True)
+            rc.DataTable(df.head(1000) if self.spec.subsample_report_data and len(df) > 1000 else df, label=col, index=True)
```
mrDzurb, line 82: NIT: Would be nice to move 1000 to constants.

ads/opctl/operator/lowcode/anomaly/model/base_model.py
```diff
@@ -94,20 +94,35 @@ def generate_report(self):
             anomaly_col = anomaly_output.get_anomalies_by_cat(category=target)[
                 OutputColumns.ANOMALY_COL
             ]
+            anomaly_indices = [i for i, index in enumerate(anomaly_col) if index == 1]
```
mrDzurb, line 97: NIT: if wea are not modifying these lists, they can be tuples instead, they are working faster.

ads/opctl/operator/lowcode/anomaly/model/base_model.py
```diff
@@ -94,20 +94,35 @@ def generate_report(self):
...
+            downsampled_time_col = time_col
+            selected_indices = list(range(len(time_col)))
+            if self.spec.subsample_report_data:
+                non_anomaly_indices = [i for i in range(len(time_col)) if i not in anomaly_indices]
+                # Downsample non-anomalous data if it exceeds the threshold (1000)
+                if len(non_anomaly_indices) > 1000:
+                    downsampled_non_anomaly_indices = non_anomaly_indices[::len(non_anomaly_indices)//1000]
+                    selected_indices = sorted(anomaly_indices + downsampled_non_anomaly_indices)
```
mrDzurb, line 105: NIT: `sorted()` creates a copy of the list, do we really need this. The `.sort()` would work faster.

mrDzurb (approved):

ads/opctl/operator/lowcode/anomaly/model/base_model.py
govarsha: Moved it to constants

ads/opctl/operator/lowcode/anomaly/model/base_model.py
govarsha: Yeah, changed it to sort.

ads/opctl/operator/lowcode/anomaly/model/base_model.py
govarsha: We have selected_indices which is created from anomaly_indices & downsampled_non_anomaly_indices lists by sort, so if we use tuples we will have to change it back to list to sort. Hence, I am just using list.

ahosler (approved):

PR #899, opened 2024-07-03
Updated Job BYOC API

lu-ohai:
### Updated Job BYOC API

- Applied payload `OcirContainerJobEnvironmentConfigurationDetails` instead of environment variables to pass to oci backend
- Added APIs for class `ContainerRuntime`:
-- `with_image_digest`
-- `with_image_signature_id`

## Notebook
Screenshot 2024-07-03 at 2 20 36 PM
Screenshot 2024-07-03 at 2 22 06 PM
Screenshot 2024-07-03 at 2 23 39 PM
Screenshot 2024-07-03 at 2 24 43 PM
Screenshot 2024-07-03 at 2 25 29 PM


mrDzurb:
Do you know if it back compatible? Will the existing  YAML work with the new spec?

ads/jobs/builders/infrastructure/dsc_job_runtime.py
```diff
@@ -157,6 +157,21 @@ def _translate_env(self, runtime: Runtime) -> dict:
         """
         return runtime.envs
 
+    def _translate_env_config(self, runtime: Runtime) -> dict:
```
qiuosier, line 160: Looks like this method is only used by one sub-class. Maybe we don't need this method here as it is not doing anything?

qiuosier (dismissed):

lu-ohai:
> Do you know if it back compatible? Will the existing YAML work with the new spec?

Yes. There is no change for current user's experience except I added several apis to configure more properties. The logic changed is that we no longer send ENV to backend under the hood but used `OcirContainerJobEnvironmentConfigurationDetails` as payload instead.

ads/jobs/builders/infrastructure/dsc_job_runtime.py
lu-ohai: Removed it in the base class

qiuosier (approved):

mrDzurb (approved):

PR #903, opened 2024-07-10
[ODSC-59694] Register model changes for 1.0.3

kumar-shivam-ranjan:


darenr (approved):

ads/aqua/model/model.py
```diff
@@ -581,9 +580,14 @@ def _create_model_catalog_entry(
...
                 **verified_model.freeform_tags,
                 Tags.AQUA_SERVICE_MODEL_TAG: verified_model.id,
+                Tags.PLATFORM: "cpu" if is_gguf_model else "gpu",
             }
             if verified_model
-            else {Tags.AQUA_TAG: "active", Tags.BASE_MODEL_CUSTOM: "true"}
+            else {
+                Tags.AQUA_TAG: "active",
+                Tags.BASE_MODEL_CUSTOM: "true",
+                Tags.PLATFORM: "cpu" if is_gguf_model else "gpu",
```
VipulMascarenhas, line 589: nit: 
this can be moved outside if-else: `tags.update({Tags.PLATFORM: "cpu" if is_gguf_model else "gpu"})`
Also, "cpu" and "gpu" can be added as constants in `ads.aqua.constants`.

ads/aqua/model/model.py
```diff
@@ -693,24 +697,25 @@ def register(
...
-        except Exception as ex:
-            logger.error(
-                f"Exception occurred while loading config file from {import_model_details.os_path}"
-                f"Exception message: {ex}"
-            )
-            raise AquaRuntimeError(
-                f"The model path {import_model_details.os_path} does not contain the file config.json. "
-                f"Please check if the path is correct or the model artifacts are available at this location."
-            )
+        is_gguf_model = import_model_details.inference_container == "odsc-llama-cpp"
```
VipulMascarenhas, line 703: can we rename this to "odsc-llama-cpp-serving" to be consistent with other names? Also, can you please add this inside `ads.common.enums.InferenceContainerTypeFamily`?

VipulMascarenhas (commented):
changes look good, left a couple minor comments.

ads/aqua/model/model.py
kumar-shivam-ranjan: Sure , will update , thanks

ads/aqua/model/model.py
kumar-shivam-ranjan: Sure , will update , thanks

ads/aqua/model/model.py
```diff
@@ -786,6 +792,7 @@ def register(
         except:
             finetuning_container = None
 
+        platform = "cpu" if is_gguf_model else "gpu"
```
harsh97, line 795: should we have this as amd-cpu and nvidia GPU?

VipulMascarenhas: ok, should be arm-cpu and nvidia-gpu in that case to be consistent with the UI filters.

PR #913, opened 2024-07-25
Add HF integration APIs

VipulMascarenhas:
### Description

This PR add the following APIs:
1. Get HF Model Details

#### Request
```
POST aqua/model/hf/search
 
BODY
{
  "model_id": "codellama/CodeLlama-7b-Instruct-hf"
}
```

#### Response
```
{
    "model_info": {
        "id": "codellama/CodeLlama-7b-Instruct-hf",
        "author": "codellama",
        "sha": "22cb240e0292b0b5ab4c17ccd97aa3a2f799cbed",
      ....
      ....

```

2. Hugging Face Login

#### Request
```
POST http://localhost:8888/aqua/hf_login
 
{
  "token": "your token here"
}
```

#### Response
```
{
    "status": 200,
    "message": "login successful"
}
```

3. Check if user is already logged in

#### Request
```
POST http://localhost:8888/aqua/hf_login
 
{
  "token": "your token here"
}
```

#### Response
```
{
    "status": 200,
    "message": "login successful"
}
```

### Unit Tests

- Will be updated in a separate PR.


ads/aqua/extension/common_handler.py
```diff
@@ -46,16 +48,75 @@ def get(self):
...
+                "The AI Quick actions extension is not compatible in the given region."
             )
 
 
+class NetworkStatusHandler(AquaAPIhandler):
+    """Handler to check internet connection."""
+
+    @handle_exceptions
+    def get(self):
+        requests.get("https://huggingface.com", timeout=2)
```
darenr, line 65: Is 2 seconds long enough?

darenr (approved):

ads/aqua/extension/common_handler.py
VipulMascarenhas: I think so - we had tested this extensively last time before these changes were removed. I'll keep an action item to review this again once we do end to end testing.

mrDzurb: I think would be better to move the default timeout to the env vars. In this case we could regulate it. We did the same with the

```
THREADED_DEFAULT_TIMEOUT = 5
try:
    THREADED_DEFAULT_TIMEOUT = int(
        os.environ.get("THREADED_DEFAULT_TIMEOUT", THREADED_DEFAULT_TIMEOUT)
    )
except ValueError:
    pass
```

I would suggest to add a method in the utils, like we did in evaluation.

mrDzurb (approved):

ads/aqua/extension/common_handler.py
```diff
@@ -46,16 +48,75 @@ def get(self):
...
+        if not input_data:
+            raise HTTPError(400, Errors.NO_INPUT_DATA)
+
+        token = input_data.get("token")
+
+        if not token:
+            raise HTTPError(400, Errors.MISSING_REQUIRED_PARAMETER.format("token"))
+
+        # Login to HF
+        huggingface_hub.login(token=token, new_session=False)
```
kumar-shivam-ranjan, line 95: what if token provided is invalid?  api is returning 5xx as of now in such cases.  shouldn't it be 4xx?

ads/aqua/extension/common_handler.py
VipulMascarenhas: added.

ads/aqua/extension/common_handler.py
VipulMascarenhas: good catch, updated the API to return 4xx.

PR #915, opened 2024-07-26
[ODSC-60499] Update validation for models registered from object storage

VipulMascarenhas:
### Description

This PR covers the following changes:
- Register api takes in a parameter download_from_hf, set to True by default. For API call via model_handler, this is set as False. For CLI command, this will be set to true by default. The changes for HF flow will be covered in a different PR.
- AquaModelSummary has a field name change, model_format to model_formats.
- Validation changes when registering model via object storage.

#### Request
```
POST http://localhost:8888/aqua/model

{
    "model": "google/gemma-2b-it",
    "os_path": "oci://<bucket>@<namespace>/<prefix>",
    "inference_container": "odsc-tgi-serving"
}
```

#### Response
```
{
    "compartment_id": "ocid1.compartment.oc1..<ocid>",
    "icon": "",
    "id": "ocid1.datasciencemodel.oc1.iad.<ocid>",
    "is_fine_tuned_model": false,
    "license": "gemma",
    "name": "google/gemma-2b-it",
    "organization": "Google",
    "project_id": "ocid1.datascienceproject.oc1.iad.<ocid>",
    "tags": {
        "aqua_service_model": "ocid1.datasciencemodel.oc1.iad.<ocid>",
        "license": "gemma",
        "task": "text_generation",
        "model_format": "SAFETENSORS",
        "organization": "Google",
        "aqua_custom_base_model": "true",
        "OCI_AQUA": "active",
        "ready_to_fine_tune": "true"
    },
    "task": "text_generation",
    "time_created": "2024-07-26 02:31:33.316000+00:00",
    "console_link": "https://cloud.oracle.com/data-science/models/ocid1.datasciencemodel.oc1.iad.<ocid>?region=us-ashburn-1",
    "search_text": "ocid1.datasciencemodel.oc1.iad.<ocid>,gemma,text_generation,SAFETENSORS,Google,true,active,true",
    "ready_to_deploy": true,
    "ready_to_finetune": true,
    "ready_to_import": false,
    "nvidia_gpu_supported": true,
    "arm_cpu_supported": false,
    "model_formats": [
        "SAFETENSORS"
    ],
    "model_card": "<content>",
    "inference_container": "odsc-tgi-serving",
    "finetuning_container": "odsc-llm-fine-tuning",
    "evaluation_container": "odsc-llm-evaluate",
    "artifact_location": "oci://<bucket>@<namespace>/<prefix>"
}
```

### Unit Tests

Tests will be covered in a separate PR.

### DIFF
--- a/ads/aqua/model/model.py
+++ b/ads/aqua/model/model.py
@@ -784,8 +784,6 @@ class AquaModelApp(AquaApp):
         self,
         os_path: str,
         model_name: str,
-        inference_container: str,
-        finetuning_container: str,
         verified_model: DataScienceModel,
     ) -> ModelValidationResult:
         """
@@ -794,8 +792,6 @@ class AquaModelApp(AquaApp):
         Args:
             os_path (str): OCI  where the model is uploaded - oci://bucket@namespace/prefix
             model_name (str): name of the model
-            inference_container (str): selects service defaults
-            finetuning_container (str): selects service defaults
             verified_model (DataScienceModel): If set, then copies all the tags and custom metadata information from the service verified model
 
         Returns:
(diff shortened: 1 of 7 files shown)


ads/aqua/model/model.py
```diff
@@ -756,141 +780,142 @@ def get_model_files(os_path: str, model_format: ModelFormat) -> [str]:
...
+                model_formats.append(ModelFormat.GGUF)
+
+        validation_result.model_formats = model_formats
+
+        # now as we know that at least one type of model files exist, validate the content of oss path.
+        # for safetensors, we check if config.json files exist, and for gguf format we check if files with
+        # gguf extension exist.
+        for model_format in model_formats:
+            if (
+                model_format == ModelFormat.SAFETENSORS
```
lu-ohai, line 835: Shouldn't all of these be `ModelFormat.SAFETENSORS.value`? Same as GGUF

VipulMascarenhas: the check is correct, but there could be a issue with serializer when output comes up on CLI or notebook. We can have a separate PR to update this from Enum to ExtendedEnum.

lu-ohai: The error comes from [here](https://github.com/oracle/accelerated-data-science/blob/656e6e11e99a31e3895d6ca50c2c9c07116e2182/ads/aqua/model/model.py#L1027), can't set `model_formats` as a list of `ModelFormat`. See same error below:
Screenshot 2024-07-29 at 3 59 02 PM

VipulMascarenhas: yes, will update this.

kumar-shivam-ranjan:
If the container selected is odsc-llama-cpp-serving while registering, there shouldn't be check for config.json right?  bt in the current implementation , it is doing and if there is no config.json found, it is throwing error.
Is this expected?

VipulMascarenhas:
> If the container selected is odsc-llama-cpp-serving while registering and the artifact has only gguf files no safetensor, there shouldn't be check for config.json right? bt in the current implementation , it is doing and if there is no config.json found, it is throwing error. Is this expected?

the exception wasn't handled correctly in that case, it's fixed now. thanks for catching this.

ads/aqua/model/model.py
```diff
@@ -756,142 +791,272 @@ def get_model_files(os_path: str, model_format: ModelFormat) -> [str]:
...
+
+        Returns
+        -------
+        model_artifact_path (str): Location where the model artifacts are downloaded.
+
+        """
+        # Download the model from hub
+        model_name = import_model_details.model
+        local_dir = import_model_details.local_dir
+        if not local_dir:
```
VipulMascarenhas, line 1030: note: we're downloading to cached folder first and then to local folder. Need to test this two step download, most likely we just need to download to local dir.

PR #924, opened 2024-08-06
Get model files changes for cli

kumar-shivam-ranjan:


ads/aqua/model/model.py
```diff
@@ -775,6 +775,11 @@ def get_model_files(os_path: str, model_format: ModelFormat) -> List[str]:
             List[str]: A list of model file names.
 
         """
+        try:
+            model_format = ModelFormat(model_format.upper())
+        except ValueError as err:
+            raise AquaValueError(f"Invalid model format: {model_format}") from err
```
lu-ohai, line 781: We could also add the supported formats in the error message

PR #934, opened 2024-08-22
ODSC-61571 : Add RCF Implementation

codeloop:
[ODSC-61571](https://jira.oci.oraclecorp.com/browse/ODSC-61571) : Add RandomCutForest as an AD Operator model .


ads/opctl/operator/lowcode/anomaly/model/base_model.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, config: AnomalyOperatorConfig, datasets: AnomalyDatasets):
     def generate_report(self):
         """Generates the report."""
         import matplotlib.pyplot as plt
+        plt.rcParams.update({'figure.max_open_warning': 0})
```
ahosler, line 58: great idea

ads/opctl/operator/lowcode/anomaly/model/randomcutforest.py
```diff
@@ -0,0 +1,116 @@
...
+        from rrcf import RCTree
+
+        model_kwargs = self.spec.model_kwargs
+
+        anomaly_output = AnomalyOutput(date_column="index")
+
+        # Set tree parameters
+        num_trees = model_kwargs.get("num_trees", 200)
+        shingle_size = model_kwargs.get("shingle_size", None)
+        anamoly_threshold = model_kwargs.get("anamoly_threshold", 95)
```
ahosler, line 39: should this be "anomaly_threshold" ?

ads/opctl/operator/lowcode/anomaly/model/base_model.py
```diff
@@ -87,43 +88,57 @@ def generate_report(self):
...
-                # Plot anomalies
-                for i in anomaly_indices:
-                    ax.scatter(time_col[i], y[i], color="red", marker="o")
-                plt.xlabel(date_column)
-                plt.ylabel(col)
-                plt.title(f"`{col}` with reference to anomalies")
-                figure_blocks.append(rc.Widget(ax))
-
-        blocks.append(rc.Group(*figure_blocks, label=target))
+            if target in anomaly_output.list_categories():
```
ahosler, line 98: Is this a way to get rid of the "series 0" in the report? 
Is there an edge case where the index column value is the same as the name of the target col?

codeloop: We are plotting for each series in the dataset, there may be edge cases where we don't have sufficient data for a particular series hence no anomaly scores, this ensures we try to plot only the series for which the anomaly scores are available.

prasankh (commented):
Can we add this model to existing test cases ?

prasankh (approved):

ahosler (approved):

PR #936, opened 2024-08-23
Support adding artifacts for container jobs

qiuosier:
This PR includes updates from #908 , and the following updates:
* Adds the `with_artifact()` method to support creating a container job with artifact.
* Raise error when running ads opctl conda create/publish if the subprocess has non-zero exit code.


liudmylaru (dismissed):

liudmylaru (dismissed):

ads/jobs/builders/infrastructure/dsc_job.py
```diff
@@ -312,7 +312,7 @@ def _create_with_oci_api(self) -> None:
         logger.debug(oci_model)
         res = self.client.create_job(oci_model)
         self.update_from_oci_model(res.data)
-        if self.lifecycle_state == "ACTIVE":
+        if not self.artifact:
```
mrDzurb, line 315: Can you add details about this change?

mrDzurb (dismissed):

ads/jobs/builders/infrastructure/dsc_job.py
qiuosier: The container job will be `ACTIVE` even if artifact is not uploaded. So we need to check if there is artifact to be uploaded instead of checking if it is `ACTIVE`.

mrDzurb (approved):

liudmylaru (approved):

PR #937, opened 2024-08-28
Introduced post processing

prasankh:
[ODSC-61573](https://jira.oci.oraclecorp.com/browse/ODSC-61573)

Following schema can be used to set min value for the forecast value
```
postprocessing:
 enabled: true
 steps:
   set_min_forecast: 0
```


ads/opctl/operator/lowcode/forecast/model/prophet.py
```diff
@@ -81,6 +81,13 @@ def _train_model(self, i, series_id, df, model_kwargs):
             )
 
             data = self.preprocess(df, series_id)
+            cap = data['y'].max() * 1.2
```
ahosler, line 84: Do we need to have a cap?

ads/opctl/operator/lowcode/forecast/model/prophet.py
```diff
@@ -81,6 +81,13 @@ def _train_model(self, i, series_id, df, model_kwargs):
...
+            if self.spec.postprocessing.enabled:
+                floor = self.spec.postprocessing.steps.set_min_forecast
+            else:
+                floor = data['y'].min() * 0.8
```
ahosler, line 88: Shouldn't we leave this to the framework? Why 0,8?

ads/opctl/operator/lowcode/forecast/schema.yaml
```diff
@@ -317,6 +317,25 @@ spec:
               required: false
               default: false
 
+    postprocessing:
```
ahosler, line 320: This is a great way to add it 👍

ads/opctl/operator/lowcode/forecast/model/prophet.py
prasankh: Yes, It's mandatory with logistic growth function. Removed logistic growth function for now as it fails during generating local explanations.

ads/opctl/operator/lowcode/forecast/model/prophet.py
prasankh: It's mandatory with logistic growth function. Removed logistic growth function for now.

codeloop (approved):

PR #942, opened 2024-09-06
[AQUA][Evaluate] Externalize Supported Shapes List to Global Config.

mrDzurb:
## Description
* Shifts the management of the supported shapes list from being hardcoded in the ADS SDK to being stored and managed at the global service config level.
* Reads the evaluation inferencing parameters from the service config.
* Introduces the inferencing service model sampling parameters. The system parameters that will not be visible for user, but will be applied implicitly for evaluation. The service model params will be extracted based on the container family and version.


darenr (approved):

ads/aqua/config/config.py
```diff
@@ -15,7 +15,7 @@
 DEFAULT_EVALUATION_CONTAINER = "odsc-llm-evaluate"
 
 
-@cached(cache=TTLCache(maxsize=1, ttl=timedelta(hours=5), timer=datetime.now))
```
mayoor, line 18: I am not big fan of caching in our application unless we see that the users are frequently accessing this page. The caching prevents us from urgently addressing a fix if there is a problem. We will have to wait for the cache expiry.

qiuosier (approved):

PR #955, opened 2024-09-27
AD | Added AutoTS/Sklearn Models as Operator Models with Contamination Parameter Support

prasankh:
[ODSC-62963](https://jira.oci.oraclecorp.com/browse/ODSC-62963)
The following models have been added:

* IQR (Interquartile Range)
* LOF (Local Outlier Factor)
* Z-Score
* Rolling Z-Score
* Isolation Forest
* MAD (median abs dev)
* EE (Elliptic Envelope)

Some models like **EE**, **Isolation Forest**, and **LOF** support the `contamination` parameter while others do not.


ahosler (approved):

ahosler:
Can we also update the unit tests from "autots" to iterate through each of these models?

codeloop (approved):

PR #956, opened 2024-09-29
Add sf merlion for time-based and forecast AD

codeloop:
Add support for salesforce merlion model for time-based and forecast anomaly detection.


pyproject.toml
```diff
@@ -179,7 +179,8 @@ anomaly  = [
   "oracledb",
   "report-creator==1.0.9",
   "rrcf==0.4.4",
-  "scikit-learn"
+  "scikit-learn",
+  "salesforce-merlion[all]==2.0.4"
```
ahosler, line 183: What features does the "all" get us?

ads/opctl/operator/lowcode/anomaly/schema.yaml
```diff
@@ -364,6 +364,7 @@ spec:
         - oneclasssvm
         - isolationforest
         - randomcutforest
+        - merlion_ad
```
ahosler, line 367: can we change this to "merlion"?

ads/opctl/operator/lowcode/anomaly/model/randomcutforest.py
```diff
@@ -36,7 +36,7 @@ def _build_model(self) -> AnomalyOutput:
         # Set tree parameters
         num_trees = model_kwargs.get("num_trees", 200)
         shingle_size = model_kwargs.get("shingle_size", None)
-        anomaly_threshold = model_kwargs.get("anamoly_threshold", 95)
+        anomaly_threshold = model_kwargs.get("anomaly_threshold", 95)
```
ahosler, line 39: great catch

ahosler (changes requested):
In testing, I see that the anomaly score is no longer normalized from 0-1. Does Merlion have a different interpretation for the meaning of this score (CSV uploaded)? 
I believe we should stick with the [0,1] convention unless we have good reason to change.

Also Merlion seemed over-reactive on a few of the tests I've run. Are there any default settings we should be tweaking?

[outliers.csv](https://github.com/user-attachments/files/17294813/outliers.csv)

pyproject.toml
codeloop: `[all]` comprises of AD, forecast needed for AD, DL support we have some deep learning based AD algos, + few things on distributed processing & others.

codeloop:
> 1. In testing, I see that the anomaly score is no longer normalized from 0-1. Does Merlion have a different interpretation for the meaning of this score ? (CSV uploaded below)
>    I believe we should stick with the [0,1] convention unless we have good reason to change.
> 2. Also Merlion seemed over-reactive on a few of the tests I've run. Are there any default settings we should be tweaking?
> 3. Could Merlion be made to work for non-timeseries AD? Or is it primarily time-based?
> 
> [outliers.csv](https://github.com/user-attachments/files/17294813/outliers.csv)



> 1. In testing, I see that the anomaly score is no longer normalized from 0-1. Does Merlion have a different interpretation for the meaning of this score ? (CSV uploaded below)
>    I believe we should stick with the [0,1] convention unless we have good reason to change.
> 2. Also Merlion seemed over-reactive on a few of the tests I've run. Are there any default settings we should be tweaking?
> 3. Could Merlion be made to work for non-timeseries AD? Or is it primarily time-based?
> 
> [outliers.csv](https://github.com/user-attachments/files/17294813/outliers.csv)

1. Added score normalization.
2. We can use an alarm threshold param to control the sensitivity of the models.
3. Currently it only supports time-series data.

prasankh (approved):

ahosler (approved):

PR #960, opened 2024-10-01
Update version to 2.12.0

qiuosier:


VipulMascarenhas (dismissed):

docs/source/release_notes.rst
```diff
@@ -2,7 +2,7 @@
 Release Notes
 =============
 
-2.11.20
+2.12.1
```
mrDzurb, line 5: Just curious, why it is ```12.1``` but not ```12.0```? :)

mrDzurb (dismissed):

docs/source/release_notes.rst
qiuosier: I saw we did not have `11.0`. But looks like we have `10.0`. Which one do you think we should use?

qiuosier: Let me update it to `12.0`

VipulMascarenhas (approved):

lu-ohai (approved):

PR #962, opened 2024-10-09
Feature/ms backup retention

iamishaan:
This PR aims to update the data science model class to introduce new feature of Model Backups and retention.
Also to provide a new method/api call to restore an archived model artifact.
User can enable model backup or retention for their model while creation or while updating a model.
Both of them are Optional fields.
attaching screenshots  for model creation with the 2 new fields 
Screenshot 2024-10-24 at 5 21 48 PM
Screenshot 2024-10-24 at 5 22 13 PM


lu-ohai:
@iamishaan Could you add some user experience example here and also update the user guide? DataScience Model related user guide should be placed under [here](https://accelerated-data-science.readthedocs.io/en/latest/user_guide/model_registration/introduction.html).

ads/model/datascience_model.py
```diff
@@ -685,6 +711,114 @@ def with_model_file_description(
...
+
+        Returns
+        -------
+        DataScienceModel
+            The `DataScienceModel` instance (self) for method chaining.
+        """
+        if retention_setting and isinstance(retention_setting, dict):
+            try:
+                retention_setting = ModelRetentionSetting.from_dict(retention_setting)
+            except Exception as err:
```
lu-ohai, line 742: We should probably raise error here and provide the required parameters for `RetentionSetting`

ads/model/datascience_model.py
```diff
@@ -685,6 +711,114 @@ def with_model_file_description(
...
+
+        Returns
+        -------
+        DataScienceModel
+            The `DataScienceModel` instance (self) for method chaining.
+        """
+        if backup_setting and isinstance(backup_setting, dict):
+            try:
+                backup_setting = ModelBackupSetting.from_dict(backup_setting)
+            except Exception as err:
```
lu-ohai, line 777: Same here

ads/model/datascience_model.py
```diff
@@ -685,6 +711,114 @@ def with_model_file_description(
...
+    def retention_operation_details(self) -> ModelRetentionOperationDetails:
+        """
+        Gets the retention_operation_details of this Model using the spec constant.
+
+        :return: The retention_operation_details of this Model.
+        :rtype: ModelRetentionOperationDetails
+        """
+        return self.get_spec(self.CONST_RETENTION_OPERATION_DETAILS)
+
+    @retention_operation_details.setter
```
lu-ohai, line 792: Why do we need the property setter, instead of adding the -with method?

ads/model/model_metadata.py
```diff
@@ -1761,3 +1761,328 @@ def __repr__(self):
...
+        Serializes the backup settings into a YAML string.
+    validate(self) -> bool:
+        Validates the backup settings details.
+    """
+
+    def __init__(self, is_backup_enabled: Optional[bool] = None, backup_region: Optional[str] = None,
+                 customer_notification_type: Optional[CustomerNotificationType] = None):
+        self.is_backup_enabled = is_backup_enabled if is_backup_enabled is not None else False
+        self.backup_region = backup_region
+        self.customer_notification_type = customer_notification_type if customer_notification_type is not None else CustomerNotificationType.NONE
```
lu-ohai, line 1797: This can be simplified as:
```
self.customer_notification_type = customer_notification_type or CustomerNotificationType.NONE
```
Same below

ads/model/model_metadata.py
```diff
@@ -1761,3 +1761,328 @@ def __repr__(self):
...
+        if not isinstance(self.is_backup_enabled, bool):
+            return False
+        if self.backup_region and not isinstance(self.backup_region, str):
+            return False
+        if not isinstance(self.customer_notification_type, CustomerNotificationType):
+            return False
+        return True
+
+    def __repr__(self):
+        return f"ModelBackupSetting(is_backup_enabled={self.is_backup_enabled}, backup_region={self.backup_region}, customer_notification_type={self.customer_notification_type})"
```
lu-ohai, line 1841: Can we change this to represent the info as yaml format? Same below

ads/model/model_metadata.py
```diff
@@ -1761,3 +1761,328 @@ def __repr__(self):
...
+    """
+
+    def __init__(self,
+                 archive_state: Optional[SettingStatus] = None,
+                 archive_state_details: Optional[str] = None,
+                 delete_state: Optional[SettingStatus] = None,
+                 delete_state_details: Optional[str] = None,
+                 time_archival_scheduled: Optional[int] = None,
+                 time_deletion_scheduled: Optional[int] = None):
+        self.archive_state = archive_state if archive_state is not None else SettingStatus.PENDING
```
lu-ohai, line 1948:
```
self.archive_state = archive_state or SettingStatus.PENDING
```
Same below

ads/model/model_metadata.py
```diff
@@ -1761,3 +1761,328 @@ def __repr__(self):
...
+    from_json(cls, json_str: str) -> 'ModelBackupOperationDetails':
+        Constructs backup operation details from a JSON string.
+    to_yaml(self) -> str:
+        Serializes the backup operation details into a YAML string.
+    validate(self) -> bool:
+        Validates the backup operation details.
+    """
+
+    def __init__(self,
+                 backup_state: Optional['SettingStatus'] = None,
```
lu-ohai, line 2035: Why the SettingStatus is string format here?

lu-ohai (changes requested):
Added some comments.

mrDzurb:
@iamishaan could you add more details in the PR description?

ads/model/service/oci_datascience_model.py
```diff
@@ -278,6 +278,40 @@ def get_artifact_info(self) -> Dict:
...
+            If model artifact not found.
+        """
+        try:
+            # Call the underlying client method to restore the model artifact
+            return self.client.restore_archived_model_artifact(
+                model_id=model_id,
+                restore_model_for_hours_specified=restore_model_for_hours_specified,
+            ).data.content
+        except ServiceError as ex:
+            if ex.status == 404:
```
mrDzurb, line 312: In case if any other error occurs, the method will not return any result.

```
except ServiceError as ex:
      if ex.status == 404:
            raise ModelArtifactNotFoundError()
      raise
```

ads/model/service/oci_datascience_model.py
```diff
@@ -278,6 +278,40 @@ def get_artifact_info(self) -> Dict:
                 raise ModelArtifactNotFoundError()
         return {}
 
+    @check_for_model_id(
+        msg="Model needs to be restored before the archived artifact content can be accessed."
+    )
+    def restore_archived_model_artifact(
+            self, model_id: str, restore_model_for_hours_specified: Optional[int] = None
```
mrDzurb, line 285: NIT: The ```restore_duration_hours``` would be more intuitive name.

ads/model/datascience_model.py
```diff
@@ -120,6 +124,19 @@ class DataScienceModel(Builder):
...
+    backup_setting: ModelBackupSetting
+        The value to assign to the backup_setting property of this CreateModelDetails.
+    retention_setting: ModelRetentionSetting
+        The value to assign to the retention_setting property of this CreateModelDetails.
+    retention_operation_details: ModelRetentionOperationDetails
+        The value to assign to the retention_operation_details property for the Model.
+    backup_operation_details: ModelBackupOperationDetails
+        The value to assign to the backup_operation_details property for the Model.
+
+
```
mrDzurb, line 136: NIT: remove extra empty lines

ads/model/datascience_model.py
```diff
@@ -907,6 +1034,49 @@ def _remove_file_description_artifact(self):
...
+        restore_model_for_hours_specified: Optional[int] = None,
+    ):
+        """
+        Restore archived model artifact.
+
+        Parameters
+        ----------
+        model_id : str
+            The `OCID` of the model to be restored.
+            restore_model_for_hours_specified : Optional[int]
```
mrDzurb, line 1049: NIT: formatting?

ads/model/datascience_model.py
```diff
@@ -907,6 +1034,49 @@ def _remove_file_description_artifact(self):
         if self.local_copy_dir:
             shutil.rmtree(self.local_copy_dir, ignore_errors=True)
 
+    def restore_model(
+        self,
+        model_id: str,
```
mrDzurb, line 1039: This either should be a class method or ```model_id``` parameter is not required. See the update nethod.

ads/model/datascience_model.py
```diff
@@ -907,6 +1034,49 @@ def _remove_file_description_artifact(self):
...
+        restore_model_for_hours_specified: Optional[int] = None,
```
mrDzurb, line 1040: SHouldn't we have some default value? FOr example 24 hours?

ads/model/datascience_model.py
```diff
@@ -907,6 +1034,49 @@ def _remove_file_description_artifact(self):
...
+        if restore_model_for_hours_specified is not None:
+            if (
+                not isinstance(restore_model_for_hours_specified, int)
+                or restore_model_for_hours_specified <= 0
+            ):
+                raise ValueError(
+                    "restore_model_for_hours_specified must be a positive integer."
+                )
+
+        self.dsc_model.restore_archived_model_artifact(
```
mrDzurb, line 1075: It can be something like this: 
```
self.dsc_model.restore_archived_model_artifact(
            model_id=self.id,
            restore_model_for_hours_specified=restore_model_for_hours_specified,
        )

return self.sync()
```

check the update method.

ads/model/datascience_model.py
```diff
@@ -1162,6 +1332,8 @@ def _init_complex_attributes(self):
         self.with_provenance_metadata(self.provenance_metadata)
         self.with_input_schema(self.input_schema)
         self.with_output_schema(self.output_schema)
+        self.with_backup_setting(self.backup_setting)
```
mrDzurb, line 1335: Probably we don't need to add it here. The complex attributes mean that some extra logic is required to initialize these attributes. For example if to check the input schema it has some try/catch mechanism to validate the schema.

ads/model/datascience_model.py
```diff
@@ -1197,6 +1369,8 @@ def _to_oci_dsc_model(self, **kwargs):
                 dsc_spec[dsc_attr] = value
 
         dsc_spec.update(**kwargs)
+        print("Model Dsc spec")
```
mrDzurb, line 1372: Should be removed?

ads/model/datascience_model.py
```diff
@@ -1219,10 +1393,15 @@ def _update_from_oci_dsc_model(
...
             self.CONST_CUSTOM_METADATA: ModelCustomMetadata._from_oci_metadata,
             self.CONST_DEFINED_METADATA: ModelTaxonomyMetadata._from_oci_metadata,
+            self.CONST_BACKUP_SETTING: ModelBackupSetting.to_dict,
+            self.CONST_RETENTION_SETTING: ModelRetentionSetting.to_dict,
         }
 
         # Update the main properties
+
         self.dsc_model = dsc_model
+        print("Model Details from here")
```
mrDzurb, line 1403: Should be removed?

ads/model/model_metadata.py
```diff
@@ -1761,3 +1761,391 @@ def __repr__(self):
...
         return self.to_yaml()
+
+
+class CustomerNotificationType(str, metaclass=ExtendedEnumMeta):
+    NONE = "NONE"
+    ALL = "ALL"
+    ON_FAILURE = "ON_FAILURE"
+    ON_SUCCESS = "ON_SUCCESS"
+
+    _value_map = {
```
mrDzurb, line 1772: Why do we need this map?

ads/model/model_metadata.py
```diff
@@ -1761,3 +1761,391 @@ def __repr__(self):
...
+        "NONE": NONE,
+        "ALL": ALL,
+        "ON_FAILURE": ON_FAILURE,
+        "ON_SUCCESS": ON_SUCCESS,
+    }
+
+    @classmethod
```
mrDzurb, line 1779: Not sure about all of the methods in this class, why do we need them?

ads/model/model_metadata.py
```diff
@@ -1761,3 +1761,391 @@ def __repr__(self):
...
+
+class SettingStatus(str, metaclass=ExtendedEnumMeta):
+    """Enum to represent the status of retention settings."""
+
+    PENDING = "PENDING"
+    SUCCEEDED = "SUCCEEDED"
+    FAILED = "FAILED"
+
+    @classmethod
+    def is_valid(cls, state: str) -> bool:
```
mrDzurb, line 1984: How this method is used?

ads/model/model_metadata.py
```diff
@@ -1761,3 +1761,391 @@ def __repr__(self):
...
+    @classmethod
+    def is_valid(cls, value):
+        return value in (cls.NONE, cls.ALL, cls.ON_FAILURE, cls.ON_SUCCESS)
+
+    @property
+    def value(self):
+        return str(self)
+
+
+class ModelBackupSetting:
```
mrDzurb, line 1801: I think it would be better to move the new added classes to the ```datascience_model.py```. The ```model_metadata.py``` is used for the different purpose. Also please take a look on the ```DataClassSerializable``` class and see how it used in the code. It includes that many pre-define methods.

ads/model/service/oci_datascience_model.py
iamishaan: keeping it consistent with the python sdk

ads/model/service/oci_datascience_model.py
iamishaan: update : taking off the try catch

ads/model/model_metadata.py
iamishaan: it is used to verify if the notification type is valid

iamishaan: was using it to validate type earlier,  removing it now

ads/model/model_metadata.py
iamishaan: to verify if the notificationType instance is being created with a valid notification type

ads/model/datascience_model.py
iamishaan: updated the code

ads/model/datascience_model.py
iamishaan: removed

ads/model/datascience_model.py
iamishaan: done

ads/model/datascience_model.py
iamishaan: done

ads/model/datascience_model.py
iamishaan: removed

ads/model/model_metadata.py
iamishaan: done

ads/model/model_metadata.py
iamishaan: done

ads/model/model_metadata.py
iamishaan: done

ads/model/datascience_model.py
```diff
@@ -78,6 +78,374 @@ class InvalidArtifactType(Exception):  # pragma: no cover
...
+    SUCCEEDED = "SUCCEEDED"
+    FAILED = "FAILED"
+
+    @classmethod
+    def is_valid(cls, state: str) -> bool:
+        """Validates the given state against allowed SettingStatus values."""
+        return state in (cls.PENDING, cls.SUCCEEDED, cls.FAILED)
+
+
+class ModelBackupSetting:
```
mrDzurb, line 109: The ```class ModelBackupSetting:``` can be replaced with 

```
@dataclass(repr=False)
class ModelBackupSetting(DataClassSerializable):
      is_backup_enabled: Optional[bool] = None,
      backup_region: Optional[str] = None,
      customer_notification_type: Optional[str] = None

      def validate(self) -> bool:
                    pass
```

The DataClassSerializable is already implements bunch of the helper functions like form_dict, to_tict, from_json, e.t.c. 

DITO for the other classes.

ads/model/datascience_model.py
mrDzurb: With this update the case shared below will not work. I would rather prefer to remove the ```model_id``` attribute from the function. 

This case will not work:
```
DataScienceModel.from_id("OCID"). restore_model()
```
It will expect from e to provide OCID one more time, 
like this
```
DataScienceModel.from_id("OCID"). restore_model("OCID")
```
But why should i do this, if I already provided the OCID before? 

**The better solution would be**

```
    def restore_model(
            cls,
            restore_model_for_hours_specified: Optional[int] = HOURS_24,
    ):

    if not self.id:
          raise ValueError("Model needs to be saved to the model catalog before it can be updated.")
```

Or if we really want to support both cases below

```
DataScienceModel.from_id("OCID"). restore_model()
DataScienceModel.restore_model("OCID")
```

Then the solution could be 

```
@class_or_instance_method
def restore_model(
            cls,
            model_id: Optional[str] = None,
            restore_model_for_hours_specified: Optional[int] = HOURS_24,
    ):

   if not inspect.isclass(cls): # if the instance used
        pass
   
   # otherwise the class is used
   
   # see GenericModel.update_deployment as example.    
```

iamishaan: Have updated the method to use self and model_id isn't a required parameter anymore now.
just passing it to `restore_archived_model_artifact` method here

```
self.dsc_model.restore_archived_model_artifact(
            model_id=self.id,
            restore_model_for_hours_specified=restore_model_for_hours_specified,
        )
```

ads/model/datascience_model.py
iamishaan: That is already handled in backend, if the value passed is null, it defaults it to 24 hours

ads/model/datascience_model.py
iamishaan: updated

ads/model/datascience_model.py
iamishaan: removed

ads/model/service/oci_datascience_model.py
```diff
@@ -278,6 +278,34 @@ def get_artifact_info(self) -> Dict:
...
```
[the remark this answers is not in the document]
iamishaan: have refactored the method

ads/model/model_metadata.py
```diff
@@ -256,9 +255,9 @@ def to_json(self):
         return json.dumps(self.to_dict())
 
     def to_json_file(
-        self,
-        file_path: str,
-        storage_options: dict = None,
+            self,
```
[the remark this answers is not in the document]
iamishaan: undo-ed

ads/model/datascience_model.py
```diff
@@ -78,6 +78,374 @@ class InvalidArtifactType(Exception):  # pragma: no cover
...
 
 
+class CustomerNotificationType(str, metaclass=ExtendedEnumMeta):
+    NONE = "NONE"
+    ALL = "ALL"
+    ON_FAILURE = "ON_FAILURE"
+    ON_SUCCESS = "ON_SUCCESS"
+
+    @classmethod
+    def is_valid(cls, value):
```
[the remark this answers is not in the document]
iamishaan: updated

ads/model/datascience_model.py
```diff
@@ -78,6 +78,374 @@ class InvalidArtifactType(Exception):  # pragma: no cover
...
+        return value in (cls.NONE, cls.ALL, cls.ON_FAILURE, cls.ON_SUCCESS)
+
+    @property
+    def value(self):
```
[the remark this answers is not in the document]
iamishaan: removed the value method

ads/model/datascience_model.py
mrDzurb: The ```return self.sync()``` would get the updated model, in this case user would have the object containing the fresh status and details. Otherwise, after the calling this method user will have an old object.

ads/model/datascience_model.py
```diff
@@ -907,16 +1366,54 @@ def _remove_file_description_artifact(self):
...
-        self,
-        target_dir: str,
-        auth: Optional[Dict] = None,
-        force_overwrite: Optional[bool] = False,
-        bucket_uri: Optional[str] = None,
-        region: Optional[str] = None,
-        overwrite_existing_artifact: Optional[bool] = True,
-        remove_existing_artifact: Optional[bool] = True,
-        timeout: Optional[int] = None,
+            self,
```
mrDzurb, line 1408: Formatting issue?

mrDzurb (approved):

mrDzurb (approved):

lu-ohai (approved):

PR #976, opened 2024-10-22
Moving config from ADS to UI for unverified models

kumar-shivam-ranjan:
This PR is intended to remove dependency of unverified models default config from ADS. 
Currently, the default shapes config for deployment, finetuning and evaluation are hardcoded in ADS. This makes it difficult to make changes ad-hoc with new ADS release. 
This info is needed for unverified models only and can be maintained at client side.
Config that needs to be removed are
- [Deployment config](https://github.com/oracle/accelerated-data-science/blob/main/ads/aqua/config/deployment_config_defaults.json)
- [Resource limit names](https://github.com/oracle/accelerated-data-science/blob/main/ads/aqua/config/resource_limit_names.json)
- [Finetuning config](https://github.com/oracle/accelerated-data-science/blob/505985bd77239fdd69db0d61f1f412cce38d1690/ads/aqua/config/config.py#L35)

### UTs
Screenshot 2024-10-29 at 12 45 21 AM


ads/aqua/finetuning/finetuning.py
```diff
@@ -563,8 +562,9 @@ def get_finetuning_config(self, model_id: str) -> Dict:
 
         config = self.get_config(model_id, AQUA_MODEL_FINETUNING_CONFIG)
         if not config:
-            logger.info(f"Fetching default fine-tuning config for model: {model_id}")
-            config = get_finetuning_config_defaults()
+            logger.info(
+                f"default fine-tuning config will be used for model: {model_id}"
```
VipulMascarenhas, line 566: this info message can be misleading since user might see this in ads logs and expect some default to be returned, but the defaults are applied outside of ads. Maybe a better phrasing would be "Fine-tuning config for custom model {model_id} is not available."  We can also use logger.debug instead, user need not see this message every time.
@mrDzurb thoughts?

ads/aqua/modeldeployment/deployment.py
```diff
@@ -44,11 +43,8 @@
 from ads.common.object_storage_details import ObjectStorageDetails
 from ads.common.utils import get_log_links
 from ads.config import (
-    AQUA_CONFIG_FOLDER,
     AQUA_DEPLOYMENT_CONTAINER_METADATA_NAME,
-    AQUA_DEPLOYMENT_CONTAINER_OVERRIDE_FLAG_METADATA_NAME,
     AQUA_MODEL_DEPLOYMENT_CONFIG,
-    AQUA_MODEL_DEPLOYMENT_CONFIG_DEFAULTS,
```
VipulMascarenhas, line 51: can you please the unused constants and json files/functions that were deprecated here since we're not using them anymore?

VipulMascarenhas (commented):
are we rolling out the relevant UI changes this week? If that's the case, we can set the target branch to main and release ADS this Thursday. That way, the next NBVM update in a week would have these changes.

ads/aqua/finetuning/finetuning.py
kumar-shivam-ranjan: Updated

ads/aqua/modeldeployment/deployment.py
kumar-shivam-ranjan: Sorry didn't understand this statement. perhaps you a word or two?

VipulMascarenhas: I missed the word - " can you please remove the unused constants and json files/functions" :)

mrDzurb:
Please add more details in the description. It will be helpful for the future maintenance.

VipulMascarenhas:
> > are we rolling out the relevant UI changes this week? If that's the case, we can set the target branch to main and release ADS this Thursday. That way, the next NBVM update in a week would have these changes.
> 
> UI changes are ready. I was under assumption that only embedding changes + adding logs in terminal are part of release this week. But if we plan to release config migration changes as well , we can.
> 
> Looping @mayoor @NitinMore7

yes, let's switch target to main if we're comfortable with the UI changes to go this week.
cc : @mayoor

ads/aqua/modeldeployment/deployment.py
kumar-shivam-ranjan: Sure , will update

mrDzurb (dismissed):

mrDzurb (dismissed):

VipulMascarenhas (approved):

VipulMascarenhas (approved):

PR #989, opened 2024-10-30
Adding cache delete method for model details

kumar-shivam-ranjan:
This PR is intended to add delete model details cache method which will be helpful while editing default config for unverified models.  The cache needs to be deleted for the particular model for which update operation has been performed to get latest updated data in get call.

### DIFF
--- a/ads/aqua/extension/model_handler.py
+++ b/ads/aqua/extension/model_handler.py
@@ -164,12 +164,13 @@ class AquaModelHandler(AquaAPIhandler):
 
         enable_finetuning = input_data.get("enable_finetuning")
         task = input_data.get("task")
+        app=AquaModelApp()
         self.finish(
-            AquaModelApp().edit_registered_model(
+            app.edit_registered_model(
                 id, inference_container, enable_finetuning, task
             )
         )
-        AquaModelApp().clear_model_details_cache(model_id=id)
+        app.clear_model_details_cache(model_id=id)
 
 
 class AquaModelLicenseHandler(AquaAPIhandler):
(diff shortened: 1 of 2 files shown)


ads/aqua/extension/model_handler.py
```diff
@@ -169,6 +169,7 @@ def put(self, id):
                 id, inference_container, enable_finetuning, task
             )
         )
+        AquaModelApp().clear_model_details_cache(model_id=id)
```
VipulMascarenhas, line 172: can we do the following? We can avoid initialization of client in init twice: 
```
app = AquaModelApp()
self.finish(
            app.edit_registered_model(
                id, inference_container, enable_finetuning, task
            )
        )
app.clear_model_details_cache(
...
```

ads/aqua/model/model.py
```diff
@@ -718,6 +718,22 @@ def clear_model_list_cache(
...
 
+    def clear_model_details_cache(self, model_id):
+        """
+        Allows user to clear model details cache item
+        Returns
+        -------
+            dict with the key used, and True if cache has the key that needs to be deleted.
+        """
+        res = {}
+        logger.info("Clearing _service_model_details_cache")
```
VipulMascarenhas, line 729: nit: add model id in the log
```
logger.info(f"Clearing _service_model_details_cache for {model_id}")
```

ads/aqua/extension/model_handler.py
kumar-shivam-ranjan: Updated. Thanks!

ads/aqua/model/model.py
kumar-shivam-ranjan: Updated. Thanks!

VipulMascarenhas (approved):
lgtm!

PR #996, opened 2024-11-07
Improved compability check error handling.

lu-ohai:
### Improved compability check error handling.

- Added specific message when time out error happens.


VipulMascarenhas:
@lu-ohai we can close this PR, the UI will handle this for now as discussed on slack.

PR #1016, opened 2024-11-27
Bug Fix for MD failure for deployment of Aqua models registered from OSS

VipulMascarenhas:
### Description

When registering Aqua models via OSS flow, since user does not set the task, it is not available in the freeform tags. While deploying, this should be set to empty instead of None, else model deployment creation fails with "Invalid request tags" as shown below.

```
{
    "status": 400,
    "message": "Authorization Failed: Could not create resource. Operation Name: create_model_deployment.",
    "service_payload": {
        "target_service": "data_science",
        "status": 400,
        "code": "InvalidParameter",
        "message": "Invalid request tags",
        "operation_name": "create_model_deployment",
    ...
```


darenr (approved):
For future releases typing has been deprecated for List/Dict/Tuple in favor of just list/dict/tuple

mayoor (approved):

PR #1018, opened 2024-12-03
Dataflow changes

ahosler:


VipulMascarenhas (approved):

mrDzurb (approved):

VipulMascarenhas (approved):

pyproject.toml
```diff
@@ -202,7 +201,7 @@ pii = [
   "scrubadub_spacy",
   "spacy-transformers==1.2.5",
   "spacy==3.6.1",
-  "report-creator==1.0.28",
+  "report-creator==1.0.32",
```
darenr, line 204: latest is 1.0.37

ads/opctl/config/merger.py
```diff
@@ -41,8 +39,9 @@ class ConfigMerger(ConfigProcessor):
     """
 
     def process(self, **kwargs) -> None:
-        config_string = Template(json.dumps(self.config)).safe_substitute(os.environ)
-        self.config = json.loads(config_string)
+        for key, value in self.config.items():
+            if isinstance(value, str):  # Substitute only if the value is a string
```
darenr, line 43: the comment is factual, but why only string values? My mind went to things like pulling a host and port out of the env, but that would break with only strings being pulled in? I guess a comment explaining why only keeping strings would be useful

ads/opctl/operator/lowcode/forecast/model/prophet.py
```diff
@@ -369,11 +369,7 @@ def _generate_report(self):
                 logger.debug(f"Full Traceback: {traceback.format_exc()}")
 
         model_description = rc.Text(
-            "Prophet is a procedure for forecasting time series data based on an additive "
-            "model where non-linear trends are fit with yearly, weekly, and daily seasonality, "
-            "plus holiday effects. It works best with time series that have strong seasonal "
-            "effects and several seasons of historical data. Prophet is robust to missing "
-            "data and shifts in the trend, and typically handles outliers well."
+            """Prophet is a procedure for forecasting time series data based on an additive model where non-linear trends are fit with yearly, weekly, and daily seasonality, plus holiday effects. It works best with time series that have strong seasonal effects and several seasons of historical data. Prophet is robust to missing data and shifts in the trend, and typically handles outliers well."""
```
darenr, line 372: this is what DeepSeek things sounds better:

Prophet is a time-series forecasting tool that uses an additive model.  It models non-linear trends and incorporates yearly, weekly, and daily seasonality, as well as the impact of holidays.  Prophet is particularly effective for time series exhibiting strong seasonality and benefiting from multiple years of historical data.  It's designed to be resilient to missing data, shifts in trends, and outliers.

darenr (approved):
minor comments but nothing preventing merge

ads/opctl/config/merger.py
ahosler: That's a great point. It's so that the validator doesn't try to substitute anything within the pandas dataframe object. 
We were getting all sorts of issues with Cerberus on a pandas df. As a first pass, removing non-string from the validator. 

Ultimately most numbers get converted to string within yaml

PR #1023, opened 2024-12-12
Add ignore validation flag while registering model & improve logging

VipulMascarenhas:
### Description

Mistral has released some of their models in the Mistral format where you get one consolidated.safetensors and a params.json , and there is no config.json file. We still want users to proceed with model registration since the current set up restricts user to proceed. 

This PR adds a input check `--ignore-model-artifact-check` that will not throw an error during registration when `config.json` is not in the model artifacts. 

Also, some improvements have been made:
- get_model_files() API will include `.safetensors` file names along with config.json. This ensures that the model file list is not empty when config.json is not present. This applies to files read both from Object Storage and HF repo.
- For HF models, telemetry name is set as HF model name.
- There was a validation added to check if model_type attribute from custom metadata matches the one in config.json. The metadata in service models don't have this field, hence any exception thrown was just ignored. We now log the details, and can update the code to throw an error when this field is added in all verified models.

In addition to this:
This PR adds additional logging statements in the CRUD operations for model, deployment, finetuning and evaluation operations in AI Quick Actions. In addition to this, the PR also covers the following:
- base_handler has been updated to log request id in the error logs.

### Jira
More details on testing are in the ticket:
[ODSC-65657](https://jira.oci.oraclecorp.com/browse/ODSC-65657)

### Testing

**Register via HuggingFace**

**Scenario 1: Model has config.json file, user can proceed with registration.**

```
ads aqua model register --model google/gemma-2-2b --os_path oci://<bucket_name>@<namespace>/prefix/ --download_from_hf True --compartment_id $PROJECT_COMPARTMENT_OCID --inference_container odsc-tgi-serving
```

**Scenario 2: Model has no config.json, but is supported by vLLM. Validation does not allow registration to proceed.**

```
ads aqua model register --model mistralai/Pixtral-12B-Base-2409 --os_path --os_path oci://<bucket_name>@<namespace>/prefix/ --download_from_hf True --compartment_id $PROJECT_COMPARTMENT_OCID  
```

**Scenario 3: Model has no config.json, but with `--ignore_model_artifact_check True` parameter can proceed with registration.**  
 ```
ads aqua model register --model mistralai/Pixtral-12B-Base-2409 --os_path --os_path oci://<bucket_name>@<namespace>/prefix/ --download_from_hf True --compartment_id $PROJECT_COMPARTMENT_OCID --inference_container odsc-vllm-serving --ignore_model_artifact_check True
```

**Register via Object Storage**

**Scenario 4: Model comes from OSS and the path is correct. Validation says that there is no config.json file.**

```
ads aqua model register --model mistralai/Pixtral-12B-Base-2409-from-OSS --os_path oci://<bucket_name>@<namespace>/prefix/ --download_from_hf False --compartment_id $PROJECT_COMPARTMENT_OCID --inference_container odsc-vllm-serving
```

**Scenario 5: Model comes from OSS and the path is correct, and we use the override flag. Registration is allowed to proceed.** 

 ```
ads aqua model register --model mistralai/Pixtral-12B-Base-2409-from-OSS --os_path oci://<bucket_name>@<namespace>/prefix/ --download_from_hf False --compartment_id $PROJECT_COMPARTMENT_OCID --inference_container odsc-vllm-serving --ignore_model_artifact_check True

### CODE REVIEW

File: ads/aqua/model/model.py [hunk 1]
```diff
@@ -1206,7 +1210,9 @@ def _validate_safetensor_format(
             except Exception as ex:
                 logger.error(
                     f"Exception occurred while loading config file from {import_model_details.os_path}"
-                    f"Exception message: {ex}"
+                )
+                logger.error(
```
mrDzurb, line 1214: Why do we need a second error? Why not:

```
error_reason = ex.reason if isinstance(ex, AquaFileNotFoundError) else str(ex)
logger.error(
                    f"Exception occurred while loading config file from {import_model_details.os_path} "
                    f"Details: {error_reason}"
                )
```

VipulMascarenhas: yes, combined this and updated logic to show error and warning accordingly.

darenr (approved):

mrDzurb (approved):

mayoor (approved):

mayoor (approved):

PR #1025, opened 2024-12-13
AutoMLx internal explainability mode

codeloop:
Add automlx internal explainability as a mode explainability mode. this PR adds an additional explainability mode “AUTOMLX”, which when used with automlx will use the internal implementation of the Automlx package.


ads/opctl/operator/lowcode/forecast/model/base_model.py
```diff
@@ -668,7 +686,50 @@ def explain_model(self):
         for s_id, data_i in self.datasets.get_data_by_series(
             include_horizon=False
         ).items():
-            if s_id in self.models:
+            if (
+                self.spec.model == SupportedModels.AutoMLX
```
prasankh, line 690: Can we remove reference of automlx from the BaseModel class?

ads/opctl/operator/lowcode/forecast/model/automlx.py
```diff
@@ -284,8 +285,11 @@ def _generate_report(self):
                 )
 
                 # Append the global explanation text and section to the "other_sections" list
+                if global_explanation_section:
+                    other_sections.append(global_explanation_section)
+
+                # Append the local explanation text and section to the "other_sections" list
                 other_sections = other_sections + [
-                    global_explanation_section,
```
ahosler, line 288: Why are we removing this?

ads/opctl/operator/lowcode/forecast/const.py
```diff
@@ -27,10 +27,12 @@ class SpeedAccuracyMode(str, metaclass=ExtendedEnumMeta):
     HIGH_ACCURACY = "HIGH_ACCURACY"
     BALANCED = "BALANCED"
     FAST_APPROXIMATE = "FAST_APPROXIMATE"
+    AUTOMLX = "AUTOMLX"
```
ahosler, line 30: Is automlx faster/slower or more/less accurate than the other methods?

ahosler (dismissed):

ads/opctl/operator/lowcode/forecast/model/automlx.py
codeloop: As the automlx currently does not support global explainability, this is being removed will be added as default when we have the support

ads/opctl/operator/lowcode/forecast/model/base_model.py
codeloop: moved to the subclass, still some condition checks are there to manage exceptions.

ads/opctl/operator/lowcode/forecast/const.py
codeloop: AutoMLx is faster compared to the permutation explainer, on accuracy don't have numbers yet, will check with labs team if they have anything on this.

tests/operators/forecast/test_errors.py
```diff
@@ -672,6 +672,7 @@ def test_arima_automlx_errors(operator_setup, model):
         yaml_i["spec"]["model_kwargs"] = {"model_list": "superfast"}
     if model == "automlx":
         yaml_i["spec"]["model_kwargs"] = {"time_budget": 1}
+        yaml_i["spec"]["explanations_accuracy_mode"] = "AUTOMLX"
```
ahosler, line 675: Does this actually generate explanations when time-budget=1?

codeloop: Yes, it does, the time budget is used for model building. Also verified the same from the local runs.

prasankh (approved):

ahosler (approved):

PR #1030, opened 2024-12-19
Adding inference container URI for edit model handler

kumar-shivam-ranjan:
This PR intends to update EDIT model handler for editing registered model config in AQUA.
The edit feature is available for registered unverified model.
This will allow user to edit the following config for models in AQUA after the model is registered:

1.  Deployment container
2. Deployment container URI for TEI (Text Embedding inference container)
3. Task information of the model
4. Finetuning config

```
curl --location --request PUT 'http://localhost:8888/aqua/model/ocid1.datasciencemodel.oc1.iad.xxxxx' \
--header 'Content-Type: application/json' \
--data '{
    "inference_container": "odsc-tei-serving",
    "inference_container_uri": "iad.ocir.io/ociodscdev/text-embeddings-inference:cpu-1.5",
    "task": "feature-extraction"
}'
```


ads/aqua/extension/model_handler.py
```diff
@@ -163,10 +164,13 @@ def put(self, id):
             raise HTTPError(400, Errors.NO_INPUT_DATA)
 
         inference_container = input_data.get("inference_container")
+        inference_container_uri = input_data.get("inference_container_uri")
         inference_containers = AquaModelApp.list_valid_inference_containers()
         if (
             inference_container is not None
             and inference_container not in inference_containers
+            and inference_container
```
VipulMascarenhas, line 172: To avoid hardcoding, would it be better to move `AQUA_TEI_CONTAINER_FAMILY` from `InferenceContainerTypeFamily` to a new enum `CustomInferenceContainerTypeFamily(str, metaclass=ExtendedEnumMeta)`? That way the final inference_containers list can be `inference_containers.append(CustomInferenceContainerTypeFamily.values())`.

ads/aqua/model/model.py
```diff
@@ -407,13 +411,40 @@ def edit_registered_model(self, id, inference_container, enable_finetuning, task
...
                 if inference_container:
-                    custom_metadata_list.add(
-                        key=ModelCustomMetadataFields.DEPLOYMENT_CONTAINER,
-                        value=inference_container,
-                        category=MetadataCustomCategory.OTHER,
-                        description="Deployment container mapping for SMC",
-                        replace=True,
-                    )
+                    if (
+                        inference_container
```
VipulMascarenhas, line 415: this can be replaced with `inference_container in CustomInferenceContainerTypeFamily.values()`

ads/aqua/model/model.py
```diff
@@ -407,13 +411,40 @@ def edit_registered_model(self, id, inference_container, enable_finetuning, task
...
+                        custom_metadata_list.add(
+                            key=ModelCustomMetadataFields.DEPLOYMENT_CONTAINER_URI,
+                            value=inference_container_uri,
+                            category=MetadataCustomCategory.OTHER,
+                            description=f"Inference container URI for {ds_model.display_name}",
+                            replace=True,
+                        )
+                    else:
+                        raise AquaRuntimeError(
+                            f"Failed to edit model:{id}. Inference container URI can be edited only with TEI container."
```
VipulMascarenhas, line 445: f"Inference container URI can be edited only with container values: {CustomInferenceContainerTypeFamily.values()}".

ads/aqua/extension/model_handler.py
```diff
@@ -163,7 +166,9 @@ def put(self, id):
             raise HTTPError(400, Errors.NO_INPUT_DATA)
 
         inference_container = input_data.get("inference_container")
+        inference_container_uri = input_data.get("inference_container_uri")
         inference_containers = AquaModelApp.list_valid_inference_containers()
```
mrDzurb, line 170: Will it always return a list?

ads/aqua/model/model.py
```diff
@@ -377,7 +378,9 @@ def delete_model(self, model_id):
             )
 
     @telemetry(entry_point="plugin=model&action=delete", name="aqua")
```
mrDzurb, line 380: Why in telemetry we say delete?

ads/aqua/model/model.py
```diff
@@ -386,6 +389,8 @@ def edit_registered_model(self, id, inference_container, enable_finetuning, task
             The model OCID.
         inference_container: str.
             The inference container family name
+        inference_container_uri: str
+            The inference container uri for embedding models
```
mrDzurb, line 393: s it only for embedding models?

ads/aqua/model/model.py
```diff
@@ -401,19 +406,46 @@ def edit_registered_model(self, id, inference_container, enable_finetuning, task
...
-                    custom_metadata_list.add(
-                        key=ModelCustomMetadataFields.DEPLOYMENT_CONTAINER,
-                        value=inference_container,
-                        category=MetadataCustomCategory.OTHER,
-                        description="Deployment container mapping for SMC",
-                        replace=True,
-                    )
+                    if (
+                        inference_container
+                        in CustomInferenceContainerTypeFamily.values()
```
mrDzurb, line 417: ExtendedEnumMeta implement the ```__contains__``` method, that means we can do something like this:
```
inference_container in CustomInferenceContainerTypeFamily
```

ads/aqua/model/model.py
```diff
@@ -401,19 +406,46 @@ def edit_registered_model(self, id, inference_container, enable_finetuning, task
...
+                        and inference_container_uri is None
+                    ):
+                        raise AquaRuntimeError(
+                            "Inference container URI must be provided."
+                        )
+                    else:
+                        custom_metadata_list.add(
```
mrDzurb, line 424: Do we really need to store this info in metadata? Can't it be extracted from MD?

ads/aqua/extension/model_handler.py
kumar-shivam-ranjan: yes , either an empty list or list of valid containers from container_index.json stored in OSS bucket

ads/aqua/model/model.py
kumar-shivam-ranjan: Updated. Thanks

ads/aqua/model/model.py
kumar-shivam-ranjan: Yes as per now , we accept inference container URI for embedding models only

ads/aqua/model/model.py
kumar-shivam-ranjan: Updated

ads/aqua/model/model.py
kumar-shivam-ranjan: This metadata contains deployment container.  Via edit feature , user can edit/change deployment container as well. Hence , adding updated container to metadata.

mrDzurb: However user can update deployment from console, and in this case the container stored on the model level will not match a real one?

kumar-shivam-ranjan: yes but source of truth here should be model right , not the model deployment.
User can create multiple model deployments from a single model.  What we are trying  to do here is changing/updating the model config.
What if user registered a model for a particular deployment container.  Now , they due to some reason want to change the deployment container config of the model.

mrDzurb (approved):

VipulMascarenhas (approved):

PR #1031, opened 2024-12-20
Add AutoGen logger for generating report and integrating with OCI monitoring.

qiuosier:
This PR adds two AutoGen loggers:
```
from ads.llm.autogen.v02.loggers import MetricLogger, SessionLogger
```
* The ``SessionLogger`` saves events to a log file and generates report to for users to profile and debug the application.
* The ``MetricLogger`` sends the metrics to OCI monitoring service, allowing users to build dashboards to gain more insights about the application usage.

Screenshot 2024-12-20 at 2 27 56 PM
Screenshot 2024-12-20 at 2 28 26 PM
Screenshot 2024-12-20 at 2 30 37 PM


ads/llm/autogen/reports/data.py
```diff
@@ -0,0 +1,103 @@
...
+    timestamp: str
+    event_name: str
+    source_id: Optional[int] = None
+    source_name: Optional[str] = None
+    # Structured data for specific type of logs
+    data: Optional[LogData] = None
+    # Freeform data
+    kwargs: dict = field(default_factory=dict)
+
+    def to_dict(self):
```
darenr, line 41: Isn't it better to use the built in protocol for conversion to dict? `__dict__`

darenr (approved):

ads/llm/autogen/reports/data.py
qiuosier: Because the `data` attribute is another data class, the build in `__dict__` does not seem to convert it to a `dict`, while the `asdict()` function will convert it to `dict` recursively.

PR #1037, opened 2025-01-09
Set defaults for optional parameters in ChatOCIModelDeploymentVLLM

qiuosier:
The `ChatOCIModelDeploymentVLLM` class contains inference parameters for vLLM APIs. However, updates on vLLM may remove some parameters, causing breaking change even if users do not use those parameters.

For example, the `early_stopping` parameter exist for [v0.6.2](https://docs.vllm.ai/en/v0.6.2/dev/sampling_params.html) but it is removed since [v0.6.3](https://docs.vllm.ai/en/v0.6.3/dev/sampling_params.html).

In the `ChatOCIModelDeploymentVLLM` class there is already [logic](https://github.com/oracle/accelerated-data-science/blob/ac40648536dcbeff4ca304fa83925c62219af2fc/ads/llm/langchain/plugins/chat_models/oci_data_science.py#L1005) to include parameters in the API request only if they are not `None`. However, since some of the parameters has a default value instead of `None`, they are added to the API request even though users did not intend to use them.

This PR sets the defaults of the optional parameters to None, so that they will not be added to the API request unless users specified them explicitly.

### DIFF
--- a/ads/llm/langchain/plugins/chat_models/oci_data_science.py
+++ b/ads/llm/langchain/plugins/chat_models/oci_data_science.py
@@ -769,6 +769,8 @@ class ChatOCIModelDeploymentVLLM(ChatOCIModelDeployment):
     Science Model Deployment endpoint. See:
     https://docs.oracle.com/en-us/iaas/data-science/using/model-dep-policies-auth.htm#model_dep_policies_auth__predict-endpoint
 
+    See https://docs.vllm.ai/en/latest/api/inference_params.html for the defaults of the parameters.
+
     Example:
 
         .. code-block:: python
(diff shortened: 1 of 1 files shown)


VipulMascarenhas (approved):

VipulMascarenhas:
@qiuosier I thought `"autogen-agentchat~=0.2"` would restrict it to 0.2.x, but seems it install 0.4.0 version for python 3.10+ but not for 3.9. The autogen and tf2onnx tests should be fixed if we change it to "autogen-agentchat<0.4".

qiuosier:
Thanks @VipulMascarenhas . I created a new [PR](https://github.com/oracle/accelerated-data-science/pull/1038) for the autogen issue. @mrDzurb please suggest whether to merge this PR since your are testing the tool calling vllm image.

mrDzurb (approved):

PR #1050, opened 2025-01-31
Switch cache delete default

VipulMascarenhas:
### Description

This is a follow up on the PR#1044 to delete cache folder from the local directly in a notebook session.

For cache delete option, we want the user to set this flag explicitly when using CLI to register models. Deleting without any user input may result in bad experience for the user, in case they would like to keep the artifacts for further user.

In the UI, we'll have a checkbox that will always be selected by default for deleting cache, but user may uncheck it if needed. User will be able to see that the cache will be deleted, so they can choose to opt out of cache deletion if needed.

### DIFF
--- a/ads/aqua/extension/model_handler.py
+++ b/ads/aqua/extension/model_handler.py
@@ -133,7 +133,7 @@ class AquaModelHandler(AquaAPIhandler):
         )
         local_dir = input_data.get("local_dir")
         cleanup_model_cache = (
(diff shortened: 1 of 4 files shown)


mrDzurb (dismissed):

mayoor (approved):

mayoor (dismissed):

kumar-shivam-ranjan (approved):
LGTM

ahosler (approved):

PR #1051, opened 2025-02-02
Refactor: replace direct use of ExtendedEnumMeta with ExtendedEnum class

mrDzurb:
## Summary

Previously, every enum-like class had metaclass=ExtendedEnumMeta applied directly. This PR introduces a unified ExtendedEnum base class (which itself uses ExtendedEnumMeta) and updates all references to inherit from ExtendedEnum instead.

## Benefits

* Reduces repetition by centralizing the metaclass logic in ExtendedEnum.
* All enum-like classes now share the same structure and behaviors.
* Future updates to the metaclass functionality only need to happen in one place, making it easier to maintain and extend.

## Changes

* Added a new ExtendedEnum base class with metaclass=ExtendedEnumMeta.
* Updated all existing classes using metaclass=ExtendedEnumMeta to inherit from ExtendedEnum instead.
* Removed direct references to metaclass=ExtendedEnumMeta on child classes.

If any additional specialized behavior is needed, it can be added in the ExtendedEnum class.


VipulMascarenhas (approved):
lgtm 👍

qiuosier (approved):

PR #1052, opened 2025-02-03
Adding API to fetch tokenizer config for model

kumar-shivam-ranjan:
This PR intends to add additional model API to fetch tokenizer config of the model. It reads the tokenizer_config.json file from the model artifact location and returns the whole json as response body. The tokenizer config of the model contains chat template of the model which is useful when we deploy the model with chat completions endpoint.

Chat templates specify how to convert conversations, represented as lists of messages, into a single tokenizable string in the format that the model expects.

For models which don't have chat_templates, it will return empty json.

```
Request

curl --location 'http://localhost:8888/aqua/model/ocid1.datasciencemodel.oc1.iad.xxxx/tokenizer'

Response

{
    "bos_token": "<s>",
    "clean_up_tokenization_spaces": true,
    "cls_token": "<s>",
    "eos_token": "</s>",
    "mask_token": {
        "__type": "AddedToken",
        "content": "<mask>",
        "lstrip": true,
        "normalized": true,
        "rstrip": false,
        "single_word": false
    },
    "model_max_length": 8192,
    "pad_token": "<pad>",
    "sep_token": "</s>",
    "sp_model_kwargs": {},
    "tokenizer_class": "XLMRobertaTokenizer",
    "unk_token": "<unk>"
}
```

## Unit tests
[
Screenshot 2025-02-05 at 12 47 47 AM
](url)


ads/aqua/app.py
```diff
@@ -328,6 +328,55 @@ def get_config(self, model_id: str, config_file_name: str) -> Dict:
 
         return config
 
+    def get_chat_template(self, model_id):
```
VipulMascarenhas, line 331: can't we do similar to deployment and fine-tuning? 

```
AQUA_MODEL_TOKENIZER_CONFIG = "tokenizer_config.json"

def get_hf_tokenizer_config(self, model_id: str) -> Dict:
        config = self.get_config(model_id, AQUA_MODEL_TOKENIZER_CONFIG)
        if not config:
            logger.debug(
                f"Tokenizer config for model: {model_id} is not available. Use defaults."
            )
        return config
```

Relevant to this, for verified models, we want to load the config from the service model bucket instead of user bucket. PR- https://github.com/oracle/accelerated-data-science/pull/1053

If we use get_config directly, the above PR should take care of it.

VipulMascarenhas, line 331: add @telemetry decorator.

ads/aqua/extension/model_handler.py
```diff
@@ -316,8 +317,24 @@ def post(self, *args, **kwargs):  # noqa: ARG002
         )
 
 
+class AquaModelChatTemplateHandler(AquaAPIhandler):
```
VipulMascarenhas, line 320: it might be good to return the entire config instead of only the chat template field since there might be use cases down the line where other values might be needed from the tokenizer. UI can parse the json and get the chat_template.
We can name this AquaModelTokenizerConfigHandler accordingly.

VipulMascarenhas (changes requested):
added some comments. Also, could you please add some unit tests as well?

ads/aqua/extension/model_handler.py
```diff
@@ -316,8 +317,24 @@ def post(self, *args, **kwargs):  # noqa: ARG002
...
+        ):
+            return self.finish(AquaModelApp().get_hf_tokenizer_config(model_id))
+        else:
+            raise HTTPError(400, f"The request {self.request.path} is invalid.")
+
+
 __handlers__ = [
     ("model/?([^/]*)", AquaModelHandler),
     ("model/?([^/]*)/license", AquaModelLicenseHandler),
+    ("model/?([^/]*)/chat_template", AquaModelTokenizerConfigHandler),
```
VipulMascarenhas, line 338: let's change the endpoint name to /tokenizer instead of /chat_template since we'll be returning the entire json file.

VipulMascarenhas (commented):
changes look good. requesting one update to the endpoint name to match what we're returning.

ads/aqua/extension/model_handler.py
kumar-shivam-ranjan: Done

VipulMascarenhas (approved):

ads/aqua/app.py
```diff
@@ -306,20 +314,22 @@ def get_config(self, model_id: str, config_file_name: str) -> Dict:
             )
             base_model = self.ds_client.get_model(base_model_ocid).data
             artifact_path = get_artifact_path(base_model.custom_metadata_list)
+            if config_folder == "artifact":
```
mrDzurb, line 317: It would be better to move strings to constants.

ads/aqua/app.py
```diff
@@ -268,7 +268,12 @@ def if_artifact_exist(self, model_id: str, **kwargs) -> bool:
                 logger.info(f"Artifact not found in model {model_id}.")
                 return False
 
-    def get_config(self, model_id: str, config_file_name: str) -> Dict:
+    def get_config(
+        self,
+        model_id: str,
+        config_file_name: str,
+        config_folder: Optional[str] = "config",
```
mrDzurb, line 275: What if to pass `None` as a `config_folder`?  I guess in the code we should add something like
```
config_folder = config_folder or DEFAULT_MODEL_CONFIG_FOLDER
```
Let's also move the strings to the constants. 

```
DEFAULT_MODEL_CONFIG_FOLDER = "config"
```

ads/aqua/app.py
```diff
@@ -277,6 +282,9 @@ def get_config(self, model_id: str, config_file_name: str) -> Dict:
             The OCID of the Aqua model.
         config_file_name: str
             name of the config file
+        config_folder: Optional[str]
```
mrDzurb, line 285: Let's add a default value to the docstring as well.

ads/aqua/app.py
mrDzurb, line 317: We might need to convert the `config_folder` to the `lowercase`?

ads/aqua/app.py
```diff
@@ -306,20 +314,22 @@ def get_config(self, model_id: str, config_file_name: str) -> Dict:
...
             artifact_path = get_artifact_path(oci_model.custom_metadata_list)
-
         if not artifact_path:
             logger.debug(
                 f"Failed to get artifact path from custom metadata for the model: {model_id}"
             )
             return config
 
-        config_path = f"{os.path.dirname(artifact_path)}/config/"
+        config_path = f"{os.path.dirname(artifact_path)}/{config_folder}/"
```
mrDzurb, line 328: What if the `config_folder` will have `/` at the end? It might be better to use `os.path.join`?

ads/aqua/extension/model_handler.py
```diff
@@ -316,8 +317,24 @@ def post(self, *args, **kwargs):  # noqa: ARG002
         )
 
 
+class AquaModelTokenizerConfigHandler(AquaAPIhandler):
+    def get(self, model_id):
+        url_parse = urlparse(self.request.path)
+        paths = url_parse.path.strip("/")
+        path_list = paths.split("/")
+        if (
+            len(path_list) == 4
```
mrDzurb, line 326: What is `4` and `3` in this code? :) Let's move them either to constants or add some meaningful comments.

mrDzurb (commented):
Hi @kumar-shivam-ranjan, could you also change the title and description for the PR, since the scope of the PR has been adjusted.

ads/aqua/extension/model_handler.py
kumar-shivam-ranjan: The request path here is:  _/aqua/models/ocid1.iad.ahdxxx/tokenizer_
path_list=['aqua','models','ocid1.iad.ahdxxx','tokenizer']

Added comments

ads/aqua/app.py
kumar-shivam-ranjan: config_folder is parameter that we pass to get_config folder.
The determines whether we need to search for config in artifact folder (like tokenizer_config.json) or config folder (like ft_config.json).
The value of the config folder will always be either "config" (default) or "artifact".

ads/aqua/app.py
kumar-shivam-ranjan: Updated

ads/aqua/app.py
kumar-shivam-ranjan: Updated

ads/aqua/app.py
kumar-shivam-ranjan: config_folder values comes from constant enums which is always lowercase. so probably no need to convert explicitly?

kumar-shivam-ranjan:
> Hi @kumar-shivam-ranjan, could you also change the title and description for the PR, since the scope of the PR has been adjusted.

Updated

ads/aqua/app.py
```diff
@@ -306,20 +315,22 @@ def get_config(self, model_id: str, config_file_name: str) -> Dict:
             )
             base_model = self.ds_client.get_model(base_model_ocid).data
             artifact_path = get_artifact_path(base_model.custom_metadata_list)
+            if config_folder == ConfigFolder.ARTIFACT:
```
VipulMascarenhas, line 318: shouldn't we move this check before the code where base_model is loaded? Otherwise, we make an unnecessary backend call only to overwrite the artifact path with the one from oci_model.
```
if config_folder == ConfigFolder.ARTIFACT:
    artifact_path = ...
else:
    base_model = ...
    artifact_path = ... 
```

ads/aqua/app.py
VipulMascarenhas: agree with DC, makes it easier to handle paths - something like
`os.path.join(os.path.dirname(artifact_path), config_folder)`?

ads/aqua/app.py
```diff
@@ -306,20 +315,22 @@ def get_config(self, model_id: str, config_file_name: str) -> Dict:
...
                 f"Failed to get artifact path from custom metadata for the model: {model_id}"
             )
             return config
 
-        config_path = f"{os.path.dirname(artifact_path)}/config/"
+        config_path = f"{os.path.dirname(artifact_path)}/{config_folder}/"
         if not is_path_exists(config_path):
-            config_path = f"{artifact_path.rstrip('/')}/config/"
-
+            config_path = f"{artifact_path.rstrip('/')}/{config_folder}/"
```
VipulMascarenhas, line 331: same here for using `os.path.join`

ads/aqua/app.py
```diff
@@ -277,12 +282,16 @@ def get_config(self, model_id: str, config_file_name: str) -> Dict:
             The OCID of the Aqua model.
         config_file_name: str
             name of the config file
+        config_folder: (str, optional):
```
VipulMascarenhas, line 285: can you also add in docstrings when ARTIFACT need to used?

VipulMascarenhas (commented):
minor comments, rest looks good.

ads/aqua/app.py
kumar-shivam-ranjan: Added

ads/aqua/app.py
kumar-shivam-ranjan: Updated

VipulMascarenhas (dismissed):
lgtm 👍

ads/aqua/app.py
mrDzurb: Partially true, however we allow to pass a custom ```config_folder``` to the function?

ads/aqua/extension/model_handler.py
```diff
@@ -316,8 +317,26 @@ def post(self, *args, **kwargs):  # noqa: ARG002
         )
 
 
+class AquaModelTokenizerConfigHandler(AquaAPIhandler):
+    def get(self, model_id):
+        url_parse = urlparse(self.request.path)
+        paths = url_parse.path.strip("/")
+        path_list = paths.split("/")
```
mrDzurb, line 324: NIT: Looks like we only use ```path_list``` further in this method, then probably can do
```
path_list = urlparse(self.request.path).path.strip("/").split("/")

```

ads/aqua/extension/model_handler.py
```diff
@@ -316,8 +317,26 @@ def post(self, *args, **kwargs):  # noqa: ARG002
         )
 
 
+class AquaModelTokenizerConfigHandler(AquaAPIhandler):
```
mrDzurb, line 320: Let's add the pydocs to the class?

```
Handles requests for retrieving the Hugging Face tokenizer configuration 
    of a specified model.
    
    Expected request format:
        GET /aqua/models/<model-ocid>/tokenizer
```

ads/aqua/extension/model_handler.py
```diff
@@ -316,8 +317,26 @@ def post(self, *args, **kwargs):  # noqa: ARG002
...
+        path_list = paths.split("/")
+        # Path should be /aqua/models/ocid1.iad.ahdxxx/tokenizer
+        # path_list=['aqua','models','<model-ocid>','tokenizer']
+        if (
+            len(path_list) == 4
+            and is_valid_ocid(path_list[2])
+            and path_list[3] == "tokenizer"
+        ):
+            return self.finish(AquaModelApp().get_hf_tokenizer_config(model_id))
+        else:
```
mrDzurb, line 333: NIT: looks like else is not required here. I think it would be more clear to do something like this:

```
 is_valid_path = (
    len(path_list) == 4
    and path_list[3] == "tokenizer"
    and is_valid_ocid(path_list[2])
)

if not is_valid_path:
       raise HTTPError(400, f"The request {self.request.path} is invalid.")

return self.finish(AquaModelApp().get_hf_tokenizer_config(model_id))
```

mrDzurb (dismissed):

ads/aqua/extension/model_handler.py
kumar-shivam-ranjan: Updated

ads/aqua/extension/model_handler.py
kumar-shivam-ranjan: Updated

mrDzurb (approved):

PR #1053, opened 2025-02-03
[ODSC-66853] Load base model config by default

VipulMascarenhas:
### Description

Currently we copy deployment_config.json and finetuning_config.json from service tenancy bucket into user bucket while registering the model. This causes inconsistency in future when new shape support is added or when we enable fine-tuning to verified models. User in this case is forced to delete and re-register the model so that the configurations get refreshed.

This PR ensures that :
- for verified models, we always refer to the service tenancy configuration so that when there is a new update, the existing registered models will pick it up. Because of this, we don't need to copy the config files in the user bucket anymore.
- for unverified models, user can create a config folder themselves and add it in artifact location. This will be a backdoor option if user wishes to customize their configuration for deployment and finetuning. If not added, then defaults from UI will be shown.

Note: this issue also comes up for freeform tags, that will be updated in a separate PR.

### Jira
https://jira.oci.oraclecorp.com/browse/ODSC-66853

### DIFF
--- a/ads/aqua/app.py
+++ b/ads/aqua/app.py
@@ -296,7 +296,21 @@ class AquaApp:
             raise AquaRuntimeError(f"Target model {oci_model.id} is not Aqua model.")
 
         config = {}
-        artifact_path = get_artifact_path(oci_model.custom_metadata_list)
+        # if the current model has a service model tag, then
+        if Tags.AQUA_SERVICE_MODEL_TAG in oci_model.freeform_tags:
+            base_model_ocid = oci_model.freeform_tags[Tags.AQUA_SERVICE_MODEL_TAG]
+            logger.info(
+                f"Base model found for the model: {oci_model.id}. "
+                f"Loading {config_file_name} for base model {base_model_ocid}."
+            )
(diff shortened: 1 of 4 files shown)


ads/aqua/app.py
```diff
@@ -296,33 +296,45 @@ def get_config(self, model_id: str, config_file_name: str) -> Dict:
...
                 config = load_config(
                     config_path,
                     config_file_name=config_file_name,
                 )
             except Exception:
+                logger.debug(
+                    f"Error loading the {config_file_name} at path {config_path}.",
+                    exc_info=True,
+                )
                 pass
```
qiuosier, line 332: pass is not need. Also, since the log level is debug, I think it is better to log the full traceback
```
logger. debug(traceback.format_exc())
```

qiuosier (approved):

ads/aqua/app.py
VipulMascarenhas: ack. updated this.

qiuosier (approved):

mrDzurb (approved):

PR #1060, opened 2025-02-05
ODSC 68320: Modify the AQUA API to Accept GPU Count as an Optional Input

elizjo:
- Updated the GET [aqua/deployments/<model_ocid>/params](https://github.com/oracle/accelerated-data-science/blob/main/ads/aqua/extension/deployment_handler.py#L302) endpoint to return vLLM Parameters specifically for a set number of GPUs
- accepts gpu_count as an optional parameter
- used the new AQUA Service Managed Model Configs with 'multi_model_deployment' key

**Request**
```GET aqua/deployments/<model_ocid>/params?instance_shape=<shape_name>&gpu_count=<gpu_count_value>```

**Output**
```
{
    "data": [
        "--max-model-len 4096"
    ]
} 
```
**Modified the test_get_deployment_default_params() unit test**
we test cases such as:
- gpu_count = 1, we return the vLLM params as specified by the gpu_count : 1 field in the AQUA Service Managed Model Configs with 'multi_model_deployment' key
- gpu_count = 2, we return the vLLM params as specified by the gpu_count : 2 field in the AQUA Service Managed Model Configs with 'multi_model_deployment' key
- gpu_count is None, we always return the single model vLLM params
- TGI is attached as the inference container for the model and gpu_count is specified (always returns data: [] or single model TGI params since TGI is not supported by multi model)


mrDzurb (approved):

darenr (commented):
Also really nice code.

VipulMascarenhas (approved):
lgtm 👍

PR #1068, opened 2025-02-12
ODSC-68333: Adds Config Validation for Multi-Model Deployment

elizjo:
Finished writing method validate_config and pytest unit tests for class CreateModelDeploymentDetails

- validate_config takes in a ModelDeploymentConfigSummary (bundles all deployment configs, gpu_count, total available gpus on a selected shape into one pydantic data class)

In a Multi-Model Deployment, the method validates the following:
- checks if deployment is a multi-model deployment
- assigned GPU allocations per model are within the number of GPUs available in the shape, instance_shape
- validate if all models in model group can be deployed on the selected shape, instance_shape

See initial testing of method in tests/unitary/with_extras/aqua/test_deployment.py for class TestCreateModelDeploymentDetails. See attached screenshot for successful tests.
Screenshot 2025-02-14 at 11 57 06 AM


ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +319,173 @@ class ModelDeploymentConfigSummary(Serializable):
...
+        Raises
+        -------
+        ValueError:
+            When the deployment is NOT a multi model deployment
+            When assigned GPU Allocations per model are NOT within the number of GPUs available in the instance shape
+            When all models in model group can NOT be deployed on the instance shape with the selected GPU count
+        """
+        if self.freeform_tags.get(Tags.MULTIMODEL_TYPE_TAG) == "true":
+            selected_shape = self.instance_shape
+            total_available_gpus = getattr(
```
mrDzurb, line 434: I think the ```getattr``` will throw an error, if ```models_config_summary.gpu_allocation.get(selected_shape)``` return ```None```?  Also I think we might not need to use getattr at all, we can do this instead:
```

if selected_shape not in models_config_summary.gpu_allocation:
     raise ValueError(...)

total_available_gpus = models_config_summary.gpu_allocation[selected_shape].total_gpus_available
```

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +319,173 @@ class ModelDeploymentConfigSummary(Serializable):
...
+                models_config_summary.gpu_allocation.get(selected_shape),
+                "total_gpus_available",
+                None,
+            )
+            models_allocated_gpus = getattr(
```
mrDzurb, line 439: See the above comment

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +319,173 @@ class ModelDeploymentConfigSummary(Serializable):
...
+                models_config_summary.gpu_allocation.get(selected_shape), "models", None
+            )
+
+            if not isinstance(total_available_gpus, int):
+                raise ValueError(
```
mrDzurb, line 444: The error message should be **clear and meaningful** for the user. We can include **debug information** in the logs, but the error shown to the user should be **understandable and actionable**.  

Since users are not aware of our **internal configurations**, the message should focus on what went wrong and how they can resolve it, rather than exposing internal details.

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +319,173 @@ class ModelDeploymentConfigSummary(Serializable):
...
+                    f"Missing total GPU allocation for the selected shape {selected_shape}"
+                )
+
+            if not all(
```
mrDzurb, line 448: I don't think we need to check this, we expect that pydantic will do this job for us.

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +319,173 @@ class ModelDeploymentConfigSummary(Serializable):
...
+        """Ensures exactly one of `model_id` or `models` is provided."""
+        model_id = values.get("model_id")
+        models = values.get("models")
+        if bool(model_id) == bool(models):  # Both set or both unset
+            raise ValueError(
+                "Exactly one of `model_id` or `models` must be provided to create a model deployment."
+            )
+        return values
+
+    def validate_config(self, models_config_summary: ModelDeploymentConfigSummary):
```
mrDzurb, line 415: NIT: maybe ```validate_multimodel_deployment_feasibility``` would be the better name for this validator?

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +319,173 @@ class ModelDeploymentConfigSummary(Serializable):
...
+        """
+        Validates the model configuration for multi model deployments.
```
mrDzurb, line 417: Let's add more description here, what exactly we are validating in this validator.

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +319,173 @@ class ModelDeploymentConfigSummary(Serializable):
...
+            ):
+                raise ValueError(
+                    "GPU allocations must be instances of GPUModelAllocation"
+                )
+
+            model_deployment_config = models_config_summary.deployment_config
+
+            sum_model_gpus = 0
+
+            for model in models_allocated_gpus:
```
mrDzurb, line 459: I think this logic needs some adjustments. I noticed that this **validator** does not utilize the information provided in **`self.models`**, even though this attribute contains the key details we need to validate.  

```python
class AquaMultiModelRef(Serializable):
    model_id: str
    model_name: Optional[str] = None
    gpu_count: Optional[int] = None
    env_var: Optional[dict] = None
```

Since this **config represents the user's input**, our validation should ensure that the specified **`model_id`** and **`gpu_count`** can be deployed within the selected **shape**. Right now, it seems like we are not leveraging this information in the validation process.

ads/aqua/modeldeployment/entities.py
elizjo: Fixed names in this file, deployment.py, and the test cases

ads/aqua/modeldeployment/entities.py
elizjo: Removed

ads/aqua/modeldeployment/entities.py
elizjo: Fixed validation logic to use models (List[AquaMultiModelRef]) in main loop, using the model_id to look up the configuration details per a specific model. Then we compare the available gpu_count entries in the multi model deployment section of the config to the user selected gpu_count in the List[AquaMultiModelRef]. 

We do not use the gpu_allocation field in the ModelDeploymentConfigSummary object, rather we use the variable models (List[AquaMultiModelRef]).

ads/aqua/modeldeployment/entities.py
elizjo: removed & replaced with suggested code snippet above.

ads/aqua/modeldeployment/entities.py
elizjo: Changed several error messages to start with an action that the user can initiate to fix the issue.

ads/aqua/modeldeployment/entities.py
```diff
@@ -382,6 +288,17 @@ class GPUShapeAllocation(Serializable):
     class Config:
         extra = "allow"
 
+class ConfigValidationError(Exception):
+    """Exception raised for config validation."""
+
+    def __init__(
+        self,
+        message: str = """Validation failed: The provided model group configuration is not compatible with the selected instance shape.
```
mrDzurb, line 296: NIT: 

```
The provided model group configuration is incompatible with the selected instance shape. Please verify the GPU count per model and ensure multi-model deployment is supported for the chosen instance shape.
```

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +330,190 @@ class ModelDeploymentConfigSummary(Serializable):
...
+        ConfigValidationError:
+            When the deployment is NOT a multi model deployment
+            When assigned GPU Allocations per model are NOT within the number of GPUs available in the instance shape
+            When all models in model group can NOT be deployed on the instance shape with the selected GPU count
+        """
+        if not self.models:
+            logger.error(
+                "Validation Failed: User defined model group is None."
+            )
+            raise ConfigValidationError("""Validation Failed: At least two models are required for multi-model deployment,
```
mrDzurb, line 456: I think we should be ok with one model in a group as well? If so, we can say something like:

```
Multi-model deployment requires at least one model, but none were provided. Please add one or more models to the model group to proceed.
```

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +330,190 @@ class ModelDeploymentConfigSummary(Serializable):
...
+                                        but only none were provided. Add 2 or more models in the model group to proceed.""")
+
+        selected_shape = self.instance_shape
+
+        if selected_shape not in models_config_summary.gpu_allocation:
+            logger.error(
+                    f"Validation Failed: The model group are not compatible with the selected instance shape {selected_shape}"
```
mrDzurb, line 463: NIT: ```model group are not```->```model group is not```?

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +330,190 @@ class ModelDeploymentConfigSummary(Serializable):
...
+                )
+            raise ConfigValidationError("Validation Failed: Select a different instance shape. The selected instance shape is not supported.")
```
mrDzurb, line 465: We don't need to add ```Validation Failed:``` in the beginning I think. 

We can rephrase: 
```
The model group is not compatible with the selected instance shape '{selected_shape}'. Select a different instance shape. 
```

elizjo: Should I remove `Validation Failed:` from all Error messages or just this one at line 465?

mrDzurb: Let's remove it from the all error messages.

mrDzurb (approved):

lu-ohai (approved):

ads/aqua/modeldeployment/entities.py
```diff
@@ -413,3 +330,189 @@ class ModelDeploymentConfigSummary(Serializable):
...
+        models = values.get("models")
+        if bool(model_id) == bool(models):  # Both set or both unset
+            raise ValueError(
+                "Exactly one of `model_id` or `models` must be provided to create a model deployment."
+            )
+        return values
+
+    def validate_multimodel_deployment_feasibility(self, models_config_summary: ModelDeploymentConfigSummary):
+        """
+        Validates whether the user input of a model group (List[AquaMultiModelRef], 2+ models with a specified gpu count per model)
```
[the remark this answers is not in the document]
elizjo: will fix on next PR, thanks!

PR #1085, opened 2025-02-25
ODSC 68580- Update Evaluation SDK to Support Multi-Model Deployment

elizjo:
The "name" parameter in the model parameters had to be updated from using the same default name for all deployments to the model's specific name.  

This PR introduces a method to validate the user's input when they specify the model name for targeting a single model for evaluation when the model is within a multi model deployment. 

Added method validate_name_multi_model():
- used in AquaEvaluationApp.create() in evaluation.py.
- assumes that the UI and evaluation handler sends back the user's input model parameters, which includes the "name" parameter with the key being the model specific name
- we check the DataScienceModel metadata for the model specific name (configuration setting) and whether this name matches the user's input.

Wrote test cases in test_deployment() in the test_validate_multi_model_evaluation() method.
- see Jira Ticket for screenshots of passed test cases.

### DIFF
--- a/ads/aqua/evaluation/evaluation.py
+++ b/ads/aqua/evaluation/evaluation.py
@@ -185,25 +185,22 @@ class AquaEvaluationApp(AquaApp):
             evaluation_source = ModelDeployment.from_id(
                 create_aqua_evaluation_details.evaluation_source_id
             )
-            try:
-                if Tags.MULTIMODEL_TYPE_TAG in evaluation_source.freeform_tags:
-                    multi_model_id = evaluation_source.freeform_tags.get(
-                        Tags.AQUA_MODEL_ID_TAG, UNKNOWN
+
+            if Tags.MULTIMODEL_TYPE_TAG in evaluation_source.freeform_tags:
+                multi_model_id = evaluation_source.freeform_tags.get(
+                    Tags.AQUA_MODEL_ID_TAG, UNKNOWN
+                )
+
+                if not multi_model_id:
+                    raise AquaRuntimeError(
+                        f"Invalid multi model deployment {multi_model_id}."
+                        f"Make sure the {Tags.AQUA_MODEL_ID_TAG} tag is added to the deployment."
                     )
 
-                    if not multi_model_id:
-                        raise AquaRuntimeError(
-                            f"Invalid multi model deployment {multi_model_id}."
-                            f"Make sure the {Tags.AQUA_MODEL_ID_TAG} tag is added to the deployment."
-                        )
-
-                    aqua_model = DataScienceModel.from_id(multi_model_id)
-                    AquaEvaluationApp.validate_model_name(
-                        aqua_model, create_aqua_evaluation_details
-                    )
-
-            except AquaError as err:
-                raise AquaValueError(f"{err}") from err
+                aqua_model = DataScienceModel.from_id(multi_model_id)
+                AquaEvaluationApp.validate_model_name(
+                    aqua_model, create_aqua_evaluation_details
+                )
 
             try:
                 if (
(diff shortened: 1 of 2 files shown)


ads/aqua/evaluation/evaluation.py
```diff
@@ -183,6 +184,26 @@ def create(
...
+                            f"Invalid multi model deployment {multi_model_id}."
+                            f"Make sure the {Tags.AQUA_MODEL_ID_TAG} tag is added to the deployment."
+                        )
+
+                    aqua_model = DataScienceModel.from_id(multi_model_id)
+                    AquaEvaluationApp.validate_name_multi_model(
+                        aqua_model, create_aqua_evaluation_details
+                    )
+
+            except (AquaRuntimeError, AquaValueError) as err:
```
mrDzurb, line 204: I think we don't need to add ```AquaValueError``` into except section. Ot will work by itself.

ads/aqua/evaluation/evaluation.py
```diff
@@ -550,6 +571,43 @@ def create(
             parameters=AquaEvalParams(),
         )
 
+    @staticmethod
+    def validate_name_multi_model(
```
mrDzurb, line 575: Let name it - ```validate_model_name```? Also please add description for this method.

ads/aqua/evaluation/evaluation.py
```diff
@@ -550,6 +571,43 @@ def create(
...
+        evaluation_source: DataScienceModel,
+        create_aqua_evaluation_details: CreateAquaEvaluationDetails,
+    ):
```
mrDzurb, line 578: NIT: ```->None```?

ads/aqua/evaluation/evaluation.py
```diff
@@ -550,6 +571,43 @@ def create(
...
+        user_model_parameters = create_aqua_evaluation_details.model_parameters
+        if "model" not in user_model_parameters:
+            logger.debug(
+                f"User did not input model name for multi model deployment evaluation with evaluation source ID: {create_aqua_evaluation_details.evaluation_source_id}"
+            )
+            raise AquaValueError(
+                "Provide the model name. For evaluation, a single model needs to be targeted using the name in the multi model deployment."
```
mrDzurb, line 585: I think we can show the valid names in this message.

```
Provide the model name. For evaluation, a single model needs to be targeted using the name in the multi model deployment. The valid model names for this Model Deployment are {valid_model_names}.
```

tests/unitary/with_extras/aqua/test_evaluation.py
```diff
@@ -449,7 +451,7 @@ def test_create_evaluation(
         mock_from_id.return_value = foundation_model
 
         experiment = MagicMock()
-        experiment.id = "test_experiment_id"
+        experiment.id = "ocid1.datasciencemodelversionset.oc1.iad.amaaaaaav66vvniakngdzelb5hcgjd6yvfejksu2excidvvi3s5s5whtmdea"
```
mrDzurb, line 454: Let's not use real IDs?

darenr:
If someone specifies a model name (correctly) in a single model deployment case is this valid, or does that require the old/current "odsc_model" (or whatever it is)

ads/aqua/evaluation/evaluation.py
```diff
@@ -550,6 +572,63 @@ def create(
...
+        -------
+        AquaValueError:
+            - When the user fails to specify any input for the model name.
+            - When the user supplies a model name that does not match the model name set in the DataScienceModel metadata."""
+        user_model_parameters = create_aqua_evaluation_details.model_parameters
+
+        custom_metadata_list = evaluation_source.custom_metadata_list
+        user_model_name = user_model_parameters.get("model")
+
+        model_group_count = int(
```
mrDzurb, line 602: Wouldn't this fail if **`ModelCustomMetadataFields.MULTIMODEL_GROUP_COUNT`** is missing from the custom metadata?

elizjo:
@darenr we would only allow the user to enter the name if the Model Deployment selected is a multi model deployment (we use a freeform tag to ID these deployments). 

Otherwise, the user will not be able to enter the name and we would be keeping the odsc_llm (default name for MD).

Hope this clarifies your question!

ads/aqua/evaluation/evaluation.py
elizjo: I see. I will remove this and the try block. The evaluation_handler.py post method (that calls .create() method here) has the @handle_exceptions decorator, which has the `except AquaError as error`. I believe that the decorator will catch the exception, so this code is redundant here.

ads/aqua/evaluation/evaluation.py
elizjo: done

ads/aqua/evaluation/evaluation.py
elizjo: done

tests/unitary/with_extras/aqua/test_evaluation.py
elizjo: done

ads/aqua/evaluation/evaluation.py
elizjo: corrected

ads/aqua/evaluation/evaluation.py
```diff
@@ -550,6 +572,63 @@ def create(
...
+                ModelCustomMetadataFields.MULTIMODEL_GROUP_COUNT
+            ).value
+        )
+
+        model_names = [
+            custom_metadata_list.get(f"model-name-{idx}").value
+            for idx in range(model_group_count)
+        ]
+
+        valid_model_names = ", ".join(map(str, model_names))
```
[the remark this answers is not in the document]
elizjo: Yes- I changed it now to not include None values

mrDzurb (approved):

PR #1086, opened 2025-02-26
Added Failure Indicator & Logging Enhancements

prasankh:
[ODSC-68351](https://jira.oci.oraclecorp.com/browse/ODSC-68351)
For Real Time What-If Analysis 
- If model deployment fails, the life cycle state of the model deployment is now added as status to indicate failure / success.
- log_Id is now created at runtime given log_group
- OCIDs for logs are added in the output JSON file to ease debugging.


ads/opctl/operator/common/utils.py
```diff
@@ -190,3 +192,42 @@ def print_traceback():
...
+    """
+    Creates a log within a given log group
+    :param signer: The signer
+    :param log_group_id: The OCID of the log group.
+    :param log_name: Name of the log to be created.
+    :return: Created log id.
+    """
+
+    if not log_name:
+        log_name = f"log-{int(time.time())}-{uuid.uuid4()}"
```
ahosler, line 207: Does it make sense to do utc time? Then whether the cloud region is phoenix or ash urn the outputs are independent

ads/opctl/operator/common/utils.py
```diff
@@ -190,3 +192,42 @@ def print_traceback():
...
+    logging_client = oci.logging.LoggingManagementClient(config={}, signer=signer)
```
ahosler, line 208: Probably too late now, but did you look at the ads wrapper for logging? Maybe it wouldn't work here anyways

ads/opctl/operator/lowcode/forecast/whatifserve/deployment_manager.py
```diff
@@ -184,6 +185,9 @@ def create_deployment(self):
 
         log_group = self.spec.what_if_analysis.model_deployment.log_group
         log_id = self.spec.what_if_analysis.model_deployment.log_id
+        if not log_id and not self.test_mode:
+            auth = oci.auth.signers.get_resource_principals_signer()
```
ahosler, line 189: This will break the local operator from creating a what if analysis. 
That's ok, but we just need to note that somehow in the docs

ahosler (changes requested):

ads/opctl/operator/common/utils.py
prasankh: The only reason to use timestamp is to keep the name unique and it returns epoch time which is in UTC

ads/opctl/operator/common/utils.py
prasankh: Updated code to use ads logging API for creating log.

ahosler (approved):

codeloop (approved):

PR #1092, opened 2025-03-02
Refactor Container Index JSON Loader: Move to Config Package & Switch to Pydantic

mrDzurb:
## Description
This PR enhances the container index JSON loader by moving it from the UI module to the config package and refactoring it to use Pydantic instead of dataclasses. 

### Key Changes
 - Moved the container index JSON loader to the config package for better organization.
 - Refactored the implementation from dataclasses to Pydantic, enabling stricter validation, better serialization, and extensibility.
 - Updated unit tests to reflect the new implementation while maintaining existing functionality.


mayoor (approved):

ads/aqua/extension/model_handler.py
```diff
@@ -45,7 +38,7 @@ def get(
                     400, Errors.MISSING_REQUIRED_PARAMETER.format("model_format")
                 )
             try:
-                model_format = ModelFormat(model_format.upper())
+                model_format = model_format.upper()
```
qiuosier, line 41: Seems like this code is no longer checking if the model format is valid? In that case, I guess there is no need to have try/except for `upper()`?

ads/aqua/config/container_config.py
```diff
@@ -0,0 +1,197 @@
...
+        Parameters
+        ----------
+        config (Optional[Dict]): The container index JSON.
+        enable_spec (Optional[bool]): If True, fetch container specification details.
+
+        Returns
+        -------
+        AquaContainerConfig: The constructed container configuration.
+        """
+        # if not config:
```
qiuosier, line 134: Do we need this? I see you are adding `config=get_container_config()` in every call instead. So the `config` should not be `None` anymore. Or maybe having a default is better?

qiuosier (dismissed):

kumar-shivam-ranjan (approved):

qiuosier (approved):

mayoor (approved):

PR #1093, opened 2025-03-04
Fix date format in cross-validation for auto-select

prasankh:
Using Pandas' default datetime format in cross validation now


ads/opctl/operator/lowcode/forecast/model_evaluator.py
```diff
@@ -100,6 +100,7 @@ def create_operator_config(self, operator_config, backtest, model, historical_da
         backtest_op_config_draft = operator_config.to_dict()
         backtest_spec = backtest_op_config_draft["spec"]
         backtest_spec["historical_data"]["url"] = historical_data_url
+        backtest_spec["datetime_column"]["format"] = None
```
ahosler, line 103: Should we not use the format from the user provided yaml file when provided?

ahosler (changes requested):

ads/opctl/operator/lowcode/forecast/model_evaluator.py
prasankh: Not needed in cross-validation.

ahosler (approved):

codeloop (approved):

ahosler (approved):

PR #1095, opened 2025-03-04
Support deploy multi custom models

lu-ohai:
### Support deploy multi custom models

- Updated api `get_multimodel_deployment_config` and `validate_multimodel_deployment_feasibility` to handle custom model without deployment config.

### Notebook
- Three custom models without deployment config
Screenshot 2025-03-06 at 3 15 08 PM

- Two custom models without deployment config and one service model with deployment config
Screenshot 2025-03-06 at 4 54 00 PM
Screenshot 2025-03-06 at 4 54 12 PM

- Three service models with deployment config
Screenshot 2025-03-06 at 4 55 39 PM
Screenshot 2025-03-06 at 4 55 57 PM
Screenshot 2025-03-06 at 4 56 10 PM


mrDzurb (approved):

ads/aqua/modeldeployment/entities.py
```diff
@@ -550,6 +553,15 @@ def validate_multimodel_deployment_feasibility(
             sum_model_gpus += model.gpu_count
             aqua_deployment_config = model_deployment_config[model.model_id]
 
+            source_model = DataScienceModel.from_id(model.model_id)
```
mrDzurb, line 556: I think we don't really need such validation. If we treat the all models equally. 

Wouldn't this be just enough? 
```
# Skips validating models without deployment config as the configuration is empty and gpu counts are 

generated by ads sdk
     if not aqua_deployment_config.configuration:
          continue
```

mrDzurb (approved):

ads/aqua/modeldeployment/entities.py
lu-ohai: Updated

PR #1103, opened 2025-03-08
Switch Multi-Model deployment container validation to usage-based tagging

mrDzurb:
# Description
This update modifies the **multi-model deployment validation** by replacing the **static container family check** with a **usage-based approach**. Instead of relying on a hardcoded container family, the validation now leverages the **`usages`** attribute, which includes a **`multi_model`** tag for containers that support multi-model deployment.

### DIFF
--- a/ads/aqua/config/container_config.py
+++ b/ads/aqua/config/container_config.py
@@ -4,10 +4,17 @@
 
 from typing import Dict, List, Optional
 
+from common.extended_enum import ExtendedEnum
 from pydantic import Field
 
 from ads.aqua.common.entities import ContainerSpec
 from ads.aqua.config.utils.serializer import Serializable
+
+
+class Usage(ExtendedEnum):
+    INFERENCE = "inference"
+    BATCH_INFERENCE = "batch_inference"
+    MULTI_MODEL = "multi_model"
 
 
 class AquaContainerConfigSpec(Serializable):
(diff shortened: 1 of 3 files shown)


ads/aqua/config/container_config.py
```diff
@@ -4,12 +4,19 @@
...
 from typing import Dict, List, Optional
 
+from common.extended_enum import ExtendedEnum
 from pydantic import Field
 
 from ads.aqua.common.entities import ContainerSpec
 from ads.aqua.config.utils.serializer import Serializable
 
 
+class Usage(ExtendedEnum):
```
darenr, line 14: Could you use auto here? 

class Usage(StrEnum):
    """
    Enumeration representing different usage modes.
    """
    INFERENCE = auto()
    BATCH_INFERENCE = auto()
    MULTI_MODEL = auto()

darenr (approved):
I made a suggestion that from memory should work to avoid the name of the enum having to be set explicitly, but it's worth verifying should you chose to adopt it

ads/aqua/modeldeployment/deployment.py
```diff
@@ -196,11 +196,23 @@ def create(
...
-                InferenceContainerTypeFamily.AQUA_VLLM_CONTAINER_FAMILY
+                container_config_item.family
+                for container_config_item in service_inference_containers
+                if Usage.MULTI_MODEL in container_config_item.usages
             ]
 
+            if not supported_container_families:
+                raise AquaValueError(
+                    "Currently, there are no containers that support multi-model deployment."
+                )
```
elizjo, line 214: same concern as above

elizjo (approved):

PR #1109, opened 2025-03-12
Add HttpxOCIAuth Integration Documentation for HTTPX Clients

mrDzurb:
## Overview

This PR updates the documentation to include a new section for the `HttpxOCIAuth` class.
The HttpxOCIAuth class enables HTTPX clients to sign requests using OCI signers, making them compatible
with OCI Model Deployment services and LLM models. This integration facilitates the use of OCI authentication
with HTTPX-based clients and extends compatibility to third-party libraries such as the OpenAi client.

### Changes

- Added a new documentation section titled **HTTPX Client Integration with OCI Authentication (HttpxOCIAuth)**
- Included code examples for both synchronous and asynchronous HTTPX client usage
- Enhanced the `ads.set_auth()` with default `client_kwargs = {"service_endpoint": OCI_ODSC_SERVICE_ENDPOINT}`


ads/aqua/client/client.py
```diff
@@ -797,3 +794,49 @@ async def embeddings(
...
+    Creates and returns a synchronous httpx Client configured with OCI authentication.
+
+    This function checks if an 'auth' keyword argument is provided. If not, it instantiates
+    the default OCI authentication (HttpxOCIAuth) and injects it into the client configuration.
+    Any additional keyword arguments are passed directly to the httpx.Client constructor.
+
+    Parameters
+    ----------
+    **kwargs : Any
+        Arbitrary keyword arguments for configuring the httpx.Client. An optional 'auth'
```
mayoor, line 810: Keyword arguments supported by httpx.Client.

ads/aqua/client/client.py
```diff
@@ -797,3 +794,49 @@ async def embeddings(
         logger.debug(f"Generating embeddings with input: {input}, payload: {payload}")
         payload = {**(payload or {}), "input": input}
         return await self._request(payload=payload, headers=headers)
+
+
+def get_httpx_client(**kwargs: Any) -> httpx.Client:
+    """
+    Creates and returns a synchronous httpx Client configured with OCI authentication.
```
mayoor, line 801: Creates and returns a synchronous httpx Client configured with OCI authentication signer based the authentication type setup using `ads.set_auth` method or env variable `OCI_IAM_TYPE`. More information - https://accelerated-data-science.readthedocs.io/en/stable/user_guide/cli/authentication.html

ads/aqua/client/client.py
mrDzurb: done

ads/aqua/client/client.py
mrDzurb: done

mayoor (dismissed):

lu-ohai (dismissed):

qiuosier:
I think `OCI_ODSC_SERVICE_ENDPOINT` can only be used by `DataScienceClient`. Similarly, different client will have different service endpoint (if the users choose to set it). However, seems like this PR is setting the same service endpoint for all clients, which may cause errors. There will be no errors unit tests as it is not actually connected to the endpoint.

qiuosier (approved):

PR #1110, opened 2025-03-12
Optimize auto-select data loading process and update dataset classes

prasankh:
[ODSC-64831](https://jira.oci.oraclecorp.com/browse/ODSC-64831)

Changes:
- Optimized the auto-select feature to avoid redundant data resaving during cross validations.
- Updated dataset classes to support loading data either from the config path or by directly passing dataframes.


ahosler (approved):

ahosler:
This is great! Tied up a lot of loose ends and cleans up the code base.

codeloop (approved):

PR #1121, opened 2025-03-16
Remove service-defined tags when updating the Model Catalog record

mrDzurb:
## Description
Remove service-defined tags when updating the Model Catalog record.

### DIFF
--- a/ads/model/common/utils.py
+++ b/ads/model/common/utils.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+# Copyright (c) 2022, 2025 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 
 import json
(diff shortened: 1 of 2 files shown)


qiuosier:
I think this problem is a bit more complicated than just popping the "oracle-tags". If I run the following code
```
from ads.model import DataScienceModel

model = (
    DataScienceModel(
        compartment_id="<>",
        project_id="<>"
    )
    .with_display_name("Test Tags")
    .create()
)
model_id = model.id
```
and a model is created with `oracle-tags`.

Then, if run the following line:
```
model.update(display_name="Test Tags Updated")
```
the `oracle-tags` will actually be removed, which not expected.

mrDzurb:
> I think this problem is a bit more complicated than just popping the "oracle-tags". If I run the following code
> 
> ```
> from ads.model import DataScienceModel
> 
> model = (
>     DataScienceModel(
>         compartment_id="<>",
>         project_id="<>"
>     )
>     .with_display_name("Test Tags")
>     .create()
> )
> model_id = model.id
> ```
> 
> and a model is created with `oracle-tags`.
> 
> Then, if run the following line:
> 
> ```
> model.update(display_name="Test Tags Updated")
> ```
> 
> Without this PR, the `oracle-tags` remains in the model. With this PR, the `oracle-tags` will actually be removed, which not expected.

Looks like I forgot to commit an extra line. The idea was that if there are **no defined tags** apart from the **service-provided ones**, we simply set them to **None**.

PR #1122, opened 2025-03-17
defined tag support while creating model

mayoor:


mrDzurb (approved):

qiuosier (approved):
I was able to reproduce the error and tested this code.

PR #1129, opened 2025-03-21
[DRCC] Create and Update Metadata artifact content changes

kumar-shivam-ranjan:
# Description
This PR is intended to improve API for creating and updating metadata artifact which was introduced as part of MS enhancements.

# Issue
While uploading metadata artifact , the type of content should always be bytes.  Earlier it was both bytes and strings.

# Example

When uploading the content via local path

```
from ads.model.common.utils import MetadataArtifactPathType
from ads.model.datascience_model import DataScienceModel

ds_model=DataScienceModel.from_id("ocid1.datasciencemodel.oc1.iad.xxyz")

ds_model.create_defined_metadata_artifact(
    "README",
    artifact_path_or_content="/Users/<username>/Downloads/README.md",
    path_type=MetadataArtifactPathType.LOCAL
)
```
When uploading the content via OSS path

```
from ads.model.common.utils import MetadataArtifactPathType
from ads.model.datascience_model import DataScienceModel

ds_model=DataScienceModel.from_id("ocid1.datasciencemodel.oc1.iad.xxyz")

ds_model.create_defined_metadata_artifact(
    "README",
    artifact_path_or_content="oci://path/to/oss/bucket/README.md",
    path_type=MetadataArtifactPathType.OSS
)
```

When uploading the content itself
```
from ads.model.common.utils import MetadataArtifactPathType
from ads.model.datascience_model import DataScienceModel

ds_model=DataScienceModel.from_id(
    "ocid1.datasciencemodel.oc1.iad.amaaaaaav66vvniavg7xmekk2igrj6oybxgswetlmztgaqrz274jryp4hf7a"
)
with open("/Users/kumarran/Downloads/README.md","rb") as f:
    content=f.read()

# type of content here is bytes
ds_model.create_defined_metadata_artifact(
    "README",
    artifact_path_or_content=content,
    path_type=MetadataArtifactPathType.CONTENT
)
```

# Unit tests
Screenshot 2025-03-21 at 5 37 51 PM

### DIFF
--- a/ads/model/datascience_model.py
+++ b/ads/model/datascience_model.py
@@ -89,6 +89,11 @@ class InvalidArtifactType(Exception):  # pragma: no cover
     pass
 
 
+class InvalidArtifactPathTypeOrContentError(Exception):  # pragma: no cover
+    def __init__(self, msg="Invalid type of Metdata artifact content"):
+        super().__init__(msg)
+
+
 class CustomerNotificationType(ExtendedEnum):
     NONE = "NONE"
     ALL = "ALL"
@@ -2238,7 +2243,7 @@ class DataScienceModel(Builder):
     def create_custom_metadata_artifact(
         self,
         metadata_key_name: str,
-        artifact_path_or_content: str,
+        artifact_path_or_content: Union[str, bytes],
         path_type: MetadataArtifactPathType = MetadataArtifactPathType.LOCAL,
     ) -> ModelMetadataArtifactDetails:
         """Creates model custom metadata artifact for specified model.
@@ -2248,8 +2253,10 @@ class DataScienceModel(Builder):
         metadata_key_name: str
             The name of the model custom metadata key
 
-        artifact_path_or_content: str
-            The model custom metadata artifact path to be upload. It can also be the actual content of the custom metadata
+        artifact_path_or_content: Union[str,bytes]
+            The model custom metadata artifact path to be upload. It can also be the actual content of the defined metadata
+            The type is string when it represents local path or oss path.
+            The type is bytes when it represents content itself
 
         path_type: MetadataArtifactPathType
             Can be either of MetadataArtifactPathType.LOCAL , MetadataArtifactPathType.OSS , MetadataArtifactPathType.CONTENT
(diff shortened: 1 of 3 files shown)


ads/model/datascience_model.py
```diff
@@ -2442,7 +2498,7 @@ def get_custom_metadata_artifact(
         )
         artifact_file_path = os.path.join(target_dir, f"{metadata_key_name}")
 
-        if not override and os.path.exists(artifact_file_path):
+        if not override and is_path_exists(artifact_file_path):
             raise FileExistsError(f"File already exists: {artifact_file_path}")
```
mrDzurb, line 2502: Should we say to use the override param to override?

ads/model/service/oci_datascience_model.py
```diff
@@ -656,10 +660,9 @@ def get_metadata_content(
         elif path_type == MetadataArtifactPathType.OSS:
             if not utils.is_path_exists(artifact_path_or_content):
                 raise FileNotFoundError(f"File not found: {artifact_path_or_content}")
-
-            contents = str(
-                read_file(file_path=artifact_path_or_content, auth=default_signer())
-            )
+            contents = read_file(
```
mrDzurb, line 663: just curious, why do we want to distinguish between local and OSS? Can't we use the same logic for both?

mrDzurb (dismissed):

ads/model/datascience_model.py
kumar-shivam-ranjan: Yes we can.  Will update the error message.

ads/model/service/oci_datascience_model.py
kumar-shivam-ranjan: I didn't want to change the `read_file` function defintion in `ads.common.utils` since it is already used in many other places.
read_file returns string content of the file and we need bytes.

kumar-shivam-ranjan: Although if i think now , we can use same logic for both actually.

ads/model/datascience_model.py
kumar-shivam-ranjan: Updated.

ads/model/service/oci_datascience_model.py
kumar-shivam-ranjan: Now , if think we could have used same logic though without changing read_file function definition.
Updated it. Thanks!
@mrDzurb

lu-ohai (dismissed):

ads/model/service/oci_datascience_model.py
```diff
@@ -621,15 +621,19 @@ def is_model_created_by_reference(self):
         return False
 
     def get_metadata_content(
-        self, artifact_path_or_content: str, path_type: MetadataArtifactPathType
+        self,
+        artifact_path_or_content: Union[str, bytes],
+        path_type: MetadataArtifactPathType,
     ):
```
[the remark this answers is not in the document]
kumar-shivam-ranjan: get_metadata_content always returns bytes. Updated. Thanks!

lu-ohai (approved):

mrDzurb (approved):

PR #1135, opened 2025-03-27
AD fixes: preprocessing steps, raise e remove

codeloop:


ads/opctl/operator/lowcode/anomaly/model/base_model.py
```diff
@@ -53,6 +54,11 @@ def __init__(self, config: AnomalyOperatorConfig, datasets: AnomalyDatasets):
         self.config: AnomalyOperatorConfig = config
         self.spec: AnomalyOperatorSpec = config.spec
         self.datasets = datasets
+
+        if self.spec.preprocessing.steps.missing_value_imputation:
+            for series_id, df in self.datasets.full_data_dict.items():
+                self.datasets.full_data_dict[series_id] = Transformations._fill_na(self, df)
```
ahosler, line 60: Is this the right spot for this? Why not with the rest of the transformations class?

ahosler (changes requested):

ads/opctl/operator/lowcode/anomaly/model/base_model.py
codeloop: we are reusing the methods from the transformation class, as they need to be enabled based upon the pre-processing flags in the yaml schema

prasankh (approved):

ahosler (approved):

PR #1138, opened 2025-03-28
Improved AQUA Error Messages for Authorization and Tag-Related Uses

elizjo:
Currently, the AQUA UI throws generic errors when auth issues arise- not informing the user of potential solutions and suggested policy additions to fix the error.

Added 'troubleshooting_tips' section (see example json below)
- links to the specific section (as specified by the operation error from the OCI Service Error) to our AQUA troubleshooting page 
**result** : user will obtain a specific action to fix their auth issue 

Distinguish between Auth error versus tag related error.
- certain operations have two methods for failure- auth related and tag related. We use the OCI Service Error to single out 'Invalid tag' errors for the create_model and update_model operations
**result**: user obtains a different fix for a tag-related error vs. an auth error

Refactored write_error in used in base_handler.py and WS handlers to utils.py (basically common code between these handlers were stuck into a single file)

```
{
  "status": 404,
  "troubleshooting_tips": "Unable to list model deployments. See tips for troubleshooting: https://github.com/oracle-samples/oci-data-science-ai-samples/blob/main/ai-quick-actions/troubleshooting-tips.md#list-model-deployments",
  "message": "Authorization Failed: The resource you're looking for isn't accessible. Operation Name: list_model_deployments.",
  "service_payload": {
    "target_service": "data_science",
    "status": 404,
```

### DIFF
--- a/ads/aqua/extension/model_handler.py
+++ b/ads/aqua/extension/model_handler.py
@@ -56,6 +56,7 @@ class AquaModelHandler(AquaAPIhandler):
                     ),
                 )
         elif not model_id:
+            print("***" , model_id)
             return self.list()
 
         return self.read(model_id)
--- a/ads/aqua/extension/utils.py
+++ b/ads/aqua/extension/utils.py
@@ -1,14 +1,17 @@
 #!/usr/bin/env python
 # Copyright (c) 2024 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+import traceback
+import uuid
 from dataclasses import fields
 from datetime import datetime, timedelta
+from http.client import responses
 from typing import Dict, Optional
 
 from cachetools import TTLCache, cached
 from tornado.web import HTTPError
 
-from ads.aqua import ODSC_MODEL_COMPARTMENT_OCID
+from ads.aqua import ODSC_MODEL_COMPARTMENT_OCID, logger
 from ads.aqua.common.utils import fetch_service_compartment
 from ads.aqua.extension.errors import Errors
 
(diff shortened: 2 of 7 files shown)


ads/aqua/extension/model_handler.py
```diff
@@ -56,6 +56,7 @@ def get(
                     ),
                 )
         elif not model_id:
+            print("***" , model_id)
```
VipulMascarenhas, line 59: remove print statement, use logger if necessary.

ads/aqua/extension/utils.py
```diff
@@ -32,3 +36,133 @@ def ui_compatability_check():
...
+        "list_data_science_private_endpoints": "Unable to access specified Object Storage Bucket. See tips for troubleshooting:  ",
+        "create_model": "Unable to register or create model. See tips for troubleshooting: ",
+        "create_model_invalid_tags": "Unable to add freeform tags. Use AQUA CLI to add tags, see tips here: ",
+    }
+    tip = "For general tips on troubleshooting: https://github.com/oracle-samples/oci-data-science-ai-samples/blob/main/ai-quick-actions/troubleshooting-tips.md"
+
+    if status_code in (404, 400):
+        failed_operation = service_payload.get("operation_name")
+
+        if "Invalid tags" in service_payload["message"]:
```
VipulMascarenhas, line 97: why do we need to hardcode this error message? I think we can use the operation name as is, right? We can combine the details in the Create Model section of the troubleshooting guide for any errors related to create_model.

ads/aqua/extension/utils.py
```diff
@@ -32,3 +36,133 @@ def ui_compatability_check():
...
+
+        link = get_documentation_link(failed_operation)
+
+        if failed_operation in operations and link:
+            tip = operations[failed_operation] + link
+
+    return tip
+
+
+def construct_error(status_code, **kwargs):
```
VipulMascarenhas, line 108: can you add typing to function definition? something like:
`status_code: int, **kwargs: Any) -> Tuple[dict, str, str]:`

ads/aqua/extension/utils.py
```diff
@@ -32,3 +36,133 @@ def ui_compatability_check():
...
+    messages = {
+        "400": "Something went wrong with your request.",
+        "403": "We're having trouble processing your request with the information provided.",
+        "404": "Authorization Failed: The resource you're looking for isn't accessible.",
+        "408": "Server is taking too long to response, please try again.",
+        "create": "Authorization Failed: Could not create resource.",
+        "get": "Authorization Failed: The resource you're looking for isn't accessible.",
+    }
+
+    if service_payload and "operation_name" in service_payload:
```
VipulMascarenhas, line 57: the auth code check should be done here. See comment in jira ticket.

VipulMascarenhas (commented):
we can set the target branch to main instead of the feature branch for this PR.

elizjo:
> we can set the target branch to main instead of the feature branch for this PR.

Decided to merge into feature/multi_model_deployment since I based ODSC-70595/multi_model_deployment on  feature/multi_model_deployment. will use main in future if applicable.

ads/aqua/extension/utils.py
mayoor: I wonder if we can use these API names as the anchor tag on the troubleshooting guide.

Something like this - 

```https://github.com/oracle-samples/oci-data-science-ai-samples/blob/main/ai-quick-actions/troubleshooting-tips.md#list_buckets```

elizjo: We are using the API names as the anchor tag as you show. This is only hardcoded to differentiate between create_model 404 authentication error and 400 invalid tag error.

PR #1147, opened 2025-04-04
[DRCC][AQUA] Completely Removing OSS bucket config dependency

kumar-shivam-ranjan:
# Description
This PR is intended to completely remove the dependency of AQUA on config stored in Object storage bucket.  Also , this change would ensure Aqua will be available for all realms even when no `ODSC_MODEL_COMPARTMENT_OCID` env variable is set. 

# Unit Tests

Screenshot 2025-04-05 at 12 28 04 AM

### DIFF
--- a/ads/aqua/__init__.py
+++ b/ads/aqua/__init__.py
@@ -14,7 +14,6 @@ from ads.aqua.client.client import (
     get_async_httpx_client,
     get_httpx_client,
 )
-from ads.aqua.common.utils import fetch_service_compartment
 from ads.config import OCI_RESOURCE_PRINCIPAL_VERSION
 
 ENV_VAR_LOG_LEVEL = "ADS_AQUA_LOG_LEVEL"
@@ -41,5 +40,5 @@ if OCI_RESOURCE_PRINCIPAL_VERSION:
     set_auth("resource_principal")
 
 ODSC_MODEL_COMPARTMENT_OCID = (
-    os.environ.get("ODSC_MODEL_COMPARTMENT_OCID") or fetch_service_compartment()
+    os.environ.get("ODSC_MODEL_COMPARTMENT_OCID") or "ODSC_MODEL_COMPARTMENT_OCID"
 )
--- a/ads/aqua/__init__.py
+++ b/ads/aqua/__init__.py
@@ -38,7 +38,3 @@ def set_log_level(log_level: str):
 
 if OCI_RESOURCE_PRINCIPAL_VERSION:
     set_auth("resource_principal")
-
-ODSC_MODEL_COMPARTMENT_OCID = (
-    os.environ.get("ODSC_MODEL_COMPARTMENT_OCID") or "ODSC_MODEL_COMPARTMENT_OCID"
-)
--- a/ads/aqua/extension/common_ws_msg_handler.py
+++ b/ads/aqua/extension/common_ws_msg_handler.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright (c) 2024 Oracle and/or its affiliates.
+# Copyright (c) 2024, 2025 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 
 import json
@@ -8,15 +8,12 @@ from importlib import metadata
 from typing import List, Union
 
 from ads.aqua.common.decorator import handle_exceptions
-from ads.aqua.common.errors import AquaResourceAccessError
-from ads.aqua.common.utils import known_realm
 from ads.aqua.extension.aqua_ws_msg_handler import AquaWSMsgHandler
 from ads.aqua.extension.models.ws_models import (
     AdsVersionResponse,
     CompatibilityCheckResponse,
     RequestResponseType,
 )
-from ads.aqua.extension.utils import ui_compatability_check
 
 
 class AquaCommonWsMsgHandler(AquaWSMsgHandler):
@@ -39,19 +36,8 @@ class AquaCommonWsMsgHandler(AquaWSMsgHandler):
             )
             return response
         if request.get("kind") == "CompatibilityCheck":
-            if ui_compatability_check():
-                return CompatibilityCheckResponse(
-                    message_id=request.get("message_id"),
-                    kind=RequestResponseType.CompatibilityCheck,
-                    data={"status": "ok"},
-                )
-            elif known_realm():
-                return CompatibilityCheckResponse(
-                    message_id=request.get("message_id"),
-                    kind=RequestResponseType.CompatibilityCheck,
-                    data={"status": "compatible"},
-                )
-            else:
-                raise AquaResourceAccessError(
-                    "The AI Quick actions extension is not compatible in the given region."
-                )
+            return CompatibilityCheckResponse(
+                message_id=request.get("message_id"),
+                kind=RequestResponseType.CompatibilityCheck,
+                data={"status": "ok"},
+            )
--- a/ads/aqua/extension/common_ws_msg_handler.py
+++ b/ads/aqua/extension/common_ws_msg_handler.py
@@ -5,13 +5,12 @@
 
 import json
 from importlib import metadata
-from typing import List, Union
+from typing import List, Optional, Union
 
 from ads.aqua.common.decorator import handle_exceptions
 from ads.aqua.extension.aqua_ws_msg_handler import AquaWSMsgHandler
 from ads.aqua.extension.models.ws_models import (
     AdsVersionResponse,
-    CompatibilityCheckResponse,
     RequestResponseType,
 )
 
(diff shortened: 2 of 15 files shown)


ads/aqua/__init__.py
```diff
@@ -41,5 +40,5 @@ def set_log_level(log_level: str):
     set_auth("resource_principal")
 
 ODSC_MODEL_COMPARTMENT_OCID = (
-    os.environ.get("ODSC_MODEL_COMPARTMENT_OCID") or fetch_service_compartment()
+    os.environ.get("ODSC_MODEL_COMPARTMENT_OCID") or "ODSC_MODEL_COMPARTMENT_OCID"
```
mrDzurb, line 43: I'm not sure that I understand this logic. In one case we return the value of the ODSC_MODEL_COMPARTMENT_OCID env var, but if it is not defined we just return the name of the env var?

ads/aqua/extension/utils.py
```diff
@@ -41,14 +40,14 @@ def ui_compatability_check():
     """This method caches the service compartment OCID details that is set by either the environment variable or if
     fetched from the configuration. The cached result is returned when multiple calls are made in quick succession
     from the UI to avoid multiple config file loads."""
-    return ODSC_MODEL_COMPARTMENT_OCID or fetch_service_compartment()
+    return ODSC_MODEL_COMPARTMENT_OCID
```
mrDzurb, line 43: Do we really need this compatibility check method now?

ads/aqua/__init__.py
kumar-shivam-ranjan: Currently , service model compartment ocid is used as a key to cache service model list response.  Now since we won't be fetching this compartment id in ADS from container_index.json , just replaced it with constant string.
If we explicitly set this compartment ocid via NB environment variables , then corresponding ocid will be used as key.

ads/aqua/extension/utils.py
kumar-shivam-ranjan: Yes , can be removed.
Will remove

ads/aqua/__init__.py
kumar-shivam-ranjan: Should we just remove `ODSC_MODEL_COMPARTMENT_OCID`?
I need to check if we can still use non prod backend for fetching service models list incase we do that since this is only place i feel this env variable is being used.

mrDzurb: Here i'm not sure. My vote would be for removing this variable, but @VipulMascarenhas might have a different opinion. We might use this env variable to identify if the model is service or a custom one.

kumar-shivam-ranjan: Tested the flow after removing `ODSC_MODEL_COMPARTMENT_OCID`.  Seems like we don't need this env variable for fetching service models even for non prod backend.  The new List models API would take care of it.  
We can rename the key for storing service models list response in cache.  

So , everything seems fine except there is only once place this is being used where i am not sure it is safe to remove.
CC:  @VipulMascarenhas 

```
    if service_model.compartment_id != ODSC_MODEL_COMPARTMENT_OCID:
        logger.info(
            f"Aqua Model {model_id} already exists in the user's compartment."
            "Skipped copying."
        )
        return service_model
```

kumar-shivam-ranjan: Is it safe to remove this condition?

kumar-shivam-ranjan: Replaced it with 
```
  # Skip model copying if it is registered model
  if service_model.freeform_tags.get(Tags.BASE_MODEL_CUSTOM) is not None:
      logger.info(
          f"Aqua Model {model_id} already exists in the user's compartment."
          "Skipped copying."
      )
      return service_model
```

ads/aqua/extension/common_ws_msg_handler.py
```diff
@@ -39,19 +36,8 @@ def process(self) -> Union[AdsVersionResponse, CompatibilityCheckResponse]:
             )
             return response
         if request.get("kind") == "CompatibilityCheck":
-            if ui_compatability_check():
```
VipulMascarenhas, line 42: if this always returns ok, does UI need to call this? This won't be needed, right?

ads/aqua/model/model.py
```diff
@@ -181,7 +182,9 @@ def create(
         service_model = DataScienceModel.from_id(model_id)
         target_project = project_id or PROJECT_OCID
         target_compartment = compartment_id or COMPARTMENT_OCID
-        if service_model.compartment_id != ODSC_MODEL_COMPARTMENT_OCID:
+
+        # Skip model copying if it is registered model
+        if service_model.freeform_tags.get(Tags.BASE_MODEL_CUSTOM) is not None:
```
VipulMascarenhas, line 187: can freeform_tags be None? I don't think it's going happen for aqua models, but can we add a check here to return appropriate error instead if for some reason these tags are empty.

ads/aqua/model/model.py
```diff
@@ -949,25 +952,26 @@ def list(
                 category="aqua/service/model", action="list"
             )
 
-            if ODSC_MODEL_COMPARTMENT_OCID in self._service_models_cache:
+            if AQUA_SERVICE_MODELS in self._service_models_cache:
                 logger.info(
-                    f"Returning service models list in {ODSC_MODEL_COMPARTMENT_OCID} from cache."
+                    f"Returning service models list in {AQUA_SERVICE_MODELS} from cache."
```
VipulMascarenhas, line 957: nit: might be good to just log this as `Returning service models list from cache.`

ads/aqua/model/model.py
```diff
@@ -949,25 +952,26 @@ def list(
...
+                return self._service_models_cache.get(AQUA_SERVICE_MODELS)
             logger.info("Fetching service models.")
             lifecycle_state = kwargs.pop(
                 "lifecycle_state", Model.LIFECYCLE_STATE_ACTIVE
             )
 
             models = self.list_resource(
                 self.ds_client.list_models,
-                compartment_id=ODSC_MODEL_COMPARTMENT_OCID,
+                compartment_id=compartment_id or COMPARTMENT_OCID,
```
VipulMascarenhas, line 967: wondering if we need to filter the model list here based on service model tags. If we list from users compartment, this would only return service models and not other registered models from users compartment?

ads/aqua/model/model.py
```diff
@@ -949,25 +952,26 @@ def list(
...
                 lifecycle_state=lifecycle_state,
+                category=category,
                 **kwargs,
             )
 
         logger.info(
-            f"Fetched {len(models)} model in compartment_id={ODSC_MODEL_COMPARTMENT_OCID if category==SERVICE else compartment_id}."
+            f"Fetched {len(models)} models from {AQUA_SERVICE_MODELS if category==SERVICE else compartment_id}."
```
VipulMascarenhas, line 974: the message might not be clear with the new service model key, let's just log: 
Fetched x service models.
or
Fetched x models from compartment <compartment_id>

ads/aqua/extension/common_ws_msg_handler.py
kumar-shivam-ranjan: This is compatibility check handler implemented in websockets.  Currently , we don't call it via UI . We call the one implemented via HTTP.  Both the methods should return ok always as we decide to enable AQUA everywhere.

kumar-shivam-ranjan: CC: @mayoor

kumar-shivam-ranjan: Yes , we can remove calling this API in UI.  I kept it since may be in future we can add some sort of AQUA enabling logic based on CP config.

ads/aqua/model/model.py
kumar-shivam-ranjan: Will update.

ads/aqua/model/model.py
kumar-shivam-ranjan: Not sure if i fully understand your question 
but this would return either service models or user models based on category.  If category is user (default) , it would return all registered models from user's compartment

ads/aqua/model/model.py
kumar-shivam-ranjan: Sure , will update

kumar-shivam-ranjan: Updated. Thanks

ads/aqua/model/model.py
kumar-shivam-ranjan: Updated. Thanks

ads/aqua/model/model.py
kumar-shivam-ranjan: Updated. Thanks

ads/aqua/model/model.py
kumar-shivam-ranjan: Updated. Thanks

ads/aqua/extension/common_ws_msg_handler.py
kumar-shivam-ranjan: Updated. Thanks

ads/aqua/extension/utils.py
kumar-shivam-ranjan: Updated. Thanks

mrDzurb (approved):

VipulMascarenhas (approved):

PR #1152, opened 2025-04-11
Fixed ads dataset key error

lu-ohai:
### Fixed ads dataset key error

Screenshot 2025-04-10 at 8 57 46 PM


qiuosier (approved):

ads/type_discovery/typed_feature.py
```diff
@@ -583,14 +581,14 @@ def build(name, series):
                 "missing_percentage": 100 * series.isna().sum() / series.size,
                 "low_level_type": series.dtype.name,
                 "stats": {
-                    "unique percentage": 100 * desc["unique"] / desc["count"],
-                    "first": desc["first"],
-                    "last": desc["last"],
-                    "mode": desc["top"],
+                    "unique percentage": 100 * series.nunique() / series.size,
+                    "first": desc["first"] if "first" in desc else None,
```
mrDzurb, line 585: Why can't we use ```.get()```?

```
"first": desc.get("first"),
```

mrDzurb (approved):

ads/type_discovery/typed_feature.py
lu-ohai: Wrote this just in case the `desc` doesn't have a `get` method, but if it always does I can change this to `get`

VipulMascarenhas (approved):

PR #1163, opened 2025-04-23
[AQUA] Adding ADS support for embedding models in Multi Model Deployment

elizjo:
Screenshot 2025-04-22 at 6 49 37 PM

This PR adds support of using embedding models in a multi model deployment. 

To accomplish this, we have to pass data specifying the model task from ADS to the MULTI_MODEL_CONFIG environment variable in model deployment. 

```
##Before PR
MULTI_MODEL_CONFIG=
'{ "models": [ 
{ "params": "--served-model-name bge 
--tensor-parallel-size 1 
--trust-remote-code 
--max-model-len 4096", 
"model_path": "bge-m3" } , { "params": "--served-model-name llama --enforce-eager --max-num-seqs 16 --tensor-parallel-size 2 --max-model-len 16000", "model_path": "Llama-3.2-11B-Vision" }] }'

##After PR- see 'model_task' key
MULTI_MODEL_CONFIG=
'{ "models": [  
{ "params": "--served-model-name bge
 --tensor-parallel-size 1 
--trust-remote-code 
--max-model-len 4096", 
"model_path": "bge-m3", 
"model_task": "embedding" }, { "params": "--served-model-name llama --enforce-eager --max-num-seqs 16 --tensor-parallel-size 2 --max-model-len 16000", "model_path": "Llama-3.2-11B-Vision" }, ] }'
```

We only have 'model_task' key for embedding models used in a multi model deployment. 
- We extract whether a model is an embedding model by reading the freeform tags ('task' tag) of the model (to determine whether model is an embedding model)
- We added model_task as an optional parameter in the AquaMultiModelRef object
- the model_task parameter is used to construct the MULTI_MODEL_CONFIG which has added (model_task: "embedding") key.

All unit tests pass (see screenshot). This PR was tested by modifying the existing unit test test_create_deployment_for_multi_model.

### DIFF
--- a/ads/aqua/model/model.py
+++ b/ads/aqua/model/model.py
@@ -712,7 +712,7 @@ class AquaModelApp(AquaApp):
             model.model_task = source_model.freeform_tags.get(Tags.TASK, UNKNOWN)
 
         task_tag = re.sub(r"-", "_", model.model_task).lower()
-
+        # re-visit logic when more model task types are supported
         if task_tag in MultiModelSupportedTaskType:
             model.model_task = task_tag
         else:


ads/aqua/model/model.py
```diff
@@ -316,6 +316,11 @@ def create_multi(
 
             display_name_list.append(display_name)
 
+            model_task = source_model.freeform_tags.get(Tags.TASK, UNKNOWN)
```
mrDzurb, line 319: I would rather to move this logic to the ```_get_task()``` method   

```
model.taks = self._get_task(model, source_model)
```

```
def _get_task(model_ref:AquaMultiModelRef, source_model: DataScienceModel) -> str:
    # extract task from model_ref by itself, if task is not presented there, then extract it from the freeform tags. 
    # model_task = source_model.freeform_tags.get(Tags.TASK, UNKNOWN)
    
    ....
    return taks
```

I believe we should also allow users to pass task within ```AquaMultiModelRef```,  just in case if the tags were not populated well.

ads/aqua/model/model.py
```diff
@@ -707,6 +700,25 @@ def edit_registered_model(
         else:
             raise AquaRuntimeError("Only registered unverified models can be edited.")
 
+    def _get_task(
```
mrDzurb, line 703: Looks like this method doesn't return any value, yet its signature indicates a return type of str. Should we update the type hint to reflect that it returns None, or adjust the implementation to return a string as specified?

ads/aqua/model/model.py
```diff
@@ -307,18 +308,10 @@ def create_multi(
...
-            #     not in MultiModelSupportedTaskType
-            # ):
-            #     raise AquaValueError(
-            #         f"Invalid or missing {Tags.TASK} tag for selected model {display_name}. "
-            #         f"Currently only `{MultiModelSupportedTaskType.values()}` models are supported for multi model deployment."
-            #     )
-
             display_name_list.append(display_name)
 
+            self._get_task(model, source_model)
```
mrDzurb, line 313: I think It might be clearer if we do something like this:

```
model.model_task =  self._extract_model_task(model, source_model)
```

ads/aqua/model/enums.py
```diff
@@ -28,3 +28,8 @@ class FineTuningCustomMetadata(ExtendedEnum):
 class MultiModelSupportedTaskType(ExtendedEnum):
     TEXT_GENERATION = "text-generation"
     TEXT_GENERATION_ALT = "text_generation"
+    EMBEDDING_ALT = "text_embedding"
```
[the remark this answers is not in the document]
elizjo: fixed- we add embedding in SMC level

ads/aqua/model/model.py
elizjo: we allow user to pass task, or if not provided, use the freeform tags of the source model

ads/aqua/model/model.py
```diff
@@ -707,6 +700,25 @@ def edit_registered_model(
...
+        # user does not supply model task, we extract from model metadata
+        if not model.model_task:
+            model.model_task = source_model.freeform_tags.get(Tags.TASK, UNKNOWN)
+
+        task_tag = re.sub(r"-", "_", model.model_task).lower()
+
+        if task_tag in MultiModelSupportedTaskType:
+            model.model_task = task_tag
+        else:
+            raise AquaValueError(
```
mrDzurb, line 718: We can show more informative error: 

```
raise AquaValueError(
     f"Invalid or missing {task_tag} tag for selected model {display_name}. "
     f"Currently only `{MultiModelSupportedTaskType.values()}` models are supported for multi model deployment."
```

elizjo: fixed

mrDzurb (dismissed):

ads/aqua/model/model.py
VipulMascarenhas: since we removed the task level validation in the recent release, any reason to add the validation in the function `_extract_model_task again? 

This is fine for now since we only have 1 verified embedding model, but if in the future if we start supporting (unverified) models, embedding models could have task value as feature_extraction or sentence_similarity. Might be good to add a comment here to reconsider this logic when we start supporting additional models.

VipulMascarenhas (approved):
minor comment, overall looks good.

mrDzurb (approved):

VipulMascarenhas (approved):

PR #1166, opened 2025-04-24
ODSC-70796 instead of threads for each telemetry event, created a glo…

agrimk:
instead of threads for each telemetry event, created a global thread pool


ads/telemetry/telemetry.py
```diff
@@ -23,7 +25,8 @@
 USER_AGENT_KEY = "additional_user_agent"
 UNKNOWN = "UNKNOWN"
 DELIMITER = "&"
-
+THREAD_POOL_SIZE = 16
+thread_pool = concurrent.futures.ThreadPoolExecutor(max_workers=THREAD_POOL_SIZE)
```
mrDzurb, line 29: The `telemetry.py` and `client.py` modules handle different aspects of telemetry. I suggest moving the thread pool creation to `client.py`.

ads/aqua/extension/__init__.py
```diff
@@ -37,6 +37,12 @@ def load_jupyter_server_extension(nb_server_app):
         [(url_path_join(route_pattern, url), handler) for url, handler in __handlers__],
     )
 
+    def run_on_shutdown():
+        web_app.log.info("Shutting down telemetry threadpool...")
+        thread_pool.shutdown(wait=True)
+        web_app.log.info("Telemetry threadpool shutdown completed.") 
+
+    web_app.io_loop.add_callback(run_on_shutdown)
```
mrDzurb, line 45: I'm not sure about this logic. Technically user can use AQUA CLI where we capture the telemetry as well. Users don't need to run extension in this case.

agrimk: Do you mean that I should call the shutdown there as well ?

ads/telemetry/client.py
```diff
@@ -117,9 +121,5 @@ def record_event_async(
...
             A started thread to send a head request to generate an event record.
         """
-        thread = threading.Thread(
-            target=self.record_event, args=(category, action, detail), kwargs=kwargs
-        )
-        thread.daemon = True
-        thread.start()
-        return thread
+        thread_pool.submit(self.record_event, args=(category, action, detail), kwargs=kwargs)
+        return None
```
mrDzurb, line 125: We probably don't need to return anything form this method, however it would be nice to add return typing:

```
def record_event_async(
        self, category: str = None, action: str = None, detail: str = None, **kwargs
    ) -> None:
```

ads/aqua/extension/__init__.py
```diff
@@ -37,6 +37,12 @@ def load_jupyter_server_extension(nb_server_app):
         [(url_path_join(route_pattern, url), handler) for url, handler in __handlers__],
     )
 
+    def run_on_shutdown():
```
mrDzurb, line 40: Do we really need this? ADS can be used without the running the Jlab.

ads/aqua/extension/__init__.py
mrDzurb: No I men, do we really need to shut it down? In which case it would be useful?

agrimk: ok, got it. I am removing the shutdown code in that case. 
I added it since it was part of the ticket [ODSC-70796](https://jira.oci.oraclecorp.com/browse/ODSC-70796)

mrDzurb (approved):

PR #1175, opened 2025-05-05
Enable AQUA SDK & CLI to Deploy Fine-Tuned LLMs in Multi-Model Deployment

elizjo:
This PR aims to add support for using  fine-tuned models in a Multi-Model deployment.

**When deploying a model group w/ a fine tuned model** 
- added FT_MODEL parameter (output path in oci bucket for fine tuned model) 
- we extract FT_MODEL from the fine tune single model deployment. The model_path is assumed to be the base model for the fine tuned model.
```
{ "models": [ { "params": "--served-model-name Llama-3.2-11B-Vision --enforce-eager --max-num-seqs 16 --tensor-parallel-size 2 --max-model-len 16000",  "fine_tune_path" : "oci://model-path", "model_path": "Llama-3.2-11B-Vision", "model_task": "image-text-to-text"}, ... 
```


ads/aqua/model/model.py
```diff
@@ -311,6 +312,12 @@ def create_multi(
             #         "Currently only service models are supported for multi model deployment."
             #     )
 
+            is_fine_tuned_model = Tags.AQUA_FINE_TUNED_MODEL_TAG in source_model.freeform_tags
```
mrDzurb, line 315: Could you add some comment what we are doing here?

ads/aqua/finetuning/entities.py
```diff
@@ -179,3 +185,42 @@ class CreateFineTuningDetails(Serializable):
 
     class Config:
         extra = "ignore"
+
+
+def extract_base_model_ocid(aqua_model: DataScienceModel) -> Tuple[str, str]:
```
mrDzurb, line 190: The name of the function is a bit confusing, it says that it extracts the base model ocid, but returns name and ocid. 
Also I think it would be better to create ```model/utils.py``` module  and move this logic there. Same for the ```set_finetune_env_var```.

ads/aqua/finetuning/entities.py
```diff
@@ -179,3 +185,42 @@ class CreateFineTuningDetails(Serializable):
...
+    aqua_model.model_file_description
+    )
+
+    if fine_tune_output_path and ObjectStorageDetails.is_oci_path(fine_tune_output_path):
+        os_path = ObjectStorageDetails.from_path(fine_tune_output_path)
+        fine_tune_output_path = os_path.filepath.rstrip("/")
+
+        env_var.update({"FT_MODEL": f"{fine_tune_output_path}"})
+
+        return env_var
```
mrDzurb, line 222: looks like we are already modifying the original env_var, probably we don't need to return anything form this function?

mrDzurb:
Could you add more description in the PR? It would be also helpful if you put in the details how the new config will look like.

ads/aqua/finetuning/entities.py
elizjo: changed, see new commit

ads/aqua/finetuning/entities.py
elizjo: moved logic as suggested

ads/aqua/model/utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+    if not config_source_id or not model_name:
+        raise AquaValueError(
+        f"Either {FineTuneCustomMetadata.FINE_TUNE_SOURCE} or {FineTuneCustomMetadata.FINE_TUNE_SOURCE_NAME} is missing "
+        f"from custom metadata for the model {config_source_id}")
+
+    return config_source_id, model_name
+
+
+def set_fine_tune_env_var(aqua_model: DataScienceModel, env_var: Optional[Dict[str,str]], model: Optional[AquaMultiModelRef] = None) -> None:
```
mrDzurb, line 34: I'm not sure if we really want to combine two flows in one scope. Do we really need  to setup the```FT_MODEL``` env variable for multi-model deployment case? If not, I would prefer to keep single-model and multi-model deployment logics separately.

ads/aqua/model/utils.py
```diff
@@ -0,0 +1,52 @@
...
+    if not config_source_id or not model_name:
+        raise AquaValueError(
+            f"Either {FineTuneCustomMetadata.FINE_TUNE_SOURCE} or {FineTuneCustomMetadata.FINE_TUNE_SOURCE_NAME} is missing "
+            f"from custom metadata for the model {config_source_id}"
+        )
+
+    return config_source_id, model_name
+
+
+def set_fine_tune_env_var(aqua_model: DataScienceModel) -> Tuple[str, str]:
```
mrDzurb, line 35: I think the name needs to be changed as well, it doesn't do anything related to the env variables, right?

mrDzurb (approved):

mrDzurb (approved):

ads/aqua/model/model.py
```diff
@@ -311,6 +315,14 @@ def create_multi(
             #         "Currently only service models are supported for multi model deployment."
             #     )
 
+            # check if model is a fine-tuned model and if so, pass FT_MODEL info to model's
+            # env variables & set model_path to be the base model
+            is_fine_tuned_model = Tags.AQUA_FINE_TUNED_MODEL_TAG in source_model.freeform_tags
+
+            if is_fine_tuned_model:
```
lu-ohai, line 322: We should add an `else` statement otherwise `model.artifact_location` will be duplicately assigned below: https://github.com/oracle/accelerated-data-science/blob/main/ads/aqua/model/model.py#L327

elizjo: good catch- I have fixed this issue

mrDzurb (approved):

PR #1176, opened 2025-05-06
Merge local and remote GPU shapes index, preferring Object Storage overrides

mrDzurb:
## Description:
This PR enhances the `load_gpu_shapes_index` function to merge GPU-shape definitions from two sources—an OCI Object Storage bucket and a local JSON file, so that any shapes defined remotely override their local counterparts, while still falling back cleanly if the remote load fails.

---

### What Changed

1. **Separate remote vs. local loads**

   * We now load `remote_data` and `local_data` into two distinct dictionaries.
   * If the `CONDA_BUCKET_NS` environment variable is set, we attempt to read from Object Storage.

2. **Merge logic**

   * After loading both sources, we merge the two on the `"shapes"`. Any entry present in `remote_data["shapes"]` **overrides** the same key in `local_data`, while preserving local entries not present remotely.

3. **Return the merged index**

### DIFF
--- a/ads/aqua/common/utils.py
+++ b/ads/aqua/common/utils.py
@@ -1158,9 +1158,11 @@ def validate_cmd_var(cmd_var: List[str], overrides: List[str]) -> List[str]:
 
 
 def build_pydantic_error_message(ex: ValidationError):
-    """Added to handle error messages from pydantic model validator.
+    """
+    Added to handle error messages from pydantic model validator.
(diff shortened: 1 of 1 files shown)


lu-ohai (approved):

VipulMascarenhas (approved):
looks good. I was thinking if local shapes should override the remote ones and not the other way round. But probably makes sense to keep default as remote to avoid bad shape combinations that might lead to failures.

PR #1186, opened 2025-05-16
[Ready For Review][AQUA] Add Supporting Fine-Tuned Models in Multi-Model Deployment

mrDzurb:
## Description
The current implementation of Multi-Model Deployment in AQUA supports base models only. Fine-tuned models, however, are a critical part of many customer workflows - allowing them to adapt base models to domain-specific use cases.
This PR introduces support for deploying fine-tuned LLM models as part of a multi-model deployment group on the VLLM container.

## Implementation
In the first iteration, we will treat each selected model, whether it's a base model or a fine-tuned variant—as an independent entity. Even if multiple fine-tuned models share the same base model, each one will be deployed in its own isolated VLLM instance.

On the SMC side, we will leverage VLLM's capability to dynamically merge LoRA adapter weights during runtime. This means each VLLM instance will load the base model and its corresponding fine-tuned weights independently.

To avoid routing conflicts caused by multiple instances using the same base model name, we will route the base model name to one instance only, but we will not advertise this base model as an endpoint to users (This is current behavior with Single Model Deployment).

This configuration structure will prepare us for future enhancements, such as stacked fine-tuned deployments, where multiple fine-tuned variants are hosted under a single base model within one VLLM instance. However, this future enhancement will apply to single-model deployments initially.

In a second iteration, we will explore expanding this capability to multi-model deployments, enabling grouped deployment of fine-tuned variants with shared GPU allocation. That enhancement will require additional work across the ADS SDK, AQUA UI, and validation logic.

## Related PRs
* [Enhance Multi-Model Deployment Configuration to Support FT Models](https://github.com/oracle/accelerated-data-science/pull/1188)
* [Add Support for Retrieving Deployment Configuration from Base Model for Fine-Tuned Models](https://github.com/oracle/accelerated-data-science/pull/1185)

### DIFF
--- a/ads/aqua/evaluation/evaluation.py
+++ b/ads/aqua/evaluation/evaluation.py
@@ -731,7 +731,7 @@ class AquaEvaluationApp(AquaApp):
         for metadata in multi_model_metadata:
             model = AquaMultiModelRef(**metadata)
             model_names.append(model.model_name)
-            model_names.extend(ft.model_name for ft in (model.fine_tune_weights or []))
+            model_names.extend(ft.model_name for ft in (model.fine_tune_weights or []) if ft.model_name)
 
         # Check if the provided model name is among the valid names.
         if user_model_name not in model_names:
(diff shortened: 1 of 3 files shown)


VipulMascarenhas (approved):

VipulMascarenhas (approved):
added minor comment, rest looks good.

elizjo (approved):

PR #1187, opened 2025-05-16
Adds VLLM LLaMA 4 support to the list of compatible container families for Multi-Model deployment.

mrDzurb:
## Description
Adds VLLM LLaMA 4 support to the list of compatible container families for Multi-Model deployment.

### DIFF
--- a/ads/aqua/model/model.py
+++ b/ads/aqua/model/model.py
@@ -316,11 +316,17 @@ class AquaModelApp(AquaApp):
             #     )
 
             # check if model is a fine-tuned model and if so, add the fine tuned weights path to the fine_tune_weights_location pydantic field
-            is_fine_tuned_model = Tags.AQUA_FINE_TUNED_MODEL_TAG in source_model.freeform_tags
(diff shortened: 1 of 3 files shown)


ads/aqua/model/model.py
```diff
@@ -380,7 +386,8 @@ def create_multi(
                 raise AquaValueError(
                     "The selected models are associated with different container families: "
                     f"{list(selected_models_deployment_containers)}."
-                    "For multi-model deployment, all models in the group must share the same container family."
+                    "For multi-model deployment, all models in the group must belong to the same container "
+                    "family or to compatible container families."
```
VipulMascarenhas, line 390: nit: "two" instead of "to"

VipulMascarenhas (commented):
I'm missing the context here, why separate vllm family for llama4 models? Shouldn't we force to v1 for these models?

VipulMascarenhas (approved):
discussed offline, changes look good.

lu-ohai (approved):

PR #1190, opened 2025-05-20
[WIP] [AQUA] Adding handler for streaming inference predict endpoint

kumar-shivam-ranjan:


ads/aqua/extension/deployment_handler.py
```diff
@@ -175,23 +175,11 @@ def list_shapes(self):
...
-            url = urlparse(endpoint)
-            if url.scheme != "https":
-                return False
-            if not url.netloc:
-                return False
-            return url.path.endswith("/predict")
-        except Exception:
-            return False
-
+class AquaDeploymentStreamingInferenceHandler(AquaAPIhandler):
```
mrDzurb, line 178: I didn't know that we use it at all

ads/aqua/modeldeployment/deployment.py
```diff
@@ -1309,4 +1311,57 @@ def list_shapes(self, **kwargs) -> List[ComputeShapeSummary]:
                 or gpu_specs.shapes.get(oci_shape.name.upper()),
             )
             for oci_shape in oci_shapes
-        ]
\ No newline at end of file
+        ]
+
+    @staticmethod
```
mrDzurb, line 1316: There is AQUA client, that can be utilized. 
https://accelerated-data-science.readthedocs.io/en/latest/user_guide/large_language_model/aqua_client.html#usage

ads/aqua/extension/deployment_handler.py
kumar-shivam-ranjan: It was added before we had kernel messaging solution for inference. Since then it has been lying orphaned in code.
Now we can use this handler to expose streaming inference API which our client (AQUA UI) can use for inference instead of keeping the script at UI.

ads/aqua/modeldeployment/deployment.py
kumar-shivam-ranjan: I see.  yeah we can use aqua client instead of model deployment client for streaming inference support. will update. Thanks

ads/aqua/extension/deployment_handler.py
mrDzurb: Got it, let's use it but with AQUA Client.
I've added this [PR](https://github.com/oracle/accelerated-data-science/pull/1191) to update the docs. I think In UI playground it would be useful to add a checkbox where users can choose if they want to use streaming or not. By default it can use streaming end-point.

ads/aqua/modeldeployment/deployment.py
kumar-shivam-ranjan: Updated

ads/aqua/modeldeployment/deployment.py
```diff
@@ -1309,4 +1314,71 @@ def list_shapes(self, **kwargs) -> List[ComputeShapeSummary]:
                 or gpu_specs.shapes.get(oci_shape.name.upper()),
             )
             for oci_shape in oci_shapes
-        ]
\ No newline at end of file
+        ]
+
+    @telemetry(entry_point="plugin=inference&action=get_response", name="aqua")
+    def get_model_deployment_response(
```
mrDzurb, line 1320: NIT: I think we probably don't need it on the API level, having this logic in handler would be sufficient.

mrDzurb (approved):

ads/aqua/modeldeployment/deployment.py
kumar-shivam-ranjan: I might not have properly understood your statement ..
but we have business logic based on predict endpoint type like text/chat completions and even endpoint override feature in this method.

Do you mean adding this logic inside `deployment_handler.py` ?

mrDzurb: Yes, I mean do we really need this logic to be placed in the deployment.py? Will it be used outside of the UI? If not, then probably would be better to move this logic to the handlers.

ads/aqua/extension/deployment_handler.py
```diff
@@ -205,32 +192,31 @@ def post(self, *args, **kwargs):  # noqa: ARG002
...
-        if not endpoint:
-            raise HTTPError(400, Errors.MISSING_REQUIRED_PARAMETER.format("endpoint"))
-
-        if not self.validate_predict_url(endpoint):
-            raise HTTPError(400, Errors.INVALID_INPUT_DATA_FORMAT.format("endpoint"))
-
         prompt = input_data.get("prompt")
-        if not prompt:
-            raise HTTPError(400, Errors.MISSING_REQUIRED_PARAMETER.format("prompt"))
+        messages = input_data.get("messages")
```
VipulMascarenhas, line 196: can we accept chat_template as well since we want to support it?

VipulMascarenhas (approved):

ads/aqua/modeldeployment/deployment.py
kumar-shivam-ranjan: Updated. Thanks

mrDzurb (approved):

VipulMascarenhas (approved):

PR #1193, opened 2025-05-22
[AQUA Telemetry] Update MD Tracking

agrimk:
https://jira.oci.oraclecorp.com/browse/ODSC-70841


ads/aqua/constants.py
```diff
@@ -54,6 +54,8 @@
 SERVICE_MANAGED_CONTAINER_URI_SCHEME = "dsmc://"
 SUPPORTED_FILE_FORMATS = ["jsonl"]
 MODEL_BY_REFERENCE_OSS_PATH_KEY = "artifact_location"
+DEFAULT_WAIT_TIME = 1200
```
VipulMascarenhas, line 57: since these constants are specific to model deployment status, we could have it in the ads.aqua.modeldeployment.constants.py.

ads/aqua/modeldeployment/deployment.py
```diff
@@ -80,6 +84,9 @@
 from ads.model.model_metadata import ModelCustomMetadataItem
 from ads.telemetry import telemetry
 
+THREAD_POOL_SIZE = 16
```
VipulMascarenhas, line 87: we can move this constant in ads.aqua.modeldeployment.constants.py

ads/aqua/modeldeployment/deployment.py
```diff
@@ -80,6 +84,9 @@
...
+thread_pool = concurrent.futures.ThreadPoolExecutor(max_workers=THREAD_POOL_SIZE)
```
VipulMascarenhas, line 88: I think we should use the common telemetry or the common decorator threadpool instead of creating one here.

ads/aqua/modeldeployment/deployment.py
```diff
@@ -1309,4 +1323,33 @@ def list_shapes(self, **kwargs) -> List[ComputeShapeSummary]:
...
+                poll_interval=DEFAULT_POLL_INTERVAL
+            )
+        except Exception as e:
+            logger.error(
+                "Error while trying to create model deployment: " + str(e)
+            )
+            self.telemetry.record_event_async(
+                category=f"aqua/{model_type}/deployment/status",
+                action="FAILED",
+                detail="Error creating model deployment"
```
VipulMascarenhas, line 1346: we should log the error message (`_error_message`) coming from work request here instead of a static message. This will be used to track the specific reasons why the deployment failed.

ads/aqua/modeldeployment/deployment.py
```diff
@@ -1309,4 +1323,33 @@ def list_shapes(self, **kwargs) -> List[ComputeShapeSummary]:
...
+                **telemetry_kwargs
+            )
+        
+        self.telemetry.record_event_async(
```
VipulMascarenhas, line 1350: shouldn't this be within `else` block?

ads/aqua/modeldeployment/deployment.py
```diff
@@ -1309,4 +1323,33 @@ def list_shapes(self, **kwargs) -> List[ComputeShapeSummary]:
...
+                category=f"aqua/{model_type}/deployment/status",
+                action="SUCCEEDED",
+                detail=" Create model deployment successful",
```
VipulMascarenhas, line 1353: we can skip this `detail`, action "SUCCEEDED" implies the same thing.

VipulMascarenhas (commented):
added some comments. 
Can we add some unit tests, and also set up pre-commit hook to format the code changes during commit?

mrDzurb:
Please, don't use Jira ticket as a reference. It would be better to add a description in the PR by itself.

ads/aqua/modeldeployment/constants.py
```diff
@@ -8,3 +8,6 @@
 
 This module contains constants used in Aqua Model Deployment.
 """
+
+DEFAULT_WAIT_TIME = 12000
```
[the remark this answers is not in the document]
VipulMascarenhas: I think there's an extra zero here, should have been 1200 :)

VipulMascarenhas (changes requested):
the imports for MultiModelDeploymentConfigLoader seem to be incorrect, it should be from config_loader and not utils.

VipulMascarenhas (approved):

PR #1219, opened 2025-06-30
[AQUA] Track md logs for error logging

agrimk:
### Description 

In this PR , we have added support to watch predict and access logs (if available) of Model Deployment and pass on error from the logs to telemetry. 

Following changes were made as part of this requirement - 

- we introduced a  '**get_tail_logs**' method in logging module to return the logs as a list for a given log type. Log type can be predict or access. 
- We use the 'get_tail_logs' method in '**tail_logs**' method in model_deployment module and have added some check for cases when logs are not configured.  
- We call this method in '**get_deployment_status**' for both access and predict logs if a model deployment fails . We use the last log from this list and record it in telemetry.

### Jira

https://jira.oci.oraclecorp.com/browse/ODSC-73585 

### Sample error JSON from logs
```
{
     "datetime": 1751029517799,
     "category": "aqua/service/deployment/status/FAILED",
     "action": " Errors occurred while bootstrapping the Model Deployment  Start Container Error  unable to start container  Error response from daemon  OCI runtime create failed  runc create failed  unable to start container process  error during container init  error running hook  0  error running hook  exit status 1  stdout    stderr  Auto detected mode as  legacy  nvidia container cli  requirement error  unsatisfied condition  cuda  12 6  please update your driver to a newer version  or use an earlier cuda container  unknown",
     "value": "6ihbnzka&model_mistralai/Mistral-7B-Instruct-v0.3&status= Errors occurred while bootstrapping the Model Deployment  Start Container Error  unable to start container  Error response from daemon  OCI runtime create failed  runc create failed  unable to start container process  error during container init  error running hook  0  error running hook  exit status 1  stdout    stderr  Auto detected mode as  legacy  nvidia container cli  requirement error  unsatisfied condition  cuda  12 6  please update your driver to a newer version  or use an earlier cuda container  unknown Starting containerStarting container",
     "count": 1,
     "region": "us-ashburn-1",
     "authenticationType": "resource"
   },
```


ads/model/deployment/model_deployment.py
```diff
@@ -728,6 +729,44 @@ def update(
 
         return self._update_from_oci_model(response)
 
+    def tail_logs(
```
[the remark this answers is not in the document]
VipulMascarenhas: +1. I see that  show_logs() within model_deployment.py also covers the logic described below.

ads/model/deployment/model_deployment.py
```diff
@@ -1644,22 +1687,22 @@ def _build_model_deployment_configuration_details(self) -> Dict:
...
+                infrastructure.subnet_id
+            )
 
         if infrastructure.private_endpoint_id:
             if not hasattr(
                 oci.data_science.models.InstanceConfiguration, "private_endpoint_id"
             ):
                 # TODO: add oci version with private endpoint support.
-                raise EnvironmentError(
+                raise OSError(
```
VipulMascarenhas, line 1699: just curious, did ruff suggest this change?

ads/aqua/modeldeployment/deployment.py
```diff
@@ -1300,24 +1304,52 @@ def get_deployment_status(
...
+        except Exception as e:
+            status = ""
+            predict_logs = deployment.tail_logs("predict")
+            access_logs = deployment.tail_logs("access")
+            if access_logs and len(access_logs) > 0:
+                status = access_logs[0]["message"]
+
+            if predict_logs and len(predict_logs) > 0:
+                status += predict_logs[0]["message"]
+            status = re.sub(r"[^a-zA-Z0-9]", " ", status)
```
VipulMascarenhas, line 1316: nit: any reason why we're removing the non-alphanumeric characters here?

ads/aqua/modeldeployment/deployment.py
```diff
@@ -1300,24 +1304,52 @@ def get_deployment_status(
...
-        except Exception:
+        except Exception as e:
+            status = ""
+            predict_logs = deployment.tail_logs("predict")
+            access_logs = deployment.tail_logs("access")
+            if access_logs and len(access_logs) > 0:
+                status = access_logs[0]["message"]
+
+            if predict_logs and len(predict_logs) > 0:
+                status += predict_logs[0]["message"]
```
VipulMascarenhas, line 1315: I'm thinking if we to append both predict and access logs here. For aqua, I think UI passes the same ocids for predict and access logs, so we'll be duplicating the content. Instead, would it make sense to check for access logs first, and only look to add predict logs if these are empty? Better, if we know that both logs are the same, then we can just look at access.

ads/aqua/modeldeployment/deployment.py
```diff
@@ -1300,24 +1304,52 @@ def get_deployment_status(
...
-                action="FAILED",
-                detail=error_str,
-                value=model_name,
-                **telemetry_kwargs,
-            )
+                error_str = re.sub(r"[^a-zA-Z0-9]", " ", error_str)
+                telemetry_kwargs = {
+                    "ocid": ocid,
+                    "model_name": model_name,
+                    "status": error_str + " " + status,
```
VipulMascarenhas, line 1327: let's split this in two fields: "status" can be error_str, and then add "log_message" can be the status content from the logs. This way we can identify the error messages coming from work requests and also the deployment logs separately.

tests/unitary/with_extras/aqua/test_deployment.py
```diff
@@ -2369,6 +2369,7 @@ def test_validate_multimodel_deployment_feasibility_positive_single(
         )
 
     def test_get_deployment_status(self):
+        model_deployment = copy.deepcopy(TestDataset.model_deployment_object[0])
```
VipulMascarenhas, line 2372: can you add a few unit tests to cover the status coming from the logs?

VipulMascarenhas (commented):
suggesting minor changes

ads/model/deployment/model_deployment.py
agrimk: Yes

ads/aqua/modeldeployment/deployment.py
VipulMascarenhas: can you try encoding the characters using urllib.parse instead and see if that helps? 
```
status = urllib.parse.quote(status, safe='')  
```

the missing characters is not a deal breaker, but can lead to not-so-readable output/message. If adding this also results in bad request, then we can go ahead with the change you've added.

VipulMascarenhas (approved):
changes look good.  

I thought the oci sdk would handle the encoding of the object name automatically but might be good to just check if we can keep them by manually encoding it. If it still doesn't result in the proper output, then we can go ahead with the changes you have.

mrDzurb (approved):

PR #1221, opened 2025-07-07
[AQUA] GPU Shape Recommendation

elizjo:
Wrote an additional POST API and aqua command for recommending GPU shapes for a particular model 


```
ads aqua deployment recommend_shape --model_id  'ocid1.datasciencemodel.oc1.<ocid>' 
```
Returns
<img width="1113" height="1059" alt="Screenshot 2025-08-01 at 2 19 10 PM" src="https://github.com/user-attachments/assets/61d216e8-f2d8-4a51-96c9-078d8f8237cc" />


```
GET  /aqua/deployments/recommend_shapes/<md_ocid>
```
Returns
```
{                                                                                                                                                                                                                                            
    "display_name": "Almawave/Velvet-14B",
    "recommendations": [
        {
            "shape_details": {
                "available": false,
                "core_count": null,
                "memory_in_gbs": null,
                "name": "BM.GPU.MI300X.8",
                "shape_series": "GPU",
                "gpu_specs": {
                    "gpu_memory_in_gbs": 1536,
                    "gpu_count": 8,
                    "gpu_type": "MI300X",
                    "quantization": [
                        "fp8",
                        "gguf"
                    ],
                    "ranking": {
                        "cost": 90,
                        "performance": 90
                    }
                }
            },
            "configurations": [
                {
                    "model_details": {
                        "model_size_gb": 28.16,
                        "kv_cache_size_gb": 26.84,
                        "total_model_gb": 55.0
                    },
                    "deployment_params": {
                        "quantization": "bfloat16",
                        "max_model_len": 131072,
                        "params": ""
                    },
                    "recommendation": "No override PARAMS needed. \n\nModel fits well within the allowed compute shape (55.0GB used / 1536.0GB allowed)."
                }
            ]
        }
    ],
     "troubleshoot": ""
}
```

Status: business logic works, API works, unit tests finished, rich diff CLI table finished
<img width="1045" height="389" alt="Screenshot 2025-07-29 at 12 04 57 PM" src="https://github.com/user-attachments/assets/cae1caab-ddd9-4e5d-9c0d-bd3aa73e2ce7" />


ads/aqua/shaperecommend/recommend.py
```diff
@@ -0,0 +1,499 @@
...
+    ModelConfig,
+    RequestRecommend,
+    ShapeRecommendationReport,
+    ShapeReport,
+)
+from ads.config import COMPARTMENT_OCID
+from ads.model.datascience_model import DataScienceModel
+
+
+class AquaRecommendApp(AquaApp):
```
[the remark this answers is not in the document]
mrDzurb: Technically, the shape recommendation feature falls under the broader Model Deployment functionality. Instead of creating a new `App` class, I suggest we implement the business logic in a regular helper class and then integrate it into the existing `AquaDeploymentApp`.

Since `AquaDeploymentApp` already includes a `list_shapes` method, we can add a new method called `list_recommended_shapes` to maintain backward compatibility.

Looking ahead, if we need similar logic for other modules like fine-tuning or evaluation, we can follow the same pattern and add the corresponding methods in those apps as well.

As for naming, we could rename the current `AquaRecommendApp` to `AquaShapeRecommend`. I don't think we need to inherit from `AquaApp` in this case, since it's primarily a utility class focused on recommendation logic.

ads/aqua/shaperecommend/recommend.py
```diff
@@ -0,0 +1,499 @@
...
+        -------
+        List[ComputeShapeSummary]
+            List of ComputeShapeSummary objects passing the checks.
+
+        Raises
+        ------
+        ValueError
+            If the file cannot be opened, parsed, or the 'shapes' key is missing.
+        """
+        user_shapes = AquaDeploymentApp().list_shapes(compartment_id=compartment_id)
```
mrDzurb, line 337: To avoid calling `AquaDeploymentApp().list_shapes`, I think it would be cleaner to add a `shapes` method directly to the `OCIDataScienceModelDeployment` class. 

For example:

```python

class OCIDataScienceModelDeployment(
    OCIDataScienceMixin,
    oci.data_science.models.ModelDeployment,
):

  @classmethod
  def shapes(
      cls,
      compartment_id: str = None,
      **kwargs,
  ) -> List[oci.data_science.models.ModelDeploymentShapeSummary]:
      return oci.pagination.list_call_get_all_results(
          cls().client.list_model_deployment_shapes,
          compartment_id or COMPARTMENT_ID,
          **kwargs
      ).data
```

This makes the logic more self-contained within the deployment model and avoids relying on external app instances just to retrieve shape info.

The usage could be:

```
user_shapes = OCIDataScienceModelDeployment.shapes(compartment_id=compartment_id)
```

ads/aqua/shaperecommend/recommend.py
```diff
@@ -0,0 +1,499 @@
...
+
+        gpu_shapes_metadata = load_gpu_shapes_index().shapes
+
+        valid_shapes = []
+        # only loops through GPU shapes, update later to include CPU shapes
+        for name, spec in gpu_shapes_metadata.items():
+            if name in set_user_shapes:
+                compute_shape = set_user_shapes.get(name)
+                compute_shape.available = True
+                compute_shape.shape_series = "GPU"
```
mrDzurb, line 348: The ```oci.data_science.models.ModelDeploymentShapeSummary``` already contains the ```def shape_series(self, shape_series):```. Maybe we can use it?

ads/aqua/shaperecommend/recommend.py
```diff
@@ -0,0 +1,421 @@
...
+        Validates input, retrieves the model configuration, checks the requested sequence length,
+        identifies available and valid compute shapes, and summarizes which shapes are compatible
+        with the current model settings.
+
+        Parameters
+        ----------
+        ocid : str
+           OCID of the model to recommend feasible compute shapes.
+
+        available_shapes : List[ComputeShapeSummary]
```
mrDzurb, line 58: If we want to prepare the available shapes in advance and pass them to the recommender, I think it would make more sense to move this parameter into the constructor:

```python
class AquaShapeRecommend(BaseModel):
    available_shapes: List[ComputeShapeSummary] = Field(
        ..., description="List of available shapes in OCI."
    )
```

This way, we can decouple the shape fetching logic from the recommender logic and keep the design cleaner.

Alternatively, as previously discussed, we could define a method like `def available_shapes()` directly within the `AquaShapeRecommend` class. This method can internally use the `OciDataScienceModelDeployment.shapes()` function to retrieve the initial list of shapes. We can then move the ```valid_compute_shapes``` mehtod form ```deployment.py```  there.

My vote would be for the second approach.

mrDzurb (approved):

PR #1226, opened 2025-07-14
Fix inference failure caused by improper handling of large chat templates

siddhantm005:
### **Description:**

This PR addresses an issue where large chat templates stored in the model's custom metadata were being returned as improperly encoded strings, resulting in frontend inference failures.

The returned response is now safely decoded to UTF-8 when it’s in bytes form, preserving chat template regardless of size.

---

### **Changes:**
- Updated the get() handler in AquaModelChatTemplateHandler to correctly decode byte strings returned by get_custom_metadata_artifact.
- Adjusted related unit tests to reflect updated error messages and behavior.


kumar-shivam-ranjan (approved):
LGTM

PR #1234, opened 2025-07-22
List models using open ai api

agrimk:
### Description

- Add a new handler in AQUA SDK to call /predict/v1/models and return the deployed model list. 
- this is called from AQUA UI via the following format - 
`BASEURL/aqua/deployments/models/list/<md_ocid>`

### DIFF
--- a/ads/aqua/extension/deployment_handler.py
+++ b/ads/aqua/extension/deployment_handler.py
@@ -395,10 +395,10 @@ class AquaModelListHandler(AquaAPIhandler):
         self.set_header("Content-Type", "application/json")
         endpoint: str = ""
         model_deployment = AquaDeploymentApp().get(model_deployment_id)
-        endpoint = model_deployment.endpoint.rstrip("/") + "/predict/v1/models"       
+        endpoint = model_deployment.endpoint.rstrip("/") + "/predict/v1/models"
         aqua_client = Client(endpoint=endpoint)
         try:
-            list_model_result = aqua_client.list_models()
+            list_model_result = aqua_client.fetch_data()
             return self.finish(list_model_result)
         except Exception as ex:
             raise HTTPError(500, str(ex))


ads/aqua/extension/deployment_handler.py
```diff
@@ -373,6 +373,38 @@ def post(self, *args, **kwargs):  # noqa: ARG002
...
+        Raises
+        ------
+        HTTPError
+            Raises HTTPError if inputs are missing or are invalid
+        """
+
+        self.set_header("Content-Type", "application/json")
+
+        model_deployment = AquaDeploymentApp().get(model_deployment_id)
+        endpoint = model_deployment.endpoint + "/predict/v1/models"
```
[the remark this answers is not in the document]
agrimk: I have added a check for the same.

ads/aqua/client/client.py
```diff
@@ -582,6 +582,19 @@ def embeddings(
         payload = {**(payload or {}), "input": input}
         return self._request(payload=payload, headers=headers)
 
+    def list_models(self) -> Union[Dict[str, Any], Iterator[Mapping[str, Any]]]:
```
VipulMascarenhas, line 585: the logic in this method here seems generic, should we call it `fetch_data` or `get_json_response` instead of`list_models`? Also, docstrings need to be changed here.

ads/aqua/extension/deployment_handler.py
```diff
@@ -373,6 +373,40 @@ def post(self, *args, **kwargs):  # noqa: ARG002
...
+        Raises
+        ------
+        HTTPError
+            Raises HTTPError if inputs are missing or are invalid
+        """
+
+        self.set_header("Content-Type", "application/json")
+        endpoint: str = ""
+        model_deployment = AquaDeploymentApp().get(model_deployment_id)
+        if model_deployment.endpoint.endswith("/"):
```
VipulMascarenhas, line 398: we can simplify this with something like:
endpoint = model_deployment.endpoint.rstrip("/") + "/predict/v1/models"

VipulMascarenhas (commented):
added some minor comments

VipulMascarenhas (approved):
lgtm 👍 
there are a few failing tests, can you check them?

mrDzurb (approved):

vikasray0208 (approved):

PR #1236, opened 2025-07-30
[AQUA] Use custom model name in deployment

agrimk:
### Description

- Updated the create Model Deployment API to accept custom Model name which would be used while inferencing.


ads/aqua/model/model.py
```diff
@@ -147,6 +147,7 @@ class AquaModelApp(AquaApp):
     def create(
         self,
         model_id: Union[str, AquaMultiModelRef],
+        display_name: Optional[str] = None,
```
mrDzurb, line 150: It can be risky to insert attributes in the middle of an existing list, as it may lead to unintended side effects or compatibility issues. A better practice is to append new attributes at the end of the list .

ads/aqua/model/model.py
```diff
@@ -204,6 +207,9 @@ def create(
             **(defined_tags or {}),
         }
 
+        if display_name:
```
mrDzurb, line 210: I’m not sure I understand the logic here. I thought we wanted to introduce the model name that would be used for inferencing, but here we’re changing the display name of the MC record.
Could you add more description in the PR explaining what this PR for?

ads/aqua/modeldeployment/deployment.py
```diff
@@ -529,6 +529,15 @@ def _create(
...
         # validate user provided params
         user_params = env_var.get("PARAMS", UNKNOWN)
+
+        if (
+            create_deployment_details.model_name
+            and "--served-model-name" not in user_params
+        ):
+            user_params = f"{user_params} --served-model-name {create_deployment_details.model_name}"
+        elif "--served-model-name" not in user_params:
+            user_params = f"{user_params} --served-model-name odsc-llm"
```
mrDzurb, line 539: Can the ```odsc-llm``` be taken from the container params? I don't think that we want to hardcode this on the ADS side. Also ```--served-model-name``` will come from the container config as well, and then the user's params and containers one will be merged using 

```
deployment_params = get_combined_params(config_params, user_params)
```

Could you double check if the user's defined will not be overwritten with the default one from the container config?

Wouldn't this be enough?

```
if (
            create_deployment_details.model_name
            and "--served-model-name" not in user_params
        ):
            user_params = f"{user_params} --served-model-name {create_deployment_details.model_name}"
```

VipulMascarenhas (approved):

ads/aqua/modeldeployment/deployment.py
```diff
@@ -554,6 +552,18 @@ def _create(
         deployment_params = get_combined_params(config_params, user_params)
 
         params = f"{params} {deployment_params}".strip()
+
+        if create_deployment_details.model_name:
+            # Replace existing --served-model-name argument if present, otherwise add it
+            if "--served-model-name" in params:
```
lu-ohai, line 558: We have helper api to update the parameter in `params` string and you can do something like this
```
params_dict = get_params_dict(params)
params_dict.update({"--served-model-name": create_deployment_details.model_name})
params = build_params_string(params_dict)
```

lu-ohai (approved):

PR #1239, opened 2025-08-01
Custom model telemetry

agrimk:
### Description

- Passing the aqua_custom_base_model tag along of model to model deployment

### DIFF
--- a/ads/aqua/modeldeployment/deployment.py
+++ b/ads/aqua/modeldeployment/deployment.py
@@ -199,7 +199,7 @@ class AquaDeploymentApp(AquaApp):
         if create_deployment_details.instance_shape.lower() not in available_shapes:
             raise AquaValueError(
                 f"Invalid Instance Shape. The selected shape '{create_deployment_details.instance_shape}' "
-                f"is not available in the {self.region} region. Please choose another shape to deploy the model."
+                f"is not supported in the {self.region} region. Please choose another shape to deploy the model."
             )
 
         # Get container config


ads/aqua/modeldeployment/deployment.py
```diff
@@ -785,8 +794,8 @@ def _create_deployment(
         # we arbitrarily choose last 8 characters of OCID to identify MD in telemetry
         telemetry_kwargs = {"ocid": get_ocid_substring(deployment_id, key_len=8)}
 
-        if Tags.BASE_MODEL_CUSTOM in tags:
-            telemetry_kwargs["custom_base_model"] = True
+        if is_custom_base_model:
```
VipulMascarenhas, line 797: can you add some details in the description why this approach is needs vs. what we already had?

ads/aqua/modeldeployment/deployment.py
```diff
@@ -214,6 +214,7 @@ def create(
             )
             aqua_model = model_app.create(
                 model_id=create_deployment_details.model_id,
+                display_name=create_deployment_details.model_name,
```
VipulMascarenhas, line 217: I think we did not add model name intentionally to avoid leaking user input model names, instead relied on the config information to get the model family. Can you clarify what is sent via UI here?

ads/aqua/modeldeployment/deployment.py
```diff
@@ -647,6 +651,7 @@ def _create_multi(
         }
 
         model_name = f"{MODEL_NAME_DELIMITER} ".join(model_name_list)
+        is_custom_base_model = Tags.BASE_MODEL_CUSTOM in aqua_model.freeform_tags
```
VipulMascarenhas, line 654: instead of creating `is_custom_base_model`, copy the tag in the _create function itself:
```
 for tag in [
            Tags.AQUA_SERVICE_MODEL_TAG,
            Tags.AQUA_FINE_TUNED_MODEL_TAG,
            Tags.AQUA_TAG,
            Tags.BASE_MODEL_CUSTOM 
        ]:
```

vikasray0208 (approved):

VipulMascarenhas (approved):

mrDzurb (approved):

PR #1247, opened 2025-08-12
[AQUA] handling Model Deployment list for model group deployment type

agrimk:
### Description

Earlier only UNKNOWN_ENUM_VALUE  deployment type was checked for and ignored. Now we are also checking for MODEL_GROUP deployment type as well. This deployment_type is introduced as part of latest version of oci.


ads/aqua/modeldeployment/deployment.py
```diff
@@ -871,14 +871,21 @@ def list(self, **kwargs) -> List["AquaDeployment"]:
...
-                        model_deployment, self.region
+                try:
+                    results.append(
+                        AquaDeployment.from_oci_model_deployment(
+                            model_deployment, self.region
+                        )
                     )
-                )
+                except Exception as e:
+                    logger.error(
```
VipulMascarenhas, line 884: should we raise error here instead of continuing with the deployment?

ads/aqua/modeldeployment/deployment.py
```diff
@@ -871,14 +871,21 @@ def list(self, **kwargs) -> List["AquaDeployment"]:
                 if (
                     "datasciencemodelgroup" in aqua_model_id
                     or model_deployment.model_deployment_configuration_details.deployment_type
-                    == "UNKNOWN_ENUM_VALUE"
+                    in ["UNKNOWN_ENUM_VALUE", "MODEL_GROUP", "SINGLE_MODEL_FLEX"]
```
lu-ohai, line 874: Can we make this as constants? There is an example: https://github.com/oracle/accelerated-data-science/blob/feature/model_group/ads/model/deployment/model_deployment.py#L84

lu-ohai (approved):

lu-ohai:
@agrimk could you check the failed unit tests and merge it after?

PR #1257, opened 2025-08-29
Update GPU shapes index with additional shapes and CPU parameters

mrDzurb:
# Description

This PR updates the GPU shapes index file to:
* Add new GPU shapes
* Include additional CPU parameters

## Motivation

Keeping the GPU shapes index up to date ensures accurate configuration and compatibility across deployments. Adding CPU parameters provides more complete resource specifications for better validation and deployment planning.

### DIFF
--- a/ads/aqua/common/entities.py
+++ b/ads/aqua/common/entities.py
@@ -46,37 +46,77 @@ class ModelConfigResult(BaseModel):
         arbitrary_types_allowed = True
         protected_namespaces = ()
 
+
 class ComputeRank(Serializable):
     """
-    Represents the cost and performance ranking for a compute shape.
+    Represents the cost and performance rankings for a specific compute shape.
+    These rankings help compare different shapes based on their relative pricing
+    and computational capabilities.
     """
-    cost: int = Field(
-    None, description="The relative rank of the cost of the shape. Range is [10 (cost-effective), 100 (most-expensive)]"
+
+    cost: Optional[int] = Field(
+        None,
+        description=(
+            "Relative cost ranking of the compute shape. "
+            "Value ranges from 10 (most cost-effective) to 100 (most expensive). "
+            "Lower values indicate cheaper compute options."
+        ),
     )
 
-    performance: int = Field(
-    None, description="The relative rank of the performance of the shape. Range is [10 (lower performance), 110 (highest performance)]"
+    performance: Optional[int] = Field(
+        None,
+        description=(
+            "Relative performance ranking of the compute shape. "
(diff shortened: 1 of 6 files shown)


darenr (dismissed):
Here are some potential issues found in the pull request changes:

1. ads/aqua/common/entities.py:

   - The `ComputeRank` and `GPUSpecs` models now use `Optional[int]` for fields that were previously just `int`. If any downstream code assumes these fields are always set (non-None), it could cause runtime errors or unexpected behavior.

   - The validator method in `ComputeShapeSummary` was renamed from `set_gpu_specs` to `populate_gpu_specs`, which is good for clarity, but make sure all references are updated elsewhere.

   - There is a code comment referencing extracting GPU count with regex for shape names like "VM.GPU3.2", but the regex used is `r"\.(\d+)$"`—this will not match "VM.GPU3.2" but will match "VM.GPU.A10.2" (dot before number at end). If you expect both formats, this may miss certain shapes.

   - Logging inside the validator uses `logger.debug`, but if `logger` is not defined or imported, this will cause a NameError.

2. ads/aqua/resources/gpu_shapes_index.json:

   - The JSON structure now includes new top-level items like `"BM.GPU.B200.8"` and `"BM.GPU.GB200.4"` outside the `"shapes"` object. This breaks the previous schema, where all shapes were inside the `"shapes"` key. Downstream code expecting all shapes under `"shapes"` will miss the new entries.

   - The new shapes and CPU fields are comprehensive, but check that clients or code loading this file are updated to handle the new structure and that it won’t break existing deployments.

Summary:

- There are potential compatibility issues with both the Python and JSON changes.  
- Make sure all code that reads or validates this data is aware of the new schema and field behaviors.
- Double-check shape name parsing and logger definitions.

ads/aqua/common/entities.py
```diff
@@ -97,50 +137,49 @@ class GPUShapesIndex(Serializable):
...
     gpu_specs: Optional[GPUSpecs] = Field(
-        default=None,
-        description="Optional GPU specifications associated with the shape.",
+        default=None, description="GPU configuration for the shape, if applicable."
     )
 
     @model_validator(mode="after")
     @classmethod
-    def set_gpu_specs(cls, model: "ComputeShapeSummary") -> "ComputeShapeSummary":
+    def populate_gpu_specs(cls, model: "ComputeShapeSummary") -> "ComputeShapeSummary":
```
VipulMascarenhas, line 171: nit: I'm assuming this name change would not affect elsewhere since I didn't come across any references/usage within ads.

lu-ohai (approved):

PR #1259, opened 2025-09-03
[ODSC-76829/76830] : improve auto-select logic and handle missing data

codeloop:
Improve auto-select logic and handle missing data

This commit introduces following to the forecasting operator.

**Improved `AUTO_SELECT_SERIES` Logic:**
 - A fallback to the `AUTO_SELECT` model has been implemented for cases where `AUTO_SELECT_SERIES` is used without specifying `target_category_columns`.

**Missing Data Handling:**
 - The `build_fforms_meta_features` function now fills missing values in the target column with zeros. This prevents errors during meta-feature calculation when the data contains NaNs.

**New Test Case:**
- A new test case has been added to verify that the `auto-select-series` model functions correctly with datasets containing missing values.


ads/opctl/operator/lowcode/common/transformations.py
```diff
@@ -329,6 +329,8 @@ def build_fforms_meta_features(self, data, target_col=None, group_cols=None):
         if target_col not in data.columns:
             raise ValueError(f"Target column '{target_col}' not found in DataFrame")
 
+        data[target_col] = data[target_col].fillna(0)
```
ahosler, line 332: why fillna with 0? why no backfill? Did we discuss this? 
Don't we already have this covered in pre-processing steps? What are we gaining from this?

ads/opctl/operator/lowcode/forecast/__main__.py
```diff
@@ -75,6 +74,15 @@ def operate(operator_config: ForecastOperatorConfig) -> ForecastResults:
...
     else:
+        # When AUTO_SELECT_SERIES is specified but target_category_columns is not,
+        # we fall back to AUTO_SELECT behavior.
+        if (
+            operator_config.spec.model == AUTO_SELECT_SERIES
+            and not operator_config.spec.target_category_columns
+        ):
+
+            operator_config.spec.model = AUTO_SELECT
+            model = ForecastOperatorModelFactory.get_model(operator_config, datasets)
```
ahosler, line 85: nice!

Can we reflect this in the report? Make sure it's still saying "auto-select-series".

Can we add a unit test for this?

ahosler (changes requested):
A couple of comments. Just needs a bit of polish

ads/opctl/operator/lowcode/common/transformations.py
codeloop: This is independent of the pre-processing, we want this to be done even if the pre-processing is not set/enabled. This helps us to create the meta-features incase we have nan's in the dataset. Filling with other values is not validated yet, can be experimented and documented to see the impact.

ahosler (approved):

mrDzurb (approved):

PR #1261, opened 2025-09-07
Add post-processing step for forecast clipping

codeloop:
Add post-processing step for forecast clipping

This change introduces a new post-processing step for the forecast operator that allows users to clip the forecast output to a specified minimum and maximum value. This is useful for ensuring that the forecast values remain within a realistic or desirable range.

  This change includes:
   - A new `PostprocessingSteps` configuration to define the min/max clipping values.
   - Updates to the operator schema to expose the new configuration.
   - Implementation of the clipping logic using `numpy.clip`.
   - A new test case to verify the clipping functionality.


prasankh (commented):
Could you also update yaml_schema.rst for the documentation?

PR #1264, opened 2025-09-09
AQUA. Support env var overrides and enhance multi-model entity structure

mrDzurb:
# Description
This pull request introduces enhancements and refactoring aimed at improving the deployment configuration flow for both single-model and multi-model deployment scenarios.

### Key Enhancements

1. **Support for Default Environment Variables in Single Model Deployment**

   * Added logic to read default environment variables from the model config.
   * Users can override these environment variables explicitly when needed.

2. **Code Refactoring for Configuration Parsing**

   * Cleaned up and consolidated helper methods for parsing CLI-style parameters (`params`) and environment variables (`env_var`).
   * Improved function docstrings and test coverage for better maintainability.

3. **Multi-Model Deployment Entity Enhancements**

   * Extended the `AquaMultiModelRef` structure to support both `params` and `env_var` sections.
   * This enhancement is **mandatory** for enabling full control over runtime configuration of individual models within a multi-model deployment group.

###  Test Coverage

* Unit tests were added and extended to ensure:

  * Proper merging of env variables from config and overrides.
  * Accurate parsing of param strings into dictionary format.
  * Backward compatibility with existing behavior (e.g., `PARAMS` env var parsing).
  * All fields in `AquaMultiModelRef` behave correctly with defaults and overrides.

### DIFF
--- a/ads/aqua/common/entities.py
+++ b/ads/aqua/common/entities.py
@@ -368,7 +368,7 @@ class AquaMultiModelRef(Serializable):
         Dict[str, str]
             Dictionary with parameter names as keys and their corresponding values as strings.
         """
-        if not params:
+        if not params or not isinstance(params, (str, list)):
             return {}
 
         # Normalize string to list of "--key value" strings
--- a/ads/aqua/common/entities.py
+++ b/ads/aqua/common/entities.py
@@ -3,7 +3,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 
 import re
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 from oci.data_science.models import Model
 from pydantic import BaseModel, ConfigDict, Field, model_validator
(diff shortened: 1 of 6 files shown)


ads/aqua/common/entities.py
```diff
@@ -302,8 +318,80 @@ def all_model_ids(self) -> List[str]:
...
+            return {}
+
+        # Normalize string to list of "--key value" strings
+        if isinstance(params, str):
+            params_list = [
+                f"--{param.strip()}" for param in params.split("--") if param.strip()
+            ]
+        else:
+            params_list = params
+
```
elizjo, line 381: maybe safer to explicitly check if isinstance(params, list)?

ads/aqua/common/entities.py
```diff
@@ -302,8 +318,80 @@ def all_model_ids(self) -> List[str]:
...
+        This is useful for backward compatibility where users pass CLI-style
+        parameters via environment variables, e.g.:
+        env_var = { "PARAMS": "--max-model-len 65536 --enable-streaming" }
+
+        If `params` is already set, values from `PARAMS` in `env_var` are added
+        only if they do not override existing keys.
+        """
+        env = values.get("env_var", {})
+        param_string = env.pop("PARAMS", None)
+
```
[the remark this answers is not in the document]
mrDzurb: Right, I don't want to keep the PARAMS in env variables, because this entire record will be saved in the MC custom metadata and will be used eventually to build env variables for the MD. So POP sill delete the PARAMS from env variables if such key is presented there.

ads/aqua/common/entities.py
mrDzurb: Done.

elizjo (approved):

ads/aqua/common/entities.py
```diff
@@ -302,8 +318,80 @@ def all_model_ids(self) -> List[str]:
...
+        Dict[str, str]
+            Dictionary with parameter names as keys and their corresponding values as strings.
+        """
+        if not params or not isinstance(params, (str, list)):
+            return {}
+
+        # Normalize string to list of "--key value" strings
+        if isinstance(params, str):
+            params_list = [
+                f"--{param.strip()}" for param in params.split("--") if param.strip()
```
VipulMascarenhas, line 377: should we use params.split("--")[1:]] in cases where there can be multiple values after the key, for example -lora-modules speech=artifact/speech-lora vision=artifact/vision-lora ?

ads/aqua/common/utils.py
```diff
@@ -800,35 +800,49 @@ def is_service_managed_container(container):
...
 
+    Example
+    -------
+    >>> get_params_list("--max-model-len 65536 --enforce-eager")
+    ['--max-model-len 65536', '--enforce-eager']
     """
     if not params:
         return []
-    return ["--" + param.strip() for param in params.split("--")[1:]]
+    return [f"--{param.strip()}" for param in params.split("--") if param.strip()]
```
VipulMascarenhas, line 823: same comment as before.

VipulMascarenhas (approved):

ads/aqua/modeldeployment/deployment.py
```diff
@@ -520,10 +520,20 @@ def _create(
...
 
+        # Loads frameworks specific default params from the configuration
         config_params = deployment_config.configuration.get(
             create_deployment_details.instance_shape, ConfigurationItem()
         ).parameters.get(get_container_params_type(container_type_key), UNKNOWN)
 
+        # Loads default environment variables from the configuration
+        config_env = deployment_config.configuration.get(
+            create_deployment_details.instance_shape, ConfigurationItem()
+        ).env.get(get_container_params_type(container_type_key), {})
```
lu-ohai, line 531: Nit: we're using the following multiple times
```
get_container_params_type(container_type_key)
```

lu-ohai (approved):

PR #1369, opened 2026-04-09
Regression Operator

skjoracle:
Contains the code of regression operator. It has the framework for the operator, along with that 4 models are implemented.
        - knn
        - xgboost
        - linear_regression
        - random_forest
If you choose 'auto' as model, then it will run all of them and will select the best based on your selected metric.
Also contains explainability and model saving and deployment code.


ads/opctl/operator/lowcode/regression/deployment/score.py
```diff
@@ -0,0 +1,84 @@
...
+    Expected payload:
+    {
+      "data": dataset_df
+    }
+    """
+    model_obj, spec = _resolve_model_and_spec(model or load_model())
+    target_column = spec.get("target_column")
+
+    x = _build_input_frame(data, target_column=target_column)
+    print(x)
```
prasankh: It ends up logging whatever payload callers send. Let's drop it before merge.

prasankh (approved):

mrDzurb (approved):

PR #1376, opened 2026-05-04
Meta Learning Based Model Selection Per Series

prasankh:
Users can rely on the `auto-select-series` option to automatically pick a model for every time-series. It captures each series characteristics and infers from the trained meta model to recommend the most suitable forecasting model per series that performed best during meta-model training.


ads/opctl/operator/lowcode/forecast/model/meta_features.py
```diff
@@ -0,0 +1,471 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2025 Oracle and/or its affiliates.
```
mrDzurb: 2026?

mrDzurb (dismissed):

skjoracle (approved):

sambitkumohanty (approved):

PR #1385, opened 2026-06-09
Support what-if analysis for auto-select-series workflow

prasankh:
## Summary
- Add what-if support for `auto-select-series` workflow.
- Use one `model.pkl` artifact for auto-select-series what-if analysis, instead of separate files for the different models used across series.
- Exclude Theta and ETS from what-if analysis because they do not use exogenous features.
- Update forecast operator docs for selection strategy guidance and model behavior.

[ODSC-87420](https://jira.oci.oraclecorp.com/browse/ODSC-87420)
[ODSC-87436](https://jira.oci.oraclecorp.com/browse/ODSC-87436)

### DIFF
--- a/ads/opctl/operator/lowcode/forecast/model/forecast_datasets.py
+++ b/ads/opctl/operator/lowcode/forecast/model/forecast_datasets.py
@@ -467,7 +467,6 @@ class ForecastResults:
         "local_explanations",
         "global_explanations",
         "model_parameters",
-        "models",
         "errors_dict",
     ]
 
--- a/ads/opctl/operator/lowcode/forecast/__main__.py
+++ b/ads/opctl/operator/lowcode/forecast/__main__.py
@@ -94,6 +94,19 @@ def _merge_auto_select_series_backtesting_results(
     return results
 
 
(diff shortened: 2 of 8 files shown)


ads/opctl/operator/lowcode/forecast/model/forecast_datasets.py
```diff
@@ -467,7 +467,6 @@ class ForecastResults:
         "local_explanations",
         "global_explanations",
         "model_parameters",
-        "models",
```
skjoracle, line 470: why is this removed? it was used by auto_select and auto_select_series backtesting merge logic. after removing is it still giving proper results.

prasankh: The merge logic for `model.pkl` has changed here intentionally. The earlier generic merge was not correct for `auto-select-series` especially for global models.

updated logic : 
https://github.com/oracle/accelerated-data-science/blob/5967fc8aa4ab04f88752ae2bf164b5993d9cdbc2/ads/opctl/operator/lowcode/forecast/__main__.py#L183-L191

skjoracle (approved):

sambitkumohanty (approved):

PR #1401, opened 2026-08-11
Batch prediction support for Regression Operator

prasankh:
Supports scoring a separate unlabeled dataset with the Regression Operator while making the Model Catalog and Model Deployment flow optional.

#### Changes
- Add optional `prediction_data` for batch scoring.
- Add optional `prediction_output` configuration.
- Write batch results to `predictions.csv`.
- Preserve `prediction_data` row order.
- Include all `prediction_data` columns by default.
- Support selecting specific output columns with `passthrough_columns`.
- Keep `test_data` for labeled model evaluation.
- Use `deployment_info.json` as the lifecycle metadata artifact.
- Support for optional Model Catalog and Model Deployment flows.
- Update the Regression Operator schema and documentation.

### DIFF
--- a/ads/opctl/operator/lowcode/regression/model/base_model.py
+++ b/ads/opctl/operator/lowcode/regression/model/base_model.py
@@ -3,6 +3,7 @@
 # Copyright (c) 2026 Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 
+import json
 import logging
 import os
 import tempfile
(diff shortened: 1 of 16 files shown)


skjoracle (approved):

ads/opctl/operator/lowcode/regression/model/base_model.py
```diff
@@ -465,6 +480,18 @@ def _build_predictions_output_df(self, predictions_df: pd.DataFrame) -> pd.DataF
...
         )
 
+    def _build_batch_predictions_output_df(self) -> pd.DataFrame:
+        """Builds batch output while preserving prediction_data row order."""
+        if self.datasets.prediction_data is None or self.prediction_values is None:
+            return pd.DataFrame()
+
+        columns = self.datasets.prediction_passthrough_columns
+        result = (
+            self.datasets.prediction_data.loc[:, columns].reset_index(drop=True).copy()
```
sambitkumohanty, line 490: So if prediction_data already has a column named prediction, and that column is included as a passthrough, it gets overwritten by the model’s output. This can happen by default because omitting
  passthrough_columns means “copy all prediction_data columns.

sambitkumohanty (approved):

ads/opctl/operator/lowcode/regression/model/base_model.py
prasankh: updated [here](https://github.com/oracle/accelerated-data-science/pull/1401/commits/f25774552ee89a71ecf1c692104336b1fa200d0d)

